PR #3, opened 2017-06-02
Added TLS1.2 support to client and server.

Zazzmatazz:
Added MqttClientSslChannel and MqttServerSslChannel which both utilize an SSL stream.  Added options to both MqttCLientOptions and MqttServerOptions.  I was unable to update the code in MQTTnet.UniversalWindows due to my lack of Windows 10, but the codeset is fairly simple to understand.

Feel free to update any restrictions such as certificate validation, or simply add them to the client options.

### DIFF
--- a/Frameworks/MQTTnet.NetCoreApp/MqttClientFactory.cs
+++ b/Frameworks/MQTTnet.NetCoreApp/MqttClientFactory.cs
@@ -13,8 +13,6 @@ namespace MQTTnet
             if (options == null) throw new ArgumentNullException(nameof(options));
 
             return new MqttClient(options,
-                // The cast to IMqttCommunicationChannel is required... Roslyn is too stupid to
-                // figure out how to cast back to the base type
                 new MqttChannelCommunicationAdapter(options.UseSSL ? new MqttClientSslChannel() : (IMqttCommunicationChannel) new MqttTcpChannel(),
                     new DefaultMqttV311PacketSerializer()));
         }
--- a/Frameworks/MQTTnet.NetCoreApp/MqttClientFactory.cs
+++ b/Frameworks/MQTTnet.NetCoreApp/MqttClientFactory.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using MQTTnet.Core.Adapter;
+using MQTTnet.Core.Channel;
 using MQTTnet.Core.Client;
 using MQTTnet.Core.Serializer;
 
@@ -11,7 +12,11 @@ namespace MQTTnet
         {
             if (options == null) throw new ArgumentNullException(nameof(options));
 
-            return new MqttClient(options, new MqttChannelCommunicationAdapter(new MqttTcpChannel(), new DefaultMqttV311PacketSerializer()));
+            return new MqttClient(options,
+                // The cast to IMqttCommunicationChannel is required... Roslyn is too stupid to
+                // figure out how to cast back to the base type
+                new MqttChannelCommunicationAdapter(options.UseSSL ? new MqttClientSslChannel() : (IMqttCommunicationChannel) new MqttTcpChannel(),
+                    new DefaultMqttV311PacketSerializer()));
         }
     }
 }
--- a/Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
+++ b/Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
@@ -1,0 +1,122 @@
+﻿using System;
+using System.IO;
+using System.Net.Security;
+using System.Net.Sockets;
+using System.Security.Authentication;
+using System.Threading.Tasks;
+using MQTTnet.Core.Channel;
+using MQTTnet.Core.Client;
+using MQTTnet.Core.Exceptions;
+
+namespace MQTTnet
+{
+    /// <summary>
+    /// Describes an SSL channel to an MQTT server.
+    /// </summary>
+    public class MqttClientSslChannel : IMqttCommunicationChannel, IDisposable
+    {
+        private readonly Socket _socket;
+        private SslStream _sslStream;
+
+        /// <summary>
+        /// Creates a new <see cref="MqttClientSslChannel"/>.
+        /// </summary>
+        public MqttClientSslChannel()
+        {
+            _socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="MqttClientSslChannel"/> with a predefined <paramref name="socket"/>.
+        /// </summary>
+        /// <param name="socket"></param>
+        public MqttClientSslChannel(Socket socket)
+        {
+            _socket = socket ?? throw new ArgumentNullException(nameof(socket));
+        }
+
+        /// <summary>
+        /// Asynchronously connects to the host described in the <see cref="MqttClientOptions"/>.
+        /// </summary>
+        /// <param name="options">The <see cref="MqttClientOptions"/> describing the connection.</param>
+        public async Task ConnectAsync(MqttClientOptions options)
+        {
+            try
+            {
+                await _socket.ConnectAsync(options.Server, options.Port);
+
+                NetworkStream ns = new NetworkStream(_socket, true);
+                _sslStream = new SslStream(ns);
+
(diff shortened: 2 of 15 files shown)


Frameworks/MQTTnet.NetCoreApp/MqttClientFactory.cs
```diff
@@ -11,7 +12,11 @@ public MqttClient CreateMqttClient(MqttClientOptions options)
         {
             if (options == null) throw new ArgumentNullException(nameof(options));
 
-            return new MqttClient(options, new MqttChannelCommunicationAdapter(new MqttTcpChannel(), new DefaultMqttV311PacketSerializer()));
+            return new MqttClient(options,
+                // The cast to IMqttCommunicationChannel is required... Roslyn is too stupid to
```
chkr1011: Please don't use such words :-) Even if they are true :-)

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// Asynchronously connects to the host described in the <see cref="MqttClientOptions"/>.
+        /// </summary>
+        /// <param name="options">The <see cref="MqttClientOptions"/> describing the connection.</param>
+        public async Task ConnectAsync(MqttClientOptions options)
+        {
+            try
+            {
+                await _socket.ConnectAsync(options.Server, options.Port);
+
+                NetworkStream ns = new NetworkStream(_socket, true);
```
chkr1011: I am not 100% sure but I assume the ns has to be disposed as well in the dispose method.

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
```diff
@@ -0,0 +1,122 @@
...
+                _sslStream = new SslStream(ns);
+
+                await _sslStream.AuthenticateAsClientAsync(options.Server, null, SslProtocols.Tls12, false);
```
chkr1011: You can return the Task directly if no code follows the await. Because otherwise the context will be created and restored for nothing. It will increase the performance.

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        /// <param name="buffer">The buffer to write data from.</param>
+        public async Task WriteAsync(byte[] buffer)
+        {
+            if (buffer == null)
+                throw new ArgumentNullException(nameof(buffer));
+
+            try
+            {
+                await _sslStream.WriteAsync(buffer, 0, buffer.Length);
```
chkr1011: You can directly return the task (see above).

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Asynchronously reads a sequence of bytes from the socket.
+        /// </summary>
+        /// <param name="buffer">The buffer to write the data into.</param>
+        public async Task ReadAsync(byte[] buffer)
+        {
+            try
+            {
+                await _sslStream.ReadAsync(buffer, 0, buffer.Length);
```
chkr1011: Same here... :)

Frameworks/MQTTnet.NetCoreApp/MqttServerFactory.cs
```diff
@@ -9,7 +10,8 @@ public MqttServer CreateMqttServer(MqttServerOptions options)
         {
             if (options == null) throw new ArgumentNullException(nameof(options));
 
-            return new MqttServer(options, new MqttServerAdapter());
+            // The cast to IMqttServerAdapter is required... stupidly...
```
chkr1011: Words... :-)

Frameworks/MQTTnet.NetCoreApp/MqttSslServerAdapter.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            _cancellationTokenSource = new CancellationTokenSource();
+
+            _socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
+            _socket.Bind(new IPEndPoint(IPAddress.Any, options.Port));
+            _socket.Listen(options.ConnectionBacklog);
+
+            Task.Run(async () => await AcceptConnectionsAsync(_cancellationTokenSource.Token), _cancellationTokenSource.Token);
+
+            _x590Certificate2 = new X509Certificate2(options.CertificatePath);
```
chkr1011: It is a very low chance but the task should be started if everything is assinged so that no null-ref is possible if the cert is used but not set.

Frameworks/MQTTnet.NetFramework/MqttServerFactory.cs
```diff
@@ -8,8 +9,9 @@ public class MqttServerFactory
         public MqttServer CreateMqttServer(MqttServerOptions options)
         {
             if (options == null) throw new ArgumentNullException(nameof(options));
-
-            return new MqttServer(options, new MqttServerAdapter());
+            
+            // The cast to IMqttServerAdapter is required... stupidly...
+            return new MqttServer(options, options.UseSSL ? (IMqttServerAdapter) new MqttSslServerAdapter() : new MqttServerAdapter());
```
chkr1011: I have some doubts because with this code either encryted or not encrypted will be supported. I have some device which can be encrypted but others are not and I don't have access to the firmware. I can just start 2 servers but then the clients are separated from each other. Maybe a list of adapters is a god solution to fix this.  So the options are having a boolean "ExposeSslEndpoint" and "ExposeUnencryptedEndpoint" and "EncryptedEndpointPort" and "UnencryptedEndpointPort" etc. and depending on this the factory can create a list of adapters with either both or only one. But this only  a draft of an idea :-) What do you think about this idea?

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
Zazzmatazz: The NetworkStream is being disposed of in Dispose(), since it is wrapped in the SSLStream instance.

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
Zazzmatazz: I didn't actually touch this code, this was part of your code, simply changed for the _sslStream (see MqttTcpChannel.cs line 55 already committed).  I will, however, change this.

Frameworks/MQTTnet.NetCoreApp/MqttServerFactory.cs
Zazzmatazz: Aw, what, programmers can't be snarky anymore?  Da-- I mean, gosh darnit!

Frameworks/MQTTnet.NetFramework/MqttServerFactory.cs
Zazzmatazz: I like this idea, though it would take a bit of work to implement.  Naturally, the encrypted endpoint should be prioritized.  The server code would need to be refactored to permit multiple adapters, but that shouldn't be terribly difficult.

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
Zazzmatazz: You can't actually return a task directly on an `async` method.  You can only `await` it.

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
Zazzmatazz: Now _here_, you can return a `Task` (by removing the `async`), since it's the only `await` in the method.

Frameworks/MQTTnet.NetCoreApp/MqttClientSslChannel.cs
Zazzmatazz: Ditto, noted above.

Zazzmatazz:
I've gone ahead and applied the changes requested.

PR #9, opened 2017-07-05
Adding support for .NET Standard 

1iveowl:
Your current library does not support .NET Standard. As in: You cannot reference the libreary from another .NET Standard project. You can only reference it from a .NET Core project.

This pull request adds this support of .NET Standard.

No code have been changed. I've added a project that links to you existing .NET Core project code and I've changed the nuspec file so that it should now create a .NET Standard library. To do this I had to reference System.Net.Security 4.3.1. This is also added to the nuspec file, but only for .NET Standard

Also, I could not go any lower that .NET Standard 1.3.

Hope you will find this helpful and will release a new Nuget version with .NET Standard 1.3 support. Let me know if it gives you any trouble.

Thank you for writing this library!

### DIFF
--- a/Frameworks/MQTTnet.Netstandard/MQTTnet.Netstandard.csproj
+++ b/Frameworks/MQTTnet.Netstandard/MQTTnet.Netstandard.csproj
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard1.3</TargetFramework>
+    <AssemblyName>MQTTnet</AssemblyName>
+    <RootNamespace>MQTTnet</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


chkr1011:
Hi,
thank you for your contribution. Good work!
But I have some problems building the new code. Maybe you can have a look at it?

I usually build everything using the "batch build" feature. So the required checkboxes must be checked. But I don't know exactly which ones. :-)

Maybe the CoreApp proejct is no longer needed? To me it seems that it is redundant now. Do you have an idea? In my opinion we can delete it and keep the Netstandard version only. But I am only guessing :-)

Best regards
Christian

1iveowl:
OK. I'm travelling. Let me look at it in the weekend. I created a PowerShell script for building, packaging and pushing the NuGet, but I don't have it here with me.

And yes, you don't need the CoreApp project.

chkr1011:
OK don't worry. It is not urgent.

PR #14, opened 2017-07-13
Added missing Nuspec file

1iveowl:
Added missing Nuspec file

### DIFF
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -1,0 +1,44 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>MQTTnet</id>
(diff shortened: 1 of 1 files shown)


1iveowl:
Looks like you deleted the Build directory. Didn't like the ps script or was it because of the missing nuget spec. file?

chkr1011:
The _nuspec_ was missing. So I decided to revert it but the paths are all different and thus it doesn't work. I can merge this and can restore the _Build_ folder because it makes sense to use it.

1iveowl:
I don't really understand what is happening here, but something is not working with the merge. I also see that the build.ps1 file is missing in your version, but it is there in mine.

PR #15, opened 2017-07-14
Merge fixes

1iveowl:
Could you try this one. 

I updated the .gitnore file as it was creating issues. I used the latest standard vs2017 .gitnore file.

I've not changed any code.

Please let me know if this works? I tested it here and I am able to create the NuGet with the `build.ps1` PowerShell script


chkr1011:
Hi, I merged it but I moved some files (to Build folder). I hope you agree :)
I have some doubts because all the platform specific entries are removed from the nuspec!?
Will it work? I will test it on my Pi later (ARM) and I hope it will :-)

1iveowl:
Perfect.

Does it build now? 

Yes, it should work. I don't think you need all those specific platformfiles. The Any CPU files I've included should be enough as they are platform agnostic as I understand it, but please test on the RPi and see if it works for you. Let me know the outcome.

Are you going to release a new NuGet?

chkr1011:
It builds. I will test it for ARM etc. If it works I will release a new nuget. Probably today...

1iveowl:
Did it work with ARM?

Anyhow, I created an Rx Wrapper for your library. See here: [MQTTClientRx](https://github.com/1iveowl/MQTTClient.rx) and here [MQTTClientRx Nuget](https://www.nuget.org/packages/MQTTClientRx/). 

It's only for the Client part. Not sure that Rx adds a lot of benefit to the server part.

chkr1011:
Works :-)
Wow good work with your library. I would like to add a link to my README. Is it OK for you?

1iveowl:
Of course you can add a link and thank you 👍

PR #34, opened 2017-09-08
fixed bug in read async and unfied stream handling

### LINKED ISSUE #33: MqttTcpChannel buffer broken
while looking at the code i found an issue with MqttTcpChannel.

ReadAsync does not handle the returncode of _socket.ReceiveAsync. so if calling code wants to read 1000 bytes and method returns 500 just 500 bytes are read not the desired 1000. (byte array has correct size but half of it is unitialized)

JanEggers:
similar change need to be applied to uwp but that did not build on my machine so i didnt change it

fixes #33


chkr1011:
Please target the develop branch until the performance changes are completed and the results are measured. For this project the master branch usually contans stable release versions.

Please fix the breaking changes. Then I will merge it.

PR #36, opened 2017-09-10
added perf test app

JanEggers:
this i use for perf test and wanted to share so we can reproduce perf monitorings / bugs because of high traffic concurrency


JanEggers:
my pr with the bugfixes has to accepted first then ill rebase and test

JanEggers:
@chkr1011 : see last comment....

 if you start perftest now it takes some time then you get exceptions in msgdispatcher that i fixed in the other pr

PR #43, opened 2017-09-24
Bug fixes v2

1iveowl:
I fixed a few bugs.
On `MQTTnet.UniversalWindows.csproj` : it's best with no refererence to System.Net.Security as it's not compatible with UWP (see [here](https://github.com/dotnet/corefx/issues/19783))

On `MqttChannelCommunicationAdapter.cs`. You lock on _channel to compose a Task. But shouldn't the FlushAsync be part of this Task?

Also, I updated .NET Standard 1.3 to use the latest version of System.Net.Security.


1iveowl:
Just another error

PR #44, opened 2017-09-24
Bug fixes v3

1iveowl:
I fixed a few bugs.

On `MQTTnet.UniversalWindows.csproj` : it's best with no refererence to System.Net.Security as it's not compatible with UWP (see [here](https://github.com/dotnet/corefx/issues/19783))

On `MqttChannelCommunicationAdapter.cs `. You lock on _channel to compose a Task. But shouldn't the FlushAsync be part of this Task?

Also, I updated .NET Standard 1.3 to use the latest version of System.Net.Security.

### DIFF
--- a/MQTTnet.Core/Adapter/MqttChannelCommunicationAdapter.cs
+++ b/MQTTnet.Core/Adapter/MqttChannelCommunicationAdapter.cs
@@ -83,7 +83,7 @@ namespace MQTTnet.Core.Adapter
                 {
                     foreach (var packet in packets)
                     {
-                        if (packet == null) {continue};
+                        if (packet == null) {continue; }
 
                         MqttTrace.Information(nameof(MqttChannelCommunicationAdapter), $"TX >>> {packet} [Timeout={timeout}]");
 


1iveowl:
Please ignore the first two. This one passes the checks.

MQTTnet.Core/Adapter/MqttChannelCommunicationAdapter.cs
```diff
@@ -83,23 +83,27 @@ public MqttChannelCommunicationAdapter(IMqttCommunicationChannel channel, IMqttP
                 {
                     foreach (var packet in packets)
                     {
-                        MqttTrace.Information(nameof(MqttChannelCommunicationAdapter), "TX >>> {0} [Timeout={1}]", packet, timeout);
+                        if (packet == null) {continue; }
+
+                        MqttTrace.Information(nameof(MqttChannelCommunicationAdapter), $"TX >>> {packet} [Timeout={timeout}]");
```
chkr1011: Please do not use string interpolation for logging. While optimizing performance we figured out that the trace is consuming a lot of performance. So the good old string.format is the best solution at the moment :-)

PR #47, opened 2017-09-27
A different approach to managing Stream.WriteAsync

1iveowl:
Trying a different approach to sending messages


chkr1011:
I think we don't need this anymore... 
But I want to thank you for this try to solve the issue!

PR #51, opened 2017-10-02
Fix50 eventing

### LINKED ISSUE #50: ClientConnected does not use ClientId and does not match connectedclientlist
hi two issue around client connected event: 

1. if a client connected event is triggered Server.GetConnectedClients() does not contain the new client.
id expect that the client is in the list of connected cliensts if an event signals this. 

2. the identifier in the event is the remote endpoint of that socket or something else for websockets.
id expect the client id of the connect packet in the event.

id like to submit a pr either change the existing behavior or add another event like session created.

@chkr1011  thoughts?

JanEggers:
fixes #50


JanEggers:
sry for changing the testproject but it wasnt building linke it was before...

PR #63, opened 2017-10-14
removed duplicated code by using csproj with targetframeworks

JanEggers:


Frameworks/MQTTnet.NetStandard/MQTTnet.Netstandard.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
```
JanEggers: this was what i meant with .net.sdk ^^

chkr1011:
Did you test the Build script? I usually use this only when deploying a new version. I assume the line for the NetFramework must be changed to build the other project using a different config.

JanEggers:
i updated the buildscript so the .net standard is build for both platforms, i didnt get msbuild to work and i dont know your setup as msbuild is not found on my machine because it is not on PATH

chkr1011:
Yes I usually use the VS2017 developer command line. If your approach works it is fine for me. But still one question: Isn't it required that the project is compiled two times with different configruations? Because of the #ifdef?

JanEggers:
nope dotnet always builds all frameworks used in csproj except you specifiy a framework with -f
https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-build?tabs=netcore2x

PR #66, opened 2017-10-17
Add support to not persistent Queue

JTrotta:
@chkr1011 Added new class MqttClientQueued , now persisted. To persist the queue we need file system access.
JJ


Frameworks/MQTTnet.NetStandard/MqttClientFactory.cs
```diff
@@ -9,5 +9,10 @@ public IMqttClient CreateMqttClient()
         {
             return new MqttClient(new MqttCommunicationAdapterFactory());
         }
+
+        public IMqttClient CreateMqttQueuedClient()
```
chkr1011: Please also expose this method for the factory interface.

MQTTnet.Core/Client/MqttClientQueued.cs
```diff
@@ -0,0 +1,530 @@
...
+using MQTTnet.Core.Adapter;
+using MQTTnet.Core.Diagnostics;
+using MQTTnet.Core.Exceptions;
+using MQTTnet.Core.Packets;
+using MQTTnet.Core.Protocol;
+using MQTTnet.Core.Internal;
+
+namespace MQTTnet.Core.Client
+{
+    public class MqttClientQueued : IMqttClient
```
chkr1011: Please do NOT just copy the whole existing MqttClient. Instead use it as a field inside your new client. You should also implement the IMqttClient interface and provide proxy-methods to the internal MqttClient. But still having a different implementation for publish, the receive event etc.

MQTTnet.Core/MQTTnet.Core.csproj
```diff
@@ -32,4 +32,10 @@
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Any CPU|x86'" />
 
+  <ItemGroup>
```
chkr1011: What is this used for?

MQTTnet.Core/Client/MqttClientQueued.cs
```diff
@@ -0,0 +1,530 @@
...
+    {
+        #region Fields
```
chkr1011: Please do not use regions. They are not used in the whole project.

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: That's what I would have done, but most methods are private, for example how to use AuthenticateAsync without copy it ? we nee an abstract base class, do we?

chkr1011: What is the reason for using _AuthenticateAsync_? Sorry but I don't get it.

JTrotta: It was an example. To be more precise, I need to use the SendeAndReceivePackets but its Private,

JTrotta:
Please check if it's OK now.
JJ

JTrotta:
Some little changes.
JJ

chkr1011:
Much better 😄 But I still have some doubts and requests:

1) I am considering calling the client _MqttClientManaged_ because it can have more options. For example the retry logic from the Wiki can be added to this implementation.

I also commented some things.

MQTTnet.Core/Client/MqttClientQueued.cs
```diff
@@ -0,0 +1,166 @@
...
+using MQTTnet.Core.Adapter;
+using MQTTnet.Core.Diagnostics;
+using MQTTnet.Core.Exceptions;
+using MQTTnet.Core.Packets;
+using MQTTnet.Core.Protocol;
+using MQTTnet.Core.Internal;
+
+namespace MQTTnet.Core.Client
+{
+    public class MqttClientQueued : MqttClient, IMqttClientQueued
```
chkr1011: Please do not use inheritance. One clean code principle is "composition over inheritance".  As I said please use the client as a field.

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿
+
+namespace MQTTnet.Core.Client
+{
+    public class MqttClientQueuedOptions: MqttClientTcpOptions
```
chkr1011: I am thinking about having two properties. So a connect method which has _MqttClientOptions_ and _MqttClientQueuedOptions_. The problem is you forgot the WebSocket options. To your client has limited functionality.

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        public bool UsePersistence { get; set; }
```
chkr1011: This is not required. The property with the "Storage" can be empty. Then this feature should be disabled at all. With this additional property you can set this to true even without a Storage implementation.

MQTTnet.Core/Client/MqttClientQueued.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            MqttNetTrace.Information(nameof(MqttClient), "Start sending packets.");
+            MqttApplicationMessage messageToSend = null;
+            try
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    while (_inflightQueue.TryDequeue(out messageToSend))
+                    {
+                        MqttQualityOfServiceLevel qosGroup = messageToSend.QualityOfServiceLevel;
```
chkr1011: Why not just using the "PublishAsync" method and catch all exceptions? In the catch clause I would assume that:
1) QoS 0 will be dropped
2 ) QoS 1 will be dropped
3 ) QoS 2 will be enqueued (or inserted as item 0?)

MQTTnet.Core/Client/MqttClientQueued.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        private async Task SendPackets(CancellationToken cancellationToken)
+        {
+            MqttNetTrace.Information(nameof(MqttClient), "Start sending packets.");
+            MqttApplicationMessage messageToSend = null;
+            try
+            {
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    while (_inflightQueue.TryDequeue(out messageToSend))
```
chkr1011: Is this blocking if the queue is empty or will it produce high CPU load because the while loop runs over and over.
I would prefer using a "BlockingCollection" here.

MQTTnet.Core/Client/IMqttClientQueuedStorage.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace MQTTnet.Core.Client
+{
+    public interface IMqttClientQueuedStorage
+    {
+        Task SaveInflightMessagesAsync(IList<MqttApplicationMessage> messages);
```
chkr1011: Why is this called "Inflight"? I would prefer "SaveQueuedMessagesAsync" to have it in the same way as "MqttClientQueued".

JTrotta: Ok, gonna change.

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Ok you do not like hineritance. Gonna change.

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Never used blocking collecton. Will try.

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Right. I agree

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
JTrotta: Sorry, what you mean for "I am thinking about having two properties"?
I have to create a connect method override?

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
JTrotta: Right

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Got it. It's better.

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
chkr1011: I am not 100 % sure here. But I was thinking about a overload of _ConnectAsync_ with two types of options. So two parameters like `ConnectAsync(MqttClientQueuedOptions queuedOptions, MqttClientOptions options)`. But now I see that then we will have two _ConnectAsync_ methods because the original from the interface is also there. So maybe we have to think about the overload stuff. I would not add the _Storage_ property to the original options because it is only relevant for the queued one. If you don't have a better idea I would stick to the overload stuff like:
`MqttClientTcpOptions > MqttClientOptions` and the new one to `MqttClientQueuedTcpOptions > MqttClientTcpOptions > MqttClientOptions` and the same for WebSocket options. Then you can add the storage property (and others) and also passing the options to the internal Client (which is the field) and finally having only one _ConnectAsync_ method because the MqttQueuedClient should also implement _IMqttClient_. But I just realize that it will not work correctly. I assume we need an interface for options. Then we can extend the interface easily and having different classes with matching properties. Another idea was to move this stuff to the constructor but then the pattern is different. Options are passed in the _ConnectAsync_ method. Hard topic :smile:

And please think about renaming the client to _MqttClientManaged_ because I also have some feature requests like having the automatic connection retries and options like `UseAutoReconnect : bool` and `AutoReconnectDelay : TimeStamp`. This is not related to queued stuff so my idea is to name it `Managed`. Also the options to reflect this but not the storage methods. They are only relevant for queuing stuff.

Frameworks/MQTTnet.NetStandard/Implementations/MqttCommunicationAdapterFactory.cs
```diff
@@ -21,6 +21,11 @@ public IMqttCommunicationAdapter CreateMqttCommunicationAdapter(MqttClientOption
                 return new MqttChannelCommunicationAdapter(new MqttWebSocketChannel(webSocketOptions), new MqttPacketSerializer { ProtocolVersion = options.ProtocolVersion });
             }
 
+            if (options is MqttClientQueuedOptions queuedOptions)
```
chkr1011: I want' to avoid such things (see the long comment for details below). In my opinion the new client should be separated from all other stuff of the already existing one. It should be like an extension. Like if you crate your own nuget package with this implementation having this packet only as a reference. The overall client factory is OK but other things should not be touched.

JTrotta: muble, muble, muble.
I do not agree. If I was a programmer,  I would like a complete client with all stuff like other clients in Java and C++. All other C# clients lacks some features, this one would the first with all the suff needed, and you can choose which one to use.

chkr1011: I absolutely agree here. But the question is why the base implementation must be changed. This is not clear to me. I still see the new client as an extension which adds functionality like queuing or retrying. In my opinion there is no need to change the already existing code in any way.  Also because your implementation here is simply wrong. You check for your (new) options and always return a TCP channel. So a user which requires WebSockets cannot use the queued client. There are also other places which have to be changed if you follow this strategy. This why I am saying lets treat it as a completely separate client. Then the chance is very small that we forget to support some combinations. Also the consideration or a channel (TCP, WS) should not be part of the new queued client. This should be still the job ob the low level client. The client factory is a separate story and the only place were I can accept it because this is the entry point for the users. There they can decide. Probably a unsafe cast is required here like `options as MqttClientTcpOptions` instead of the check with `is`. Then you can pass in the inherited options and it will still work. I also want to refer to the comment where I mentioned interfaces for the options. But to keep it short. You can also complete your features then I will merge it and rework some parts to fit his requirements. If it is OK for you.

JTrotta: Yes please. I lost you here: "You check for your (new) options and always return a TCP channel. So a user which requires WebSockets cannot use the queued client".
You want the queued client to be used with WS too. I agree. But you do not want to change existing code? Do not know how to do that. Moreover I changed some exisiting code transfomingh Private methds in Internal methods.
Sure I'm missing something. I'm committing the changes I understand. Would be better if you can continue with WS. I'm thinking about a better solution for Options.
But you, of course, have a better and complete vision.

chkr1011: OK for me it is OK. Let me implement this. I will merge this pull request because I still want to honor your work and ideas. So you will be shown as an official contributor. Please wait with all until I committed these changes. I will implement this today in the evening. The can start a new discussion if this fits your needs and what needs to be changed. Do you agree?

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Done

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Done

MQTTnet.Core/Client/MqttClientQueued.cs
JTrotta: Done, but to be tested more

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
JTrotta: Modified MqttClientOptions abstract class into IMqttClientOptions, but it's not a good solution.
Must be reviewed

MQTTnet.Core/Client/MqttClientQueuedOptions.cs
JTrotta: Done

JTrotta:
Client renamed

Frameworks/MQTTnet.NetStandard/Implementations/MqttCommunicationAdapterFactory.cs
JTrotta: OK!

Issue #65, opened 2017-10-17
How to Use With WebSocket Server


qcjxberin:
How to Use With WebSocket Server,Could you Write a Demo With this?

chkr1011:
Hi this is not fully implemented yet. I started a new library called _HTTPnet_ which will be used for this in the future. Implementing this is quire complex because the used HTTP library can change. There are several available for .NET. If you already use one you can have a look at the _https://github.com/chkr1011/MQTTnet/blob/master/Frameworks/MQTTnet.NetStandard/Implementations/MqttServerAdapter.cs_. There you can see how to implement the required interfaces. Then you can write a _binding_ for the HTTP library. Otherwise you have to wait until the _HTTPnet_ library is including and having a _binding_.

Best regards
Christian

ChristianRiedl:
I tried to integrate MQTTNet into a AspNetCore 2.0 web server. It was possible with some small additional classes and I changed _WebSocketStream_ class to wrap _WebSocket_ instead of _ClientWebSocket_ . I still get some exceptions when client disconnects but it works.

@chkr1011 : I think ASpNetCore 2.0 is netstandard 2.0 so it should run on latest Windows 10 Iot Core. I know it is heavy compared to  HTTPnet. Maybe you have some ideas how to improve my solution.    

Here is the code (Startup class)

```
public class Startup
{
        public void ConfigureServices(IServiceCollection services)
        {
        }

        public async void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
        {
            loggerFactory.AddConsole(LogLevel.Debug);

            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            var adapter = new MqttWebSocketServerAdapter();
            var options = new MqttServerOptions
            {
                ConnectionValidator = p =>
                {
                    if (p.ClientId == "SpecialClient")
                    {
                        if (p.Username != "USER" || p.Password != "PASS")
                        {
                            return MqttConnectReturnCode.ConnectionRefusedBadUsernameOrPassword;
                        }
                    }
                    return MqttConnectReturnCode.ConnectionAccepted;
                }
            };
            var mqttServer = new MqttServer(options, new List<IMqttServerAdapter> { adapter });
            mqttServer.ClientDisconnected += (s, e) =>
            {
                Debug.WriteLine($">> Disconnected {e.Client.ClientId}");
            };

            await mqttServer.StartAsync();
[...]

PR #70, opened 2017-10-22
initial version of ms.extensions.logging and di 

JanEggers:
wip for #65 (logging integrated no new packages)
logging should work
the clientid should be in the scope now so all logs for one client should share that scope. im not shure if server is working but client looks good

i guess some of the stuff in server options could also be moved into di.

i also need to update wiki...


JanEggers:
i cant view the guideline i broke. it says only repository owner may view so please provide feedback so i can improve

chkr1011:
Is it possible to support the static logging event? I use this in my projects. Also in the UWP app in this repo I use it to move all traces to a text box. Or can you migrate this code as well? 

We also have to update all Wiki entries after those changes are released 😄

JanEggers:
> Is it possible to support the static logging event?

shure i can create a LoggerProvider. then you can do something like:

```
//in startup code
serviceProvider.GetService<ILoggerFactory>().AddCallbackLogger();

//in your view
CallbackLoggerProvider.OnLog += (log) => {}; 

```

is that good enough?

chkr1011:
Sure something like this. I need an _easy_ way of just putting the traces somewhere.

JanEggers:
oki i will add it

chkr1011:
I also see no changes for the UniversalWindows library. Are there no changes at all or do don't migrate this project?

JanEggers:
i added back the old api for consuming logs that should simplify migration.

> I also see no changes for the UniversalWindows library

i will have a look at them.  (I dont have that projects loaded because uwp sdk does not work for some reason.)

chkr1011:
The problem is that some concepts of the core library are designed in the way they are to support UWP as well. Beside adapters and channel not that much but I am afraid that the new logging may break the UWP implementation.

JanEggers:
uwp projects should also be up and running (at least they build but i cannot start em on my machine)

how does the github flow for wikis work i never updated one?

chkr1011:
I gave you access as a contributor to this project. You can now create branches in this repository. But merging from develop to master is still my part 😄

chkr1011:
I assume you can directly update the Wiki via "Edit" and commit the changes...

chkr1011:
The new trace approach works for the UWP app. Visually no difference but I cannot set the client ID anymore. This was a new requested feature. Using the factory I cannot pass an identifier which let the user distinguish between several clients running in the same process. Or did I make a mistake while setting up the logging?

JanEggers:
the information is there i create a LogScope for each client with its clientid. I was not able to port that Information to MqttTrace see https://github.com/chkr1011/MQTTnet/blob/3529724bdc91f6591fa91ca3c49c38b31a70e4e1/MQTTnet.Core/Diagnostics/MqttNetLogger.cs#L37

the information is stored in callcontext so it flows to all loggers in that scope. 
it should be visibile if you use serilog or ms.extensions.logging.console.

if you know how to port this let me know! i tried to view history of that file so i get the 1.1.2 version but i didnt get it compiling.
https://github.com/aspnet/Logging/blob/a42605cc0c51b36ac3dda10401e7ffea05c8d848/src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs#L22

chkr1011:
Ah there it is. OK. My initial implementation was detached from such protocol details. There you can specify a separate identifier. The use case was if a have 2 clients (also with the same ID) which are bridging two brokers I cannot distinguish. Is there any way to inject a separate ID like before with the factory constructor?

JanEggers:
Adding a seperate property to MqttNetLogger would be possible but all other Libraries that interact with ILogger dont know it and so they wont use it, thats why i put it in the Scope. 

If a separate Id is required we can add it to the options and create the scope with that Id.

Using a factory with parameter would also be possible.

chkr1011:
Mhh it's up to you 😄 I would prefer the options with an additional _LogId_ which falls back to the clientId if not set. Because this was a user request I also can only guess what makes sense. 😄

PR #75, opened 2017-10-24
Mqttnet.aspnetcore

JanEggers:
for now just the code of the aspnetcore package.  if you are ok with the code i can add nuspec and update build script


Tests/MQTTnet.TestApp.AspNetCore2/wwwroot/app/app.ts
```diff
@@ -0,0 +1,20 @@
+﻿import { connect } from "mqtt";
```
chkr1011: Please have look at the test folder. There is also an old JS file which I used for testing. Can you may adapt the functionality and then remove it? Otherwise the gitignore must be changed to still allow JS files. But I would prefer this file. I also want to start learning type script 😄

JanEggers: oki will do

JanEggers: i cannot find any js files can you give me a link

chkr1011: Oh I forgot that I moved it to HTTPnet repo because there I implemented the WebSocket protocol. OK then lets stick to the existing one.

Do you add some buttons to the new one to make it a little bit interactive? It can be ugly for sure but I need those script for testing HTTPnet later on.

JanEggers: i added some ui code you can now send msgs from html and the received msgs are displayed

chkr1011:
Again good work! 👍

PR #81, opened 2017-10-25
Dynamic/Provider Password

JTrotta:
Christian,
may be I miss something  from your reasoning.
The test I did, need to change the Password setter from only GET to GET and SET, otherwise How can I change it dynamically?
Have  a check to the PR and let me know. (now it does not work).


MQTTnet.Core/ManagedClient/ManagedMqttClient.cs
```diff
@@ -277,6 +277,7 @@ await Task.Delay(100, _connectionCancellationToken.Token).ConfigureAwait(false);
 
             try
             {
+                _options.PasswordProvider?.Invoke(_options);
```
chkr1011: Sorry but I don't understand this. The result is not used here. I assume you need to "copy" the options for the internal client here any replace the password. At the moment I don't see a better approach.

JTrotta: In fact. Cannot be done. Password property is read-only.
I'll check for better solution.
May I ask why Interface should be read-only?

chkr1011: I mean a copy. So a new instance will all properties copied. Then you also have the setter. But let me try. I have some ideas. It is read-only because the library only needs read access to the password. There is no need to write it. What you need is a special case which must be solved in another way. The client should not manipulate user data like options or messages.

JTrotta: I see. Your "Credentials" solution should work.
In the meanwhile, I'm testing persistence and re connection on MONO since two days: it works very well!

chkr1011: Cool. Also the persistence? I was not able to test it. No messages lost and is the order also preserved?

JTrotta: Yes persistence seems perfect. Will let you know next week, we are applying many tests. Order it's not preserved right now. But I have some ideas about that.
This client is COOL!

chkr1011: :-) OK the message is append at the end if one send fails. But beside of that it should be OK or? What is the project you are working on? Can you tell me something?

JTrotta: Yes if course. But, if you don't mind, I prefer to continue to a private conversation,.

chkr1011: Sure

Issue #82, opened 2017-10-25
Release 2.5


chkr1011:
I want to release a new version soon because we have lots of new features. Before doing this I want to summarize the open topics:

1. Units Tests are not running for server tests. I am not able to fix them because I don't understand how to inject the test adapter since we have DI.  @JanEggers Can you please have a look?
2. @haeberle can you confirm that the bug with UWP is now fixed since latest changes from @JanEggers? If not I can also keep searching for a solution.
3. @JTrotta does the _ManagedClient_ have everything you need so far? We can add more features but I want to release it soon with the features it already has.

haeberle:
Not still the same issue in UWP, cureently I'm testing from version 2.1.3. to 2.3.1, so fare all fine, the issues must be in a newer version. I will test next versions asap.

JTrotta:
Let me do last test this morning. (1000 utc). But it should be almost ready and working.

JanEggers:
@chkr1011  i created a pr that fixes the unittests, addresses your todos according configuring the server and added nuspec for the new package

JTrotta:
@chkr1011 Got some problem.
Cannot instantiate ManagedClient without IOC. 
Now you added Microsoft IOC logging extensions. But I'm already using TinyIoc, and I do not like to add more and more libraries.
Is it possible to avoid this construction mode?

chkr1011:
@JTrotta Sorry but the introduction of the MS logging and also DI was already some time on the list. Already since two versions.
In general this library is only supported by 4.5.2 and above because older .net frameworks are no longer supported by MS. Mono is surely a different story. Maybe @JanEggers can shed some light here how to restore support for Mono etc. But you can already stay at your logging library and just add it to the logging infrastructure like the still existing static event. You can also still use the static trace event and push the messages into your logging component.

JTrotta:
Ok,
gonna try 4.5.2 on MONO. Please can you wait till 1600 UTC? Otherwise I have to fork it ti 451.
Thank you

chkr1011:
Don't worry. I believe we will find a solution.

JanEggers:
regarding ioc:
have a look at Frameworks/MQTTnet.NetStandard/ServiceCollectionExtensions.cs
[...]

PR #90, opened 2017-11-06
added dependencies to nuspec and fixed build

JanEggers:
fixes build for #82

### DIFF
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -34,10 +34,14 @@
       <group targetFramework="netstandard1.3">
         <dependency id="NETStandard.Library" version="1.6.1" />
         <dependency id="System.Net.Security" version="4.3.2" />
+        <dependency id="Microsoft.Extensions.DependencyInjection" version="1.1.1" />
+        <dependency id="Microsoft.Extensions.Logging" version="1.1.2" />
       </group>
       
       <group targetFramework="uap10.0">
         <dependency id="Microsoft.NETCore.UniversalWindowsPlatform" version="5.4.0" />
+        <dependency id="Microsoft.Extensions.DependencyInjection" version="1.1.1" />
+        <dependency id="Microsoft.Extensions.Logging" version="1.1.2" />
       </group>
 
     </dependencies>
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -48,6 +48,12 @@
         <dependency id="Microsoft.Extensions.Logging" version="1.1.2" />
         <dependency id="Microsoft.Extensions.Options" version="1.1.2" />
       </group>
+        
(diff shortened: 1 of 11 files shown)


chkr1011:
Isn't it required to add all libs? I added:
```
      <group targetFramework="netstandard1.3">
        <dependency id="NETStandard.Library" version="1.6.1" />
        
        <dependency id="Microsoft.Extensions.DependencyInjection" version="1.1.1" />
        <dependency id="Microsoft.Extensions.Logging" version="1.1.2" />
        <dependency id="Microsoft.Extensions.Logging.Abstractions" version="1.1.2" />
        <dependency id="Microsoft.Extensions.Options" version="1.1.2" />
        
        <dependency id="System.Net.Security" version="4.3.2" />
        <dependency id="System.Net.WebSockets" version="4.3.0" />
        <dependency id="System.Net.WebSockets.Client" version="4.3.1" />
        <dependency id="System.Threading" version="4.3.0" />
        <dependency id="System.Threading.Thread" version="4.3.0" />
      </group>
```

But I am not sure. I just added every lib which is references by the projects. Even if Options i.e. is a sub lib of the Logger one. 

I am also not sure if a group for targeting regular net framework is required. Because the DI stuff is not part of it.

JanEggers:
i added the missing packages and verified with a console app.

image

> Even if Options i.e. is a sub lib of the Logger one.
no i just forgot options.

> I am also not sure if a group for targeting regular net framework is required. Because the DI stuff is not part of it.

net461 means it runs on net461 "Platform" but all packages not present in net461 must be delivered with the app. that is why system.* are not included in net461 but ms.extensions.di and logging are not present in net461 so they must be deployed with your app.

my app is ~30 mb deployment because of all the packages of aspnetcore and ms.extensions.* bring in my own assemblies are 100KB....

JanEggers:
btw: should i check why the test is failing?

JanEggers:
gaaa i explain it and did not fix it give me a second

JanEggers:
now it should be good

chkr1011:
So does it mean I have to add the extension libs on my own if I use NET461?

Yes please check the test. I just saw it too but I have no time yet to look at it. In general it looks good so it seems some internal problem.

JanEggers:
i fixed the test and maybe introduced a new issue. i had to refactor retention manager so it comes from di so i get get it for testing. but for that i had to make it a singleton.

that implies if someone wants to start 2 servers he should create a serviceprovider for each one otherwise the singletons will overlap.

we can fix that if we make the retentionmanager scoped and begin a new scope in each server but i dont know if that is a reasonable change what do you think?

chkr1011:
I think two servers in one process is not a use case at the moment. If so, we can change it after a request was made. I don't see any use case at the moment. Across processes OK but not in one process.

As usual, good job! 👍

Issue #93, opened 2017-11-09
Microsoft.Extensions.DependencyInjection 2.0.0 breaks MQTTnet


ClaudioTF:
Package dependencies specify Microsoft.Extensions.DependencyInjection v1.1.1 or greater.  But updating to Microsoft.Extensions.DependencyInjection 2.0.0 breaks MQTTnet.  This statement

            var mqttServer = new MqttFactory().CreateMqttServer();

reproduces it.

chkr1011:
Hi,
thank you for sharing this issue. If you use the DI library in version 2.0.0 in your project please use the following code for constructing a MQTT server:
```csharp
var services = new ServiceCollection()
	.AddMqttServer(options =>
	{
	// modify options here
	})
	.AddLogging()
	.BuildServiceProvider();

var mqttServer = services.GetRequiredService<IMqttServer>();
```

@JanEggers Do we havy any chance to fix this? I found several other tickets for other projects which have the same problem because 2.0.0 of DI has a breaking change (The service provider is no longer returned as an interface).

Best regards
Christian

JanEggers:
i will update myself then i will get that same error

chkr1011:
Here is an interesting post from people having the same problem: https://github.com/OData/WebApi/pull/1082

JanEggers:
> Do we havy any chance to fix this?

shure we can add netstadard2.0 build with updated dependencies. pr incoming

chkr1011:
Well for me this not a solution because (as also discussed at the other projects which are affected) this will require all library users to use .net standard 2.0 including ASP.NET Core 2.0 and .NET 4.6+. But my goal is to still support .NET 4.5.2 and to not rely on .net core (standard) 2.0 because it is just released and I will not limit the library to only be usable by cutting edge technologies only. 

I am trying to use reflection for service lookup and using _object_ as the type for the service provider field. I will let you know if this works.

JanEggers:
no sry i didnt put it correctly we can add another targetplatform with updated dependencies but still maintain netstandard 1.3

JanEggers:
im currently working on 
    <TargetFrameworks>netstandard1.3;netstandard2.0;net452;uap10.0</TargetFrameworks>

chkr1011:
Ah OK sorry I don't get that. This is surely OK for me 😄

JanEggers:
i also added a new event for server started so there are no messages published in aspnetcore sample until the server is started

[...]

PR #94, opened 2017-11-09
Develop

JanEggers:
like discussed in #93

### DIFF
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -41,6 +41,17 @@
         <dependency id="System.Net.WebSockets.Client" version="4.3.1" />
         <dependency id="System.Threading.Thread" version="4.3.0" />
       </group>
+
+      <group targetFramework="netstandard2.0">
+        <dependency id="NETStandard.Library" version="2.0.0" />
+        <dependency id="Microsoft.Extensions.DependencyInjection" version="2.0.0" />
+        <dependency id="Microsoft.Extensions.Logging" version="2.0.0" />
+        <dependency id="Microsoft.Extensions.Options" version="2.0.0" />
+        <dependency id="System.Net.Security" version="4.3.2" />
+        <dependency id="System.Net.WebSockets" version="4.3.0" />
+        <dependency id="System.Net.WebSockets.Client" version="4.3.1" />
+        <dependency id="System.Threading.Thread" version="4.3.0" />
+      </group>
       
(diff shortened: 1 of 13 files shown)


chkr1011:
1. Is assume also some changes for the factory using some #if are required because the interface is no longer used but the service provider class itself. So the parameters must be changed too. Or am I wrong and miss something?
2. Started event is a good idea! Can you please check my Wiki entry if it makes sense with the ASP.NET Core integration? And update if required?

JanEggers:
> 1. 
no api is compatibe as we dont use the concreate return type but just System.IServiceProvider

>2. will do.

im also looking at the failing tests, the one that remains is a similar timing issue but i need access to the clientSession in order to verify presence of the WillMessage

chkr1011:
Ah OK because I had to change it in my code. But anyway we will see 😄 
Yes I already saw it but I don't know why they fail. I saw that an exception is thrown which indicates that a port is already used by the system but the adapter is the test adapter only. I also tried to debug the failing tests with no luck.

JanEggers:
so issue should be fixed now if you have a socket in use issue and dont know who is it use tcpview from sysinternals or netstat

chkr1011:
I am wondering WHY the unit test want's to open a socket. Due to the test server adapter there should be no attempt to open one event if it is not used?!

chkr1011:
Maybe you didn't noticed but I added you as a official contributor. You can create branches in this repository if you like.

JanEggers:
> I am wondering WHY the unit test want's to open a socket. Due to the test server adapter there should be no attempt to open one event if it is not used?!

because i just add the testadapter to the list of adapters if tcp listener should not be present we have to remove it from the service collection

chkr1011:
Ah that makes sense. How can it be removed?

You can still create pull request for merges. It is up to you to decide what you want to use 😄

JanEggers:
my usual approch is adding this

```
        public static IServiceCollection RemoveAll<T>(this ServiceCollection services)
        {
            var toRemove = services.Where(s => s.ServiceType == typeof(T)).ToList();

            foreach (var serviceDescriptor in toRemove)
            {
                services.Remove(serviceDescriptor);
            }

            return services;
        }

         var services = new ServiceCollection()
                .AddLogging()
                .AddMqttServer()
                .RemoveAll<IMqttServerAdapter>()
                .AddSingleton<IMqttServerAdapter>(serverAdapter)
                .BuildServiceProvider();
```

Issue #100, opened 2017-11-11
MQTTnet targeting .NET SW 4.6.1 has many dependencies


KarloX2:
Hi!

I read issue #80 
but I'm still not sure if I get this right. Foremost, I must admit I have no deep understanding of how .NET Standard / Core really works.

I target .NET Framework 4.6.1 and by installing MQTTNet via NuGet it comes up with a long list of dependencies, amongst which are animals like "System.Console.dll" and "System.Net.Sockets.dll" that seem to implement things that are already contained in FW 4.6.1, at least very similar.

First of all: What exactly is this? It must be related to .NET Standard somehow, but AFAIK ".NET Standard" is just a set of rules, not an implementation. I somewhere read that .NET FW 4.6.1 is an implementation of that standard up to some version number. So are these new assemblies then part of .NET Core?

And secondly: Do I really need all this?

Thanks!

chkr1011:
In my opinion this is correct. We specified that only for .NET 4.5.2 the .net framework shall be used in the nuget package. For other (higher) version the .net standard is used and thus several libraries are included because Microsoft has split the .NET framework into lots separate libraries.

But I get you concern. We will review the nuget package definition.

@JanEggers I assume we have to add every version of the .net framework (like 4.6, 4.6.1, 4.6.2 etc.) in the nuget package and use the binary from 4.5.2. Do you think this is a suitable way?

JanEggers:
im just creating a new net461 console app to verify what is wrong first

JanEggers:
@chkr1011 i cannot figure it out i added net461 to mqttnet.netstandard and testapp but still there are some more dlls in the testapp output of net461. 

so i added a issue to dotnetstandard so maybe they can shed some light on why this is happening:
https://github.com/dotnet/standard/issues/580

net452 is 1.4 MB and net462 is 2.2 MB.

@KarloX2 what are you concerned about when taking this dependencies?

- if you zip that folder its 900 KB more for deployment
- you fear the packages are not maintained? 

+ the packages are all maintained by microsoft.
+ support for dependency injection
+ common logging abstraction with a whole ecosystem to log to console / file / eventlag / db / whatever  

for me the good parts outweight the bad parts.

KarloX2:
@JanEggers my biggest concern is that there are things going on with all that libs that I don't understand.

Personally I wouldn't care too much, but in the project team there are people who will ask questions about why suddenly there's a need to put so many elements on top of the existing .Net FW.

[...]

PR #101, opened 2017-11-12
Targetframeworkfix

JanEggers:
this pr should improve #100

### DIFF
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -18,10 +18,10 @@
     <dependencies>
 
       <group targetFramework="netstandard1.3">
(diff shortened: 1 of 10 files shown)


chkr1011:
Thanks for fixing that issue. I will wait a couple of days and also fix two exceptions I found. Then I will release it as 2.5.2. Do you agree?

JanEggers:
yup agreed. i still have the issue open in dotnetstandard repo because i still dont fully understand whats going on.

JanEggers:
maybee we have to remove the nuget config again as it is causing appveyor to fail

chkr1011:
Mhh strange thing...

PR #102, opened 2017-11-12
added net461 build

JanEggers:


JanEggers:
build is green again. @chkr1011  what do you think about building / publishing prerelease nugets from the dev branch so we and the consumers with issues can verify before releasing the next version?

chkr1011:
Yes after this disaster I agree 😄

PR #104, opened 2017-11-12
fixed websocket subprotocol negotiation

### LINKED ISSUE #103: MqttNet client using WebSocket fails if not using .NET Framework
There must be some difference between ClientWebSocket implementation in .NET 452 and .NET Core or UWP. I am using latest MQTTnet 2.5.1.

I tried to test the **MQTTNet.TestApp.NetCore ClientTest** against **MQTTNet.TestApp.AspNetCore2**.   
When I run the AspNetCore2 Server via IISExpress and _localhost:54191_ both net452 and netcoreapp2.0 ClientTest failed. 

Than I run AspNetCore2 using Kestrel server and _192.168.1.140:54191_. Now the net452 ClientTest is running perfect but netcoreapp2.0 still gets a WebSocketException (0x80004005 : "Unable to connect to the remote server") during WebSocket connect..

I created a UWP app similar to ClientTest and got the same exception during WebSocket connect.

My real environment is an AspNetCore 2.0 (net462) server using http.sys. My Windows Forms (4.6.2) mqtt test app is running perfect against this server.  But trying the  UWP test app shows a different behavior : Connect and subscribe succeeds, client got several application messages, but after approx 30 seconds I got a mqtt timeout exception because server did not answer the clients Ping request.

JanEggers:
fixes #103

### DIFF
--- a/Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
+++ b/Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
@@ -2,6 +2,7 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.DependencyInjection;
 using MQTTnet.Core.Server;
+using System.Linq;
 
 namespace MQTTnet.AspNetCore
 {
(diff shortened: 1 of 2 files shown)


Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
```diff
@@ -14,8 +15,18 @@ public static IApplicationBuilder UseMqttEndpoint(this IApplicationBuilder app,
             {
                 if (context.Request.Path == path && context.WebSockets.IsWebSocketRequest)
                 {
+                    string subprotocol = null;
+
+                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues)
+                     && requestedSubProtocolValues.Count > 0
```
chkr1011: Can you explain why this fixes the issue? I am no expert about websockets etc. so I can only guess 😄 Especially why this fails for UWP and netcore only!?

JanEggers: it was just the part of the issue he has with establishing a connection. i just guessed this was the issue because i changed it when adding the test website. i verified with the testapp and voila problem solved. 

but that was just part of the issue the communicationtimeout issue persists.

PR #134, opened 2017-12-04
Serialized certificates loaded as X509Certificate2

cvellan:
Instances of X509Certificate2 passed into an MqttClientOptionsBuilder lose data and functionality by being loaded as an X509Certificate, most notably the private key which servers may require for client validation.

To the best of my knowledge, instances of X509Certificate do not lose data or functionality by being loaded as an X509Certificate2.


chkr1011:
Thanks for you contribution. Good work 👍

PR #157, opened 2018-01-17
Use asynchronous methods for blocking operations

kpreisser:
Hi,

we are currently evaluating MQTTnet for use within our software. What I like about this library is the use of async operations, so that it does not need to use separate threads for each client, which is important for scalability.

However, while testing the server, I found that while a lot of send/receive operations are async, there is still code that
1) does a blocking read from the client to receive a new frame, and
2) does a blocking wait for the next packet to send to the client.

For example, if you start the TestApp and start a server and then connect two clients, then if you now pause the server execution, you can see that 4 worker threads are waiting in blocking calls to `BlockingCollection.Take()` (in `MqttClientPendingMessagesQueue.SendQueuedPacketAsync()`) and in `Stream.ReadByte()` (in `MqttPacketReader.ReadHeaderFromSource()`.

However, if a lot of worker threads are blocked, new worker threads are only created slowly, so this could lead to a performance problem (waiting time) when many clients are connected.

This can be fixed by changing these operations to be async, so that they match with the other socket read/write operations that are already async, and therefore do not block worker threads.

Thanks!


Frameworks/MQTTnet.NetStandard/Serializer/MqttPacketReader.cs
```diff
@@ -80,13 +81,15 @@ public byte[] ReadRemainingData()
             return ReadBytes(_header.BodyLength - (int)BaseStream.Position);
         }
 
-        private static int ReadBodyLengthFromSource(Stream stream, CancellationToken cancellationToken)
+        private static async Task<int> ReadBodyLengthFromSourceAsync(Stream stream, CancellationToken cancellationToken)
```
chkr1011: Did you test this change with the performance test app? I agree that the readByte method should be changed because it is not async but reading the body length should not be a problem. If a previous header was received (1 byte) there should be more data and thus there is no need to use async because reading single bytes all the time async will decrease the performance a lot. But as I said the other place with the header makes sense because the thread is being blocked (instead of freed) until the next message is received.

Frameworks/MQTTnet.NetStandard/Server/MqttClientPendingMessagesQueue.cs
```diff
@@ -12,7 +12,8 @@ namespace MQTTnet.Server
 {
     public sealed class MqttClientPendingMessagesQueue : IDisposable
     {
-        private readonly BlockingCollection<MqttBasePacket> _queue = new BlockingCollection<MqttBasePacket>();
+        private readonly ConcurrentQueue<MqttBasePacket> _queue = new ConcurrentQueue<MqttBasePacket>();
+        private readonly SemaphoreSlim _queueWaitSemaphore = new SemaphoreSlim(0);
```
chkr1011: Is 0 correct here? I also used the Semaphore for several places but I had to use a different overload.

chkr1011:
I will approve this changes for now but we should still discuss my considerations. In the meantime I will compare the performance etc...

Frameworks/MQTTnet.NetStandard/Serializer/MqttPacketReader.cs
kpreisser: I did not yet test the performance. Note however, that in TCP/IP, data can be split at any location (as TCP is steam-like), so even if the first byte arrived (for a new packet), it does not mean that the second, third etc. bytes are also already available. For example, a slow client might send the first byte, then wait some time before sending the remaining bytes; or a interceptor might split the bytes at this boundary etc.

Therefore, if the first byte is read with an async method but the remaining length bytes are read using a sync method, there is still a possibility that worker threads are blocked at this location if many clients do this.

Alternatively, I think we could use [`NetworkStream.DataAvailable`](https://docs.microsoft.com/de-de/dotnet/api/system.net.sockets.networkstream.dataavailable?view=netframework-4.7#System_Net_Sockets_NetworkStream_DataAvailable) before either using an async read method (if no data available) or a synchronous read method (if data is available), if using the async read method only has a negative performance impact.

Frameworks/MQTTnet.NetStandard/Server/MqttClientPendingMessagesQueue.cs
kpreisser: According to the docs, `new SemaphoreSlim(0)` means the initial count of the SemaphoreSlim is 0, but there is no maximum count. Internally, .NET Framework uses `int32.MaxValue` for this.

The semaphore in this case represents the number of packets that are currently in the queue, which can be more than 1, so I think this should be correct.

kpreisser:
Hi @chkr1011,

thanks for the review! I added some comments.

PR #169, opened 2018-01-26
Optimize receiving the packet body

kpreisser:
Hi Christian,

I saw that you created commit 1415e4b878488a62741bb9be98e39a36365b6141 as follow-up to #166. Thank you!

However, I noticed there were some problems in the new code:
- When calling `new MemoryStream(byte[])`, its `Length` property is set to the size of the array, so the code would probably need to call `.SetLength(0)` to reset the length.
- The code for receiving into the buffer always used `buffer.Length` as count even if `header.BodyLength - body.Length < buffer.Length`, which meant it would receive too many bytes for a packet.

I think it can be optimized to only create a MemoryStream if the body length is actually greater than the read buffer length, and I think that `ReceivedMqttPacket` does not need to hold the `MemoryStream`, but instead could hold the MemoryStream's buffer as `ArraySegment<byte>`, so it does not need to be disposable, as shown in the PR.

What do you think?

Thanks!

### DIFF
--- a/Frameworks/MQTTnet.NetStandard/Adapter/MqttChannelAdapter.cs
+++ b/Frameworks/MQTTnet.NetStandard/Adapter/MqttChannelAdapter.cs
@@ -17,7 +17,10 @@ namespace MQTTnet.Adapter
     public class MqttChannelAdapter : IMqttChannelAdapter
     {
         private const uint ErrorOperationAborted = 0x800703E3;
+
         private const int ReadBufferSize = 4096;  // TODO: Move buffer size to config
+
+        private static readonly ArraySegment<byte> EmptyBody = new ArraySegment<byte>(new byte[0]);
 
         private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
         private readonly IMqttNetLogger _logger;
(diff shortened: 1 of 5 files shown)


chkr1011:
Hi @kpreisser,
thank you for your review and suggestions!

I don't get the first point. I don't see an issue. What do you mean with "probably" calling _.SetLength(0)_. If found no code where this happens.

I agree at the second point and fixed it. 

I would keep the apporach with the memory stream because the packet is deserialized from a stream and in your code I saw lots of conversions like _MemoryStream or byte[] -> ArraySegment -> MemoryStream_. Directly using the MemoryStream from the beginning is in my opinion a more cleaner solution. Also disposing the ReceivedPacket is only required in one small place. 

What do you think?

Best regards
Christian

kpreisser:
Hi,

thanks for your comments and for fixing the second issue!

Regarding my first point, consider the following code in `MqttChannelAdapter.ReceiveAsync()`:
```c#
    var body = header.BodyLength <= ReadBufferSize ? new MemoryStream(new byte[header.BodyLength]) : new MemoryStream();
            
    var buffer = new byte[ReadBufferSize];
    while (body.Length < header.BodyLength)
    // ...
```

For example, if `header.BodyLength` is 100, the constructor `new MemoryStream(new byte[100])` will be called. However, `body.Length` now returns 100 (because when creating a MemoryStream with a byte array, its length will be the array's length). This means the loop will not read any data as the condition `body.Length < header.BodyLength` will not be satisfied.

I think it will need to be changed to:
```c#
    var body = header.BodyLength <= ReadBufferSize ? new MemoryStream(new byte[header.BodyLength]) : new MemoryStream();
    // Reset the stream length
    body.SetLength(0);
```

to ensure the MemoryStream's length is reset to 0 before starting to read data.

I think keeping the `MemoryStream` is OK, I'm therefore closing the PR.

Note that the documentation about MemoryStream says:
> This type implements the IDisposable interface, but does not actually have any resources to dispose. This means that disposing it by directly calling Dispose() or by using a language construct such as using (in C#) or Using (in Visual Basic) is not necessary.

So in theory, `MqttReceivedPacket` would not need to implement `IDisposable` if it uses a MemoryStream.

Thank you!

chkr1011:
Ah OK now I got it :smile:. But isn't his the property called _Capacity_ what you mean? In my opinion _Length_ is still _0_ but _Capacity_ is 100. Since all UnitTests are green I would say it is 😄 

Yes I know that Dispose is not required but I added it because of the Disposable pattern...

kpreisser:
Hi,

actually `Length` is not 0 in this case. I noticed the issue when I was debugging the new code and found that the server didn't receive any useful packet from the client (`[Error]: Protocol name is not supported.`) :wink:

E.g. if you run the following code:
```c#
            var ms = new MemoryStream(new byte[100]);
            Console.WriteLine(ms.Length);
```
It will print `100` to the console, which means the length is set to the current length of the byte array.

chkr1011:
OK you are right. I fixed the code :smile:

PR #180, opened 2018-02-20
Fix ManagedMqttClient missing call to unsubscribe topics

lerppana:
This fixes ManagedMqttClient never unsubscribing from given topics.


chkr1011:
Hi @lerppana,
good work!

Best regards
Christian

lerppana:
Hi Christian!

I saw you refactored my code a bit before merging. It seems to be that SubscribeAsync throws exception if it is given empty parameters:

> "At least one topic filter must be set [MQTT-3.10.3-2]."

When code reaches line 323 in ManagedMqttClient.cs with empty subscriptions list and filled unsubscriptions list, first call to SubscribeAsync will be caught and line 323 will not be reached after this. So there is high possibility for unsubscriptions list to be dropped. This can be quickly fixed by checking the length of lists before calling proper methods.

Br,
Lassi

chkr1011:
Yes you are right. I will add checks for it. I am also thinking about a different solution because if an error raises the list of unsubscriptions is already cleared. So it will not unsubscribe later. One solution is to keep track of all unsubscriptions or only clearing them after success but this will block the internal collection for a long time. For now I added an event which is fired and the user can readd all unsubscriptions. What do you think?

lerppana:
Yes it would be best for client to get some kind of feedback that unsubscription has actually happened, so there would be more control for client side development. Providing event based feedback allows client to have better control over making sure that right topics are subscribed to. 👍 Surely it is not ideal to block server for too long if any error conditions happen.

PR #215, opened 2018-04-09
Update Startup.cs

biapar:
Add using MQTTnet;


Tests/MQTTnet.TestApp.AspNetCore2/Startup.cs
```diff
@@ -5,6 +5,7 @@
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.FileProviders;
+using MQTTnet;
```
chkr1011: Why did you add this? On my machine this is not required. Or do I miss something?

biapar:
It's not possible to resolve **MqttApplicationMessageBuilder** without it.

 var msg = new MqttApplicationMessageBuilder()
                        .WithPayload("Mqtt is awesome")
                        .WithTopic(topic)
                        .WithAtLeastOnceQoS();

biapar:
-:)

PR #220, opened 2018-04-14
Replace double dictionary with a tuple key

israellot:
Small PR replacing a double dictionary with a single tuple key-ed dictionary
Also replaced one timeout task with a linked cancellation token


israellot:
Maybe a rework should be done on that. A big problem I see with previous approach and the TimeoutAfter in general is that in case of a timeout, the actual work task continues to run and is not cancelled, which may not be the desired outcome. Using cancellation tokens we ensure the task is signaled to stop, so it's a clear benefit. The way I see, we could catch the token exception and throw a timeout instead, but we would lose the stack trace probably.

PR #223, opened 2018-04-16
Optmizations

israellot:
This PR improves throughput by 2x by disabling nagle algorithm on sockets. 
As many MQTT packets can fit into a single packet, we can avoid sockets lingering waiting for more bytes and send packets right away. 
I also refactored the packet serialization to construct a buffer with the header in the right position so we don't need to write to the socket twice per packet, which would cause two tcp packets to be sent over the wire for each mqtt packet. Doing that also allowed me to remove a semaphore that could bottleneck access to the socket.


israellot:
Sorry, faulty PR. Will send another one shortly

PR #225, opened 2018-04-16
Optmizations

israellot:
This PR improves throughput by 2x by disabling nagle algorithm on sockets.
As many MQTT packets can fit into a single tcp packet, we can avoid sockets lingering waiting for more bytes and send packets right away.
I also refactored the packet serialization to construct a buffer with the header in the right position so we don't need to write to the socket twice per packet, which would cause two tcp packets to be sent over the wire for each mqtt packet. Doing that also allowed me to remove a semaphore that could bottleneck access to the socket.

### DIFF
--- a/Frameworks/MQTTnet.NetStandard/Adapter/IMqttChannelAdapter.cs
+++ b/Frameworks/MQTTnet.NetStandard/Adapter/IMqttChannelAdapter.cs
@@ -15,7 +15,7 @@ namespace MQTTnet.Adapter
 
         Task DisconnectAsync(TimeSpan timeout);
 
-        Task SendPacketsAsync(TimeSpan timeout, CancellationToken cancellationToken, IEnumerable<MqttBasePacket> packets);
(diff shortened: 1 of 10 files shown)


israellot:
https://github.com/chkr1011/MQTTnet/pull/220 is also dealt with, by setting the original exception as a inner exception for timeout.

chkr1011:
Thanks 😄 I refactored the _develop_ branch and fixed the UWP build (which requires the additional _flush_). Please let me know if this version still works for you. My performance tests results in around +20.000 messages per second. Thats awesome 👍

israellot:
How are you testing performance? 
I was able to roundtrip 40k Qos 0  msgs/sec , but that number will vary a lot for each test approach.

israellot:
I wrote a little test where i have clients pushing messages and a client listening for them, so I can measure the whole roundtrip. Maybe I'll include that in your test project.

PR #226, opened 2018-04-18
Remove useless sempahore object

israellot:
Very small PR, just removing an orphaned semaphore.

I see you've been working on the socket handling. Good job again. 
I believe if we want to further improve network throughput on server we should change how we read and write sockets. Most people will agree that the read() approach is sub-optimal. 
There are a couple of solutions, like SocketAsyncEventArgs and RIO, but maybe using DotNetty could settle the issue. What do you think?

### DIFF
--- a/MQTTnet.sln
+++ b/MQTTnet.sln
@@ -39,9 +39,6 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MQTTnet.Extensions.Rpc", "Extensions\MQTTnet.Extensions.Rpc\MQTTnet.Extensions.Rpc.csproj", "{C444E9C8-95FA-430E-9126-274129DE16CD}"
 EndProject
 Global
-	GlobalSection(Performance) = preSolution
-		HasPerformanceSessions = true
-	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Debug|ARM = Debug|ARM
--- a/MQTTnet.sln
+++ b/MQTTnet.sln
@@ -39,6 +39,9 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MQTTnet.Extensions.Rpc", "Extensions\MQTTnet.Extensions.Rpc\MQTTnet.Extensions.Rpc.csproj", "{C444E9C8-95FA-430E-9126-274129DE16CD}"
 EndProject
 Global
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
 		Debug|ARM = Debug|ARM
--- a/Extensions/MQTTnet.Extensions.Rpc/MqttRpcClient.cs
+++ b/Extensions/MQTTnet.Extensions.Rpc/MqttRpcClient.cs
@@ -1,9 +1,9 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using MQTTnet.Client;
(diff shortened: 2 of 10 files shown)


MQTTnet.sln
```diff
@@ -39,6 +39,9 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MQTTnet.Extensions.Rpc", "Extensions\MQTTnet.Extensions.Rpc\MQTTnet.Extensions.Rpc.csproj", "{C444E9C8-95FA-430E-9126-274129DE16CD}"
 EndProject
 Global
+	GlobalSection(Performance) = preSolution
+		HasPerformanceSessions = true
+	EndGlobalSection
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
```
chkr1011: Is this required? I assume it is a "local" setting for your environment!?

chkr1011:
I don't want to introduce 3rd party libraries in this library. So DotNetty is not an option for me.

I saw SocketAsyncEventArgs  but never go into detail. Do you have some? What do you mean with RIO? 
But is there a more lower API than writing to a Networkstream?

chkr1011:
I also found a solution for the _TimeoutAfter_ code. Please have a look and let me know what do you think.

israellot:
RIO is aka Registered I/O http://www.serverframework.com/asynchronousevents/2011/10/windows-8-registered-io-networking-extensions.html
But it's windows only, so it's a no go as well. I also came to known that .net sockets are migrating to SocketAsyncEventArgs  under the hood already on dotnet core. 

The great challenge on server applications is to move data quickly from a lot of sockets into the application code. Queuing tasks will end up in a lot of context switches. Libraries like Netty or ZeroMQ usually address that problem by pooling multiple sockets on a single thread, thus reducing context switching. DotNetty is extensively used  inside Microsoft and is the base networking layer for their IoT hub for example, I think it's worth a try. Or maybe even as a separate package that can be plugged in, optionally by users. It's most useful on the server. 

I guess the most obvious step right now is optimize the packet parsing. I'm writing a proposal using a state machine and will let you evaluate if it's a good approach.

israellot:
Not submitting a PR, very experimental, but check my branch https://github.com/israellot/MQTTnet/tree/optmizations

The center idea is instead of reading from socket stream byte by byte, have a state machine where I can pull as many bytes as I could read and it will dispatch callbacks when a complete packet has been read. Partial data is kept on state, so we can always read the stream fully. 
By doing so we can avoid a lot of context switching, specially for small packets that will arrive in a single tcp packet. It also enables us to have a push pattern for received messages, which also avoids overhead of having a lot of tasks waiting for input. Tasks are dispatched when input is available. 

If you find the idea interesting, we can move forward. Need proper unit testing, etc.

MQTTnet.sln
israellot: not required, it's there by mistake

israellot:
Meanwhile, made a little amend to my serializer changes. Array.Resize would allocate another buffer to copy the previous one. 
It probably doesn't impact much, but feels better.

chkr1011:
From code perspective your solution looks good but from performance view I have doubts. I already tested a prototype which loads always 4KB from the underlying stream and returns the data from a cache. So if the cache has data no stream is accessed. But this leads to -5.000 messages per second in my usual tests. The overhead in managing this cache was stil too much. So I habe doubts if managing "steps" etc. will decrease or increase performance.

israellot:
On my tests i got a  bit of increase in performance. But I see it as a first step towards reducing GC. I am concerned about the number of ephemeral buffers and objects created in the heap that are garbage collected. I think we should also create more realistic tests, with more parallel message producers and listeners clients, which is the more real use case of mqtt. 

About the cache, I’ve looked at the implementation for sockets on dotnet core and they did a good job on the internals so a cache in front of it wouldn’t really help much indeed. On core 2.1 things will get even better, to a point that it will be replacing libuv on kestrel, so we should be able to steal some ideas from there on how they architectured sockets. 

Anyway it has been fun to work and experiment on this library. Thanks for the constant feedback

PR #234, opened 2018-04-24
Allow other websocket subprotocols

israellot:
Some websockets mqtt clients send subprotocols like _mqttv3.1_, http://www.hivemq.com/demos/websocket-client/ is an example. This PR handle accepting all subprotocols as long as they init with _mqtt_

### DIFF
--- a/Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
+++ b/Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
@@ -17,12 +17,11 @@ namespace MQTTnet.AspNetCore
                 {
                     string subprotocol = null;
 
-                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues)
-                     && requestedSubProtocolValues.Count > 0
-                     && requestedSubProtocolValues.Any(v => v.ToLower() == "mqtt")
-                     )
+                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues))
                     {
-                        subprotocol = "mqtt";
+                        subprotocol = requestedSubProtocolValues
+                                .OrderByDescending(p => p.Length)
+                                .FirstOrDefault(p => p.ToLower().StartsWith("mqtt"));
                     }
 
                     var adapter = app.ApplicationServices.GetRequiredService<MqttWebSocketServerAdapter>();


israellot:
Made the aspnet extension to look more like mvc style with a delegate. 
Added the option to not listen to tcp when using aspnet integration. There are use cases for websocket only server.

Frameworks/MQTTnet.AspnetCore/MQTTnet.AspnetCore.csproj
```diff
@@ -2,13 +2,14 @@
```
chkr1011: I assume this is not required and can be reverted?

Frameworks/MQTTnet.NetStandard/Client/MqttClient.cs
```diff
@@ -194,6 +194,47 @@ public MqttClient(IMqttClientAdapterFactory channelFactory, IMqttNetLogger logge
             }
         }
 
+        public async Task PublishAsync(MqttApplicationMessage applicationMessage)
```
chkr1011: Sorry but I don't agree here. A whole copy of this method including all the new interface methods only for avoiding an array with one element? Or is there a different reason?

Frameworks/MQTTnet.NetStandard/Implementations/MqttTcpChannel.cs
```diff
@@ -73,11 +74,12 @@ public Task DisconnectAsync()
         public Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
             return _stream.ReadAsync(buffer, offset, count, cancellationToken);
+
```
chkr1011: ?

Frameworks/MQTTnet.NetStandard/Implementations/MqttTcpChannel.cs
```diff
@@ -73,11 +74,12 @@ public Task DisconnectAsync()
...
         }
 
         public Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            return _stream.WriteAsync(buffer, offset, count, cancellationToken);
+             return _stream.WriteAsync(buffer, offset, count, cancellationToken);
         }
 
         public void Dispose()
```
chkr1011: ?

Frameworks/MQTTnet.NetStandard/Implementations/MqttWebSocketChannel.cs
```diff
@@ -81,7 +81,7 @@ public MqttWebSocketChannel(WebSocket webSocket)
             {
                 return;
             }
-
+
```
chkr1011: ?

Frameworks/MQTTnet.AspnetCore/MQTTnet.AspnetCore.csproj
israellot: True

Frameworks/MQTTnet.NetStandard/Client/MqttClient.cs
israellot: I guess the impact is minimal. The main idea was that most calls are single messages, and yet we were enumerating a collection at every call just to avoid having a specialized method and a collection one. The params only pattern is not very common on most libraries and even on dotnet itself, apart from places like string.format

Frameworks/MQTTnet.NetStandard/Implementations/MqttTcpChannel.cs
israellot: Trash indeed, guess I was sleepy at that time.

Frameworks/MQTTnet.NetStandard/Implementations/MqttTcpChannel.cs
israellot: good question

israellot:
Ok @chkr1011 , I will try to be more point-precise on my pull requests in the future.

PR #236, opened 2018-04-24
New topic matcher

israellot:
This implements a new topic matcher that does not allocate in the heap. String.split allocates a new string for each split item.
Took the implementation from mosquitto as reference
https://github.com/eclipse/mosquitto/blob/4f838e51611b9a71f2e3e98e6fddd02a5c1b4c03/lib/util_mosq.c

### DIFF
--- a/Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
+++ b/Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
@@ -17,12 +17,11 @@ namespace MQTTnet.AspNetCore
                 {
                     string subprotocol = null;
 
-                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues)
-                     && requestedSubProtocolValues.Count > 0
-                     && requestedSubProtocolValues.Any(v => v.ToLower() == "mqtt")
-                     )
+                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues))
                     {
-                        subprotocol = "mqtt";
+                        subprotocol = requestedSubProtocolValues
+                                .OrderByDescending(p => p.Length)
+                                .FirstOrDefault(p => p.ToLower().StartsWith("mqtt"));
                     }
 
                     var adapter = app.ApplicationServices.GetRequiredService<MqttWebSocketServerAdapter>();
--- a/Frameworks/MQTTnet.AspnetCore/AspNetMqttServerOptions.cs
+++ b/Frameworks/MQTTnet.AspnetCore/AspNetMqttServerOptions.cs
@@ -1,0 +1,12 @@
+﻿using MQTTnet.Server;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace MQTTnet.AspNetCore
+{
+    public class AspNetMqttServerOptions : MqttServerOptions, IAspNetMqttServerOptions
+    {
+        public Boolean ListenTcp { get; set; } = true;
+    }
+}
--- a/Frameworks/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
+++ b/Frameworks/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
@@ -12,7 +12,19 @@ namespace MQTTnet.AspNetCore
 
     public static class ServiceCollectionExtensions
     {
-        public static IServiceCollection AddHostedMqttServer(this IServiceCollection services, Action<AspNetMqttServerOptionsBuilder> configureOptions=null)
+
+        public static IServiceCollection AddHostedMqttServer(this IServiceCollection services)
+        {
+            return AddHostedMqttServerInternal(services, null);
+        }
+
+
+        public static IServiceCollection AddHostedMqttServer(this IServiceCollection services, Action<AspNetMqttServerOptionsBuilder> configureOptions)
+        {
+            return AddHostedMqttServerInternal(services, configureOptions);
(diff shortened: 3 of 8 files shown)


Frameworks/MQTTnet.AspnetCore/AspNetMqttServerOptions.cs
```diff
@@ -0,0 +1,12 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace MQTTnet.AspNetCore
+{
+    public class AspNetMqttServerOptions : MqttServerOptions, IAspNetMqttServerOptions
+    {
+        public Boolean ListenTcp { get; set; } = true;
+    }
```
chkr1011: There is already a option for this in the MqttServerOptions.

Frameworks/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
```diff
@@ -8,22 +8,45 @@
...
-            
+
             services.AddSingleton<MqttWebSocketServerAdapter>();
-            services.AddSingleton<MqttTcpServerAdapter>();
-            services.AddSingleton<IMqttServerAdapter>(s => s.GetService<MqttWebSocketServerAdapter>()); 
-            services.AddSingleton<IMqttServerAdapter>(s => s.GetService<MqttTcpServerAdapter>());
+            services.AddSingleton<IMqttServerAdapter>(s => s.GetService<MqttWebSocketServerAdapter>());
+
+            if (options.ListenTcp)
+            {
```
chkr1011: Please use the existing setting for this. But I assume this is not required at all because the adapter is only added but not started in the options has "IsEnabled" set to false for the TCP endpoint.

Frameworks/MQTTnet.NetStandard/Server/MqttTopicFilterComparer.cs
```diff
@@ -6,39 +6,115 @@ public static class MqttTopicFilterComparer
...
+                    tpos++;
+                    if (spos == slen && tpos == tlen)
+                    {
+                        return true;
+                    }
+                    else if (tpos == tlen && spos == slen - 1 && sub[spos] == '+')
+                    {
+                        if (spos > 0 && sub[spos - 1] != '/')
+                        {
+                            throw new ArgumentException("Invalid filter string", "filter");
```
chkr1011: I don't think that throwing an exception is not a good idea here. If the filter is invalid the server will produce lots of exceptions. In this case I would return "no match" and write a warning to the log. Or the better approach is valdiating the topic while being subscribed and return a failure to the client. Then this code needs no exceptions.

Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
```diff
@@ -17,12 +17,11 @@ public static IApplicationBuilder UseMqttEndpoint(this IApplicationBuilder app,
...
 
-                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues)
-                     && requestedSubProtocolValues.Count > 0
-                     && requestedSubProtocolValues.Any(v => v.ToLower() == "mqtt")
-                     )
+                    if (context.Request.Headers.TryGetValue("Sec-WebSocket-Protocol", out var requestedSubProtocolValues))
                     {
-                        subprotocol = "mqtt";
+                        subprotocol = requestedSubProtocolValues
+                                .OrderByDescending(p => p.Length)
```
chkr1011: Why ordering?

Frameworks/MQTTnet.AspnetCore/AspNetMqttServerOptions.cs
israellot: It's actually an option on MqttServerDefaultEndpointOptions and MqttServerTlsEndpointOptions
I was trying to be more explicit and user friendly on the helper methods. Disabling the tcp adapter is clearly an option that only applies when others are in use.

Frameworks/MQTTnet.AspnetCore/ApplicationBuilderExtensions.cs
israellot: The idea was trying to match the more specialized sub protocol eg mqttv-3.1 vs  mqttv-3.11

Frameworks/MQTTnet.NetStandard/Server/MqttTopicFilterComparer.cs
```diff
@@ -6,39 +6,115 @@ public static class MqttTopicFilterComparer
...
-            var fragmentsFilter = filter.Split(TopicLevelSeparator, StringSplitOptions.None);
+            var spos = 0;
+            var slen = sub.Length;
+            var tpos = 0;
+            var tlen = topic.Length;
+                        
+            var multilevel_wildcard = false;
 
-            for (var i = 0; i < fragmentsFilter.Length; i++)
+            while (spos < slen && tpos <= tlen)
```
sebastienwarin: `tpos' must be less the tlen, and NOT less or equal !
fixes in the PR #270

israellot: You are right @sebastienwarin 
This was referenced from a C code, o which a string[len] is a null char. Thanks for the check and benchmark on it.

chkr1011:
I close this pull request because it is no longer needed. The topic matcher is already part of the target branch and the other change (removing adapter) is done in a different way to avoid confusion (endpoint **enabled** in _MqttServerOptions_ BUT **disabled** in _AspNetMqttServerOptions_).

PR #256, opened 2018-05-10
added benchmark for serializers

JanEggers:
this should be the baseline for optimizing the serializer. 

goal should be to optimize perf and reduce allocations per message


JanEggers:
this are the baseline results


|                    Method |     Mean |     Error |    StdDev |     Gen 0 | Allocated |
| --------------------------- |---------:|----------:|----------:|----------:|----------:|
|   Serialize_10000_Messages | 3.401 ms | 0.0663 ms | 0.0737 ms |  835.9375 |   3.36 MB |
| Deserialize_10000_Messages | 8.870 ms | 0.2068 ms | 0.2124 ms | 2828.1250 |  11.33 MB |

Issue #230, opened 2018-04-21
Possible Memory Leak


VladimirAkopyan:
I am running MQTT server in a docker container, and i get strange behaviour where the amount of memory bloats into gigabytes. 

![image](https://user-images.githubusercontent.com/12278048/39085106-107f3bda-4576-11e8-9c0f-2461ebc116da.png)
![image](https://user-images.githubusercontent.com/12278048/39085113-17d1e25c-4576-11e8-8720-4296dcf11e97.png)

The system is
Producer ->  Server -> MultipleConsumers
The producer also uses MQTTnet, and sends ~100 messages per second, QOS 0
The consumers use MQTTnet too. they often disconnect without closing connection properly. 
The behaviour only occurs if the clients connect. If no client connect, then memory use stays very low (<100MB) 
It is not obvious to me why this is happening at this point. I could not reproduce it previously on my local machine in VS debugger, but I will try again soon. 

Kubernetes Version - 1.7
.Net SDK / Runtime : 
- 2.0 
- 2.1 preview 2

There is a related issue  - https://github.com/dotnet/coreclr/issues/14991#issuecomment-348428003
However I made sure that this isn't the issue appearing here by running `GC.Collect()` every second - the issue still persists. 

**Server Code**
```C#
public static readonly DateTime StartTimeUTC = DateTime.UtcNow;

static void Main(string[] args)
{
   Console.WriteLine("Server Started!");
   Task.Run(async () => await StartServer()).Wait();
}

private static async Task StartServer()
{
	var optionsBuilder = new MqttServerOptionsBuilder()
		.WithConnectionBacklog(100)
		.WithDefaultEndpointPort(1883);
	var mqttServer = new MqttFactory().CreateMqttServer();
	await mqttServer.StartAsync(optionsBuilder.Build());
	var ct = new CancellationTokenSource();
	var heartbeatTask = Task.Run(async () => await ServerHeartbeat(ct.Token, mqttServer)); 
  
	Console.WriteLine("Type 'quit' to exit");

	string input = string.Empty; 
	while (true)
	{
		//Docker has a habit of sending random shit into CMD
		char c = (char)Console.Read();
		input +=  c; 

	   if (input.Contains("quit")) 
		{
			Console.WriteLine("Shutting Down...");
			break;
		}
		else if (input.Length > 10)
		{
			input = input.Substring(input.Length - 3); 
		}
	}
	ct.Cancel();
	await heartbeatTask;
[...]

PR #266, opened 2018-05-11
Limiting the growth of PendingMessagesQueue to a configurable Value

VladimirAkopyan:
I believe this will address issue ref #230 
Unless I misunderstand logic of the project, It appears that, for a client that connects, subscribes, and disconnects, PendingMessagueue will grow without bound. 
If the client never reconnects, and the messages keep coming, PendingMessagueue will consume all memory on server. 

Here I introduce a max value for number for messages the queue will store, and if the value is exceeded, oldest packets will be dropped. 

I believe more work is required to address life-cycle of packets, connections and clients: 
* What is the lowest “message consumption rate” that should be allowed? Less then 1 per second is probably not a viable connection, and should be dropped. Or minimal data-rate should be defined. Otherwise we will have Packets piling up forever. 
* If KeepAlive crashes, it does not lead to shutting down of connection. I believe it should lead to the entire session being deleted - in my mind it's better than having some kind of weird dangling connection.


VladimirAkopyan:
damn, didn't use the dev branch.

chkr1011:
Good work! I also consider adding a strategy to either drop new packets or remove old ones (via enum).

PR #268, opened 2018-05-14
Removal of StaleSessions

VladimirAkopyan:
I want to make I can be confident in the broker's memory footprint, and be sure that in the broker nothing hangs around forever. 

I have implemented lifetime for StaleSessions. The ClientSessionManager now has a task that runs every 2 seconds, and check how much time has passed since receiving last packet/keepalive. 
If time passed is greater than `StaleSessionLifetime` it will clear the session, and even kill the connection if it's still alive. 

Is this the behaviour we want to have? 
To my use-case  cleaning up old sessions is more important than putting a limit on the amount of messages pending in a single client session. 

I also noticed that you use "normal" Dictionary and locking, not concurrent dictionary. Are you doing this to improve performance? 
I have tried re-writing the Session manager with concurrent dictionary and the code becomes a lot simpler.

### DIFF
--- a/Frameworks/MQTTnet.NetStandard/Server/IMqttServerOptions.cs
+++ b/Frameworks/MQTTnet.NetStandard/Server/IMqttServerOptions.cs
@@ -5,8 +5,13 @@ namespace MQTTnet.Server
     public interface IMqttServerOptions
     {
         int ConnectionBacklog { get; }
+
         int MaxPendingMessagesPerClient { get; }
         MqttPendingMessagesOverflowStrategy PendingMessagesOverflowStrategy { get; }
+        /// <summary>
+        /// How long server keeps sessions in memory after last packet/KeepAlive receeived. Be generous, if this setting is shorter than KeepAlive, it will close even Live connections!
+        /// </summary>
+        TimeSpan StaleSessionLifetime { get; }
 
         TimeSpan DefaultCommunicationTimeout { get; }
 
--- a/Frameworks/MQTTnet.NetStandard/Server/MqttClientPendingMessagesQueue.cs
+++ b/Frameworks/MQTTnet.NetStandard/Server/MqttClientPendingMessagesQueue.cs
@@ -58,6 +58,7 @@ namespace MQTTnet.Server
 
             if (_queue.Count >= _options.MaxPendingMessagesPerClient)
             {
+                _logger.Info($"ClientId: {_clientSession.ClientId} exceeded MaxPendingQueLength, dropping packets while Connected is {_clientSession.IsConnected}");
                 if (_options.PendingMessagesOverflowStrategy == MqttPendingMessagesOverflowStrategy.DropNewMessage)
                 {
                     return;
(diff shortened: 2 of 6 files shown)


chkr1011:
Thank you for your contribution but I have a different strategy for this topic. But don't worry this contribution is not obsolete 😄 
Background:
In my personal projects I have different requirements when a session can be deleted (truncating pending messages) or not and I want to avoid implementing them all in this projects fighting with different option combinations. So my plan is:

Rewrite (and rename) the _GetConnectedClients_-API and contain as much information about the sessions and clients as possible. This will also include a _Disconnect_ method and a _Delete_ method. The _Disconnect_ will only disconnect the adapter (if connected) and _Delete_ will disconnect the adapter and delete the session including all pending messages etc. This all is done in one transaction locking the sessions. But also additional methods like _ClearPendingMessages_ are possible.

Then your implementation will be moved to a new extension library called _MQTTnet.Extensions.ServerSessionManagement_. Then your implementation will be _attached_ to a MQTTnet server instance and pull the connected clients every 2 seconds and doing the required stuff. And my custom implementation will be not published because it is too special. 

But there will be one additional setting called _SessionHandling_ with a value of either _DeleteSessionAfterDisconnect_ or _KeepSessionAfterDisconnect_.  

What do you think about this? In my opinion this is more flexible and this library will only contain the basics.

Best regards
Christian

chkr1011:
And I used no ConcurrentDictionary because it is not required. While the session is created I have to lock the dictionary all the time. Otherwise it is filled several times leading to race conditions. With a regular lock I locked the dictionary while the critical path is being executed (check, disconnect, create new, will message etc.). With a ConcurrentDictionary I will only lock the access to the sessions collection which is not enough.

VladimirAkopyan:
Thanks for responce. Don't worry about merging my code, the implementation is really basic. I am just cautious that a naive user would expect a manages behaviour of some kind.

I am asking about concurrent dictionary because I have seen same issue as this guy
https://github.com/chkr1011/MQTTnet/issues/271
And I only had a limited look, but saw it as issue with locking. After implementing concurrent dictionary, the issue is gone. 
Could you describe what's the race condition? It seems to me that only on creation of session there would be a race. Am I missing something?

chkr1011:
Probably disabling persistent sessions via the new setting as a default is a solution. If actively enabled the user must deal with several implications.

Lets assume there are 5 clients with the same ID and they connect. There are several checks like if the session is known and if this session must be reused (due to client setting) etc. While this checks are made the code must prevent that other clients will start the same checks. Because maybe 5 threads at the same time will enter that method and create a new session object becuase all get _false- then trying to resolve the key from the dictionary. Then 5 sessions are made at the same time and the _Add_ at the end will fail. So I must lock the whole process to ensure that only one client (thread) will check the dictionary, disconect the old adapter, creates the new session and finally updates the dictionary. The following threads may do the same then afterwards. With a ConcurrentDictionary only the access to the dictionary is locked.

chkr1011:
@VladimirAkopyan Please have a look at the develop branch. I build the new "Session Status" API. On top of this your strategy can be applied. 

Let me know what you think.

VladimirAkopyan:
@chkr1011 thanks for addressing all the issues with sessions, this is clearly a better solution. 
As for locking, I could do a PR with concurrent dictionary and address the race condition by chaining tasks or putting all changes to session collection into a queue, that way you wouldn't have to obtain lock on every publish. 
But i am cautious this might be my pet issue rather than something that really matters.

Issue #269, opened 2018-05-15
SemaphoreFullException when subscribe to topic with multiple retained messages.


sebastienwarin:
Hi there,

The AsyncAutoResetEvent throws a SemaphoreFullException when we subscribe to a topic returning several message retained.

The Set() method is called to release the semaphore while no one is waiting for it.

`MQTTnet - TestApp.netstandard2.0
1 = Start client
2 = Start server
3 = Start performance test
4 = Start managed client
5 = Start public broker test
>> [2018-05-15T13:29:56.2202593Z] [3] [MqttServer] [Info]: Started.
Press any key to exit.
>> [2018-05-15T13:30:01.8067677Z] [4] [MqttTcpServerAdapter.MqttChannelAdapter] [Verbose]: RX <<< Connect: [ClientId=8a8a525ae576470db1c06355893c3251] [Username=sebastien] [Password=sebastien] [KeepAlivePeriod=60] [CleanSession=True]
>> [2018-05-15T13:30:01.8528974Z] [4] [MqttServer.MqttClientSessionsManager] [Verbose]: Created a new session for client '8a8a525ae576470db1c06355893c3251'.
>> [2018-05-15T13:30:01.8730484Z] [4] [MqttTcpServerAdapter.MqttChannelAdapter] [Verbose]: TX >>> ConnAck: [ConnectReturnCode=ConnectionAccepted] [IsSessionPresent=False] [Timeout=00:00:15]
>> [2018-05-15T13:30:01.8820406Z] [4] [MqttServer] [Info]: Client '8a8a525ae576470db1c06355893c3251': Connected.
>> [2018-05-15T13:30:02.8828588Z] [6] [MqttTcpServerAdapter.MqttChannelAdapter] [Verbose]: RX <<< Subscribe: [PacketIdentifier=1] [TopicFilters=PC-SEB/HWMonitor/intel/0/load/+@AtMostOnce]
>> [2018-05-15T13:30:02.9023461Z] [6] [MqttTcpServerAdapter.MqttChannelAdapter] [Verbose]: TX >>> SubAck: [PacketIdentifier=1] [SubscribeReturnCodes=SuccessMaximumQoS0] [Timeout=00:00:15]
[...]

PR #270, opened 2018-05-15
Replace the Async coordination primitives & replace the MqttTopicFilterComparer to improve the performance

sebastienwarin:
Fixes the #269 issue


chkr1011:
Hi @sebastienwarin,
thank you for your contribution. In general I agree with your changes (thanks for creating a benchmark) but I will not merge this pull request because:

1. The code style is not matching the existing code. Please check namings of local fields, members and remove _this_ etc.
2. This pull request contains several not related changes. So please create one for the topic filter and another one for the async _AutoResetEvent_ etc.
3. Maybe I am wrong but I assume that the _releaser_ field within the _using_ is not used and thus can be removed. 

Best regards
Christian

sebastienwarin:
hi @chkr1011,

I have update the code to match with your coding style. You right, the releaser field is not used, so I remove it.

I have also merge your last commits and fix errors in the RetainedMessagesManager and ClientSessionsManager.

All these changes are all in this unique PR and I have no time to split it into differents branches.

Frameworks/MQTTnet.NetStandard/Server/MqttClientSessionsManager.cs
```diff
@@ -262,6 +258,7 @@ private MqttConnectReturnCode ValidateConnection(MqttConnectPacket connectPacket
 
                 foreach (var clientSession in _sessions.Values)
                 {
+                    _logger.Verbose($"DispatchApplicationMessageAsync for {clientSession.ClientId} ");
```
chkr1011: Please do not use string interpolation or string.format here. The logger will generate the final string if there is a listener for such messages. With string interpolation here there is a overhead I want to avoid.

chkr1011:
Good work 👍 I only have one small topic left. After this is fixed I will merge it.
Best regards
Christian

Frameworks/MQTTnet.NetStandard/Server/MqttClientSessionsManager.cs
sebastienwarin: It's a mistake, I added these logs to debug and forgot to delete them;)

chkr1011:
Good work! 👍

PR #286, opened 2018-05-29
ManagedClientStorage now works

JTrotta:
Some refactoring on 4th dec 2017, avoided the ManagedClientStorage to work correctly.


chkr1011:
Good work 👍 
I will refactor the storage manager a little bit. I assume there is no need to access the application messages from the storage manager directly right?

JTrotta:
Yes. It was the fastest way to let it work.

PR #291, opened 2018-05-31
use new project system for benchmarks

JanEggers:


chkr1011:
Good work 👍 I am wondering how to create new projects with the new system. Is this a new project type is the wizard and I was not able to find it?

Issue #250, opened 2018-05-02
Refactoring Transport once more (aspnetcore 2.1)


JanEggers:
Hi @chkr1011  im planning to adapt the Connection abstraction of aspnetcore 2.1 for mqttnet. these abstractions provide a unified way to handle tcp, tls and websockets.

common:
https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Connections.Abstractions/ConnectionContext.cs

tcp:
https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs

websocket:
https://github.com/aspnet/SignalR/blob/dev/src/Microsoft.AspNetCore.Http.Connections/Internal/Transports/WebSocketsTransport.cs

I see two ways of integration. What would you prefere?

Option 1: Implement IMqttServerAdapter, IMqttChannelAdapter for tcp in MQTTnet.AspnetCore (that depends on aspnetcore already just need to increment version to 2.1 once released)

Option 2: remove IMqttChannel, IMqttChannelAdapter, ... and let MQTT.Net depend on Kestrel 

what I expect from the change:

1. perf boost
2. less errors

markusschaber:
A hard dependency on Kestrel might be too much for some use cases (not all of them use Kestrel).

But as long as we can still plug in our own adapters, I won't mind too much... :-)

israellot:
Depending on kestrel for a client-only use case would seem odd. I know @chkr1011 doesn't want to split the library on client / server though.

JanEggers:
lets see what the aspnet comes up with. 

client side story should be supported in 2.2.

also I could instead just depend on https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Connections.Abstractions and copy the code for the implementations so there is no dependency on kestrel. but I dont really like that approach as it negates the advantage of having the code maintened by the aspnet team

israellot:
We could fork and submodule kestrel code

JanEggers:
@chkr1011 can you give me some feedback pls 2.1 is final and I would like to start experimenting over the weekend

chkr1011:
[...]

PR #292, opened 2018-05-31
Improved Serializer memory usage

JanEggers:
before:

|                      Method |     Mean |     Error |    StdDev |     Gen 0 | Allocated |
| --------------------------- |---------:|----------:|----------:|----------:|----------:|
|    Serialize_10000_Messages | 2.496 ms | 0.0463 ms | 0.0410 ms |  781.2500 |   3.13 MB |
|  Deserialize_10000_Messages | 7.791 ms | 0.0842 ms | 0.0746 ms | 3765.6250 |  15.11 MB |

after:

|                      Method |      Mean |     Error |    StdDev |     Gen 0 |  Allocated |
| --------------------------- |---------:|----------:|----------:|----------:|-----------:|
|    Serialize_10000_Messages |     2.830 ms | 0.0558 ms | 0.0597 ms |  152.3438 |  625.01 KB |
|  Deserialize_10000_Messages |      5.313 ms | 0.0790 ms | 0.0739 ms | 1218.7500 | 5000.15 KB |

we can argue about serializer performance as it is 10% slower now but I guess the memory pooling should be worth it especially for bigger messages. anyway let me know what u think about this.

there is still some potential in string encoding MqttPacketWriter.cs line 33 - 35 but that will have to wait until the package is final.

preview with that  change:

|                      Method |     Mean |     Error |    StdDev |     Gen 0 |  Allocated |
| --------------------------- |---------:|----------:|----------:|----------:|-----------:|
|    Serialize_10000_Messages | 2.613 ms | 0.0507 ms | 0.0623 ms |   74.2188 |   312.5 KB |
|  Deserialize_10000_Messages | 5.488 ms | 0.1090 ms | 0.1455 ms | 1218.7500 | 5000.15 KB |

the header deserialization can also be improved if we use new connection abstractions (see issue #250 )

### DIFF
--- a/Source/MQTTnet/Adapter/MqttChannelAdapter.cs
+++ b/Source/MQTTnet/Adapter/MqttChannelAdapter.cs
@@ -94,7 +94,7 @@ namespace MQTTnet.Adapter
 
                 var packetData = PacketSerializer.Serialize(packet);
 
-                await _channel.WriteAsync(packetData.Array, packetData.Offset, packetData.Count, cancellationToken).ConfigureAwait(false);
+                await _channel.WriteAsync(packetData, 0, packetData.Length, cancellationToken).ConfigureAwait(false);
             }
             catch (Exception exception)
             {
@@ -169,17 +169,14 @@ namespace MQTTnet.Adapter
                     return new ReceivedMqttPacket(fixedHeader.Flags, null);
                 }
 
-                var body = new byte[fixedHeader.RemainingLength];
                 var bodyOffset = 0;
+                var body = new byte[fixedHeader.RemainingLength];
                 var chunkSize = Math.Min(ReadBufferSize, fixedHeader.RemainingLength);
 
-                do
+                while (bodyOffset < fixedHeader.RemainingLength)
                 {
                     var bytesLeft = body.Length - bodyOffset;
-                    if (chunkSize > bytesLeft)
-                    {
-                        chunkSize = bytesLeft;
-                    }
+                    chunkSize = Math.Max(chunkSize, bytesLeft);
 
                     // async/await is not used to avoid the overhead of context switches. We assume that the reamining data
                     // has been sent from the sender directly after the initial bytes.
@@ -190,9 +187,9 @@ namespace MQTTnet.Adapter
                     }
 
                     bodyOffset += readBytes;
-                } while (bodyOffset < body.Length);
-
-                return new ReceivedMqttPacket(fixedHeader.Flags, new MqttPacketBodyReader(body));
+                }
+            
+                return new ReceivedMqttPacket(fixedHeader.Flags, body);
             }
             finally
             {
--- a/Source/MQTTnet/Adapter/ReceivedMqttPacket.cs
+++ b/Source/MQTTnet/Adapter/ReceivedMqttPacket.cs
@@ -1,10 +1,11 @@
 ﻿using MQTTnet.Serializer;
+using System;
 
 namespace MQTTnet.Adapter
 {
     public class ReceivedMqttPacket
     {
-        public ReceivedMqttPacket(byte fixedHeader, MqttPacketBodyReader body)
+        public ReceivedMqttPacket(byte fixedHeader, Memory<byte> body)
         {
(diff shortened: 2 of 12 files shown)


JanEggers:
updated performance improvements after rebase:

|                      Method | Iterations |     Mean |     Error |    StdDev |     Gen 0 |  Allocated |
| --------------------------- |----------- |---------:|----------:|----------:|----------:|-----------:|
|    Serialize_10000_Messages |      10000 | 2.768 ms | 0.0609 ms | 0.0770 ms |  152.3438 |  625.01 KB |
|  Deserialize_10000_Messages |      10000 | 4.560 ms | 0.0901 ms | 0.1578 ms | 1218.7500 | 5000.15 KB |

JanEggers:
@chkr1011 can you please drop a comment if you cnsider accepting this. and if not close it i have done 3 rebases already and will do a final one if you intend to merge

chkr1011:
I already pushed a different strategy. But I will compare this results with the results of my change. If this one is faster I will consider using this PR. Otherwise I will keep my optimization.

JanEggers:
I rebased again and this is the result:

without my changes:

|                     Method |     Mean |     Error |    StdDev |     Gen 0 | Allocated |
|--------------------------- |---------:|----------:|----------:|----------:|----------:|
|   Serialize_10000_Messages | 2.835 ms | 0.0386 ms | 0.0323 ms |  781.2500 |   3.13 MB |
| Deserialize_10000_Messages | 8.068 ms | 0.0499 ms | 0.0442 ms | 3671.8750 |  14.73 MB |

with my changes:

|                     Method | Iterations |     Mean |     Error |    StdDev |     Gen 0 |  Allocated |
| --------------------------- |----------- |---------:|----------:|----------:|----------:|-----------:|
|   Serialize_10000_Messages |      10000 | 2.920 ms | 0.0562 ms | 0.0730 ms |  152.3438 |  625.01 KB |
| Deserialize_10000_Messages |      10000 | 5.344 ms | 0.0431 ms | 0.0382 ms | 1195.3125 | 4922.02 KB |

so serializing is a little slower but deserialing is nearly twice as fast, but the main advantage is that memory usage is much lower, this will come more into play with bigger messanges and larger topic strings

chkr1011:
Good work! That's a lot better than mine. But I still have to check what needs to be done regarding the copy of the code from the other project. I will read the license and see if we can just copy it or if we have to include some more details and copyright texts etc.

Maybe I am also able to improve my solution 😄

JanEggers:
that might help:

https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)

chkr1011:
Interesting reference. Thanks 👍 
I accepted the challenge to get a better performance without include code from other projects 😄. I am on a good way. I first optimized the benchmark because it was not fair. The the benchmark itself created a lot of memory consuming buffers. I fixed this and already migrated the "indea" behind the SingalR code and there are my results:

|                     Method |     Mean |     Error |    StdDev |    Gen 0 |  Allocated |
|--------------------------- |---------:|----------:|----------:|---------:|-----------:|
|   Serialize_10000_Messages | 1.147 ms | 0.0053 ms | 0.0049 ms | 101.5625 |  625.01 KB |
| Deserialize_10000_Messages | 2.252 ms | 0.0119 ms | 0.0105 ms | 964.8438 | 5937.55 KB |

I believe that the different execution time is caused by different machines. The interesting thing is that the memory for serializing is now exactly the same.

I am still working on the deserialize code.

chkr1011:
I updated the values.

The buffer is part of the ThreadStatic _MqttPacketWriter_. If the code is really on a different thread it will have a dedicated instance.

JanEggers:
that is exactly my point so if you have a large message build up a buffer. then async send it and land on a differnt thread then the clear will have no effect since the buffer is on the original thread

chkr1011:
OK now I got it. Mhhh. AsyncLocal is not available for .net452. I will try to find another solution.

JanEggers:
anyway I will close this pr as your results are now very close to my my results with system.memory and System.Buffers

PR #293, opened 2018-05-31
ApplicationMessageWithId

JTrotta:
Please have  a look, we are still testing this feature.
Right now is 10 times faster than having a single file, during SaveQueuedMessagesAsync. 
(test with 4000 messages serialized on disk one by one, instead having a single file with a json list of 4000 messages).


chkr1011:
Hi,
I got your point and agree that it is a good feature but I don't want to integrate this in the regular basic client implementation. This core lib should only implement the MQTT basics. So to implement this I only accept changes in the _Managed_ client and used classes etc. 

I am still evaluating some possible solutions. One of them is adding a dictionary with _user_ properties like in ASP.net. This dictionary is not bound to a particular feature and allows users to extend a message with custom data. This can be used by the storage to separate files. BUT these properties are also supported in MQTT v5. So it will give them a different meaning. 

I am also thinking about creating a _MqttManagedApplicationMessage_. This class uses NO inheritance and combines an _ID_ with a regular _ApplicationMessage_ as properties. Then the regular builder can be used. The _ManagedClient_ will get an overload for the publish method accepting this new type of message. For the basic type an empty _MqttManagedApplicationMessage_ is generated internally with a ID of _null_ (nullable guid). The storage interface will only accept the new _MqttManagedApplicationMessage_. In my opinion this will not affect any basic protocol and message format stuff and let you save the pending messages more quickly. We can also extend the new _MqttManagedApplicationMessage_ later with _managed_ features like _RetryCount_, _Priority_ etc. without touching the basic classes in any way. I am still evaluating this approach. What do you think about this?

JTrotta:
Hello Chris,
I absolutely agree about the first point: we must first implement the standard MQTT protocol as stated by OASIS into the various versions (3, 3.1.1, next 5, etc.), then add some useful/customization features. My intent is to integrate _MqttManagedApplicationMessage_ the it only into ManagedClient. 
I'm working on Mono in an embedded system and the second approach should be the best and the preferable one. But why not use inheritance? In  my opinion, even if we should first follow standards, derive from it using object programming inheritance, ensures the new features to enhance the basic class without "forget" the protocol's basics. Create a new class without inheritance expose to the risk of having, in the future, two completely different elements to manage separately , while the new _MqttManagedApplicationMessage_ it's only an enhance of the basic standard one.
On the other hand I cannot see how inheritance would create a problem to the standard basic class. What I'm missing?

chkr1011:
Hi,
I want to avoid inheritance to me more flexible and maintainable. I refer to the _Composition over Inheritance_ principle (https://en.wikipedia.org/wiki/Composition_over_inheritance). This article also describes the benefit from using this approach.

Best regards
Christian

JTrotta:
I see.
I agree with you, but not in any situations. In our case we are talking about few properties and methods and benefits and drawbacks are equivalent. 
Anyway it's just a matter of pattern. Let's create a new class. 
Just one questions: what you mean for RPC? Remote Procedure Call? You told me that, in a near future, you would have moved the managed client to the RPC library, but RPC, for me, is something different from the managed client. Or you would like to create a library containing all the extensions included RPC and ManagedClient?

chkr1011:
Sorry what I mean is creating a new extension lib _like_ for RPC called _MQTTnet.Extensions.ManagedClient_. This will contain all the managed client stuff. Advantage is a hard separation between core implementation and _features on top_, releases of managed client without updating the base lib and I got statistics from nuget to see how much the managed client is used.

JTrotta:
OK.
I would have preferred to have everything in one library but, I got you point of view.

PR #294, opened 2018-05-31
added extension method

JanEggers:


chkr1011:
Good iead 👍 Removes a lot of duplicate and annoying code.

PR #301, opened 2018-06-04
Respect keep alive interval after sending

cstichlberger:
Currently, the MqttClient potentially waits up to twice the configured keepalive timeout before sending a keepalive message to the broker. This can lead to the broker disconnecting the client.
Even with the broker tolerance (e.g. mosquitto uses 1.5 times the keepalive timeout configured by the client), minor delays on the connection (and possibly the low accuracy of Stopwatch) can lead to the client getting disconnected by the broker under quite normal circumstances.

This happens frequently for me, mainly when starting the client. The scenario plays out as follows: 

- the keepalive loop is started and enters the wait. Shortly thereafter some messages are sent. 
 - The first wait of the keepalive loop expires, but because the sent messages did reset the _sendTracker some milliseconds after the loop was started, the condition (_sendTracker.Elapsed > keepAliveSendInterval) determines that the keepalive packet is not yet due.
 - Then the loop again waits the full waittime. 

In total, nearly 2 times the configured interval was spent without any message to the broker. Adding timer inaccuracy and network delays, the total time during which the broker received no message exceeds the timeout, and the broker breaks the connection.  
Note that the broker's tolerance of timeout * 1.5 and the client's tolerance of timeout * 0.75 factors up to exactly 2 times the configured wait time, which is exceeded by just a bit in these cases.

The fix has the keepalive loop wait only the remaining time until a keepalive packet is due (and not the full keepalive time as it is now).

### DIFF
--- a/Frameworks/MQTTnet.NetStandard/Client/MqttClient.cs
+++ b/Frameworks/MQTTnet.NetStandard/Client/MqttClient.cs
@@ -419,12 +419,14 @@ namespace MQTTnet.Client
                         keepAliveSendInterval = _options.KeepAliveSendInterval.Value;
                     }
 
-                    if (_sendTracker.Elapsed > keepAliveSendInterval)
+                    var waitTime = keepAliveSendInterval - _sendTracker.Elapsed;
+                    if (waitTime <= TimeSpan.Zero)
                     {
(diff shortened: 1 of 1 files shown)


chkr1011:

PR #303, opened 2018-06-05
Fix dual result setting

### LINKED ISSUE #161: Unobserved exception
Broker: Mosquitto v1.4.14
MQTTnet: v2.6.0

Both the connection and server are stable but exceptions like these happen alot.

```There are no context policies.

 System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. ---> MQTTnet.Exceptions.MqttCommunicationTimedOutException: Exception of type 'MQTTnet.Exceptions.MqttCommunicationTimedOutException' was thrown.

   at MQTTnet.Internal.TaskExtensions.<TimeoutAfter>d__1`1.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at MQTTnet.Client.MqttPacketDispatcher.<WaitForPacketAsync>d__4.MoveNext()

   --- End of inner exception stack trace ---

   at System.Threading.Tasks.TaskExceptionHolder.Finalize()

---> (Inner Exception #0) MQTTnet.Exceptions.MqttCommunicationTimedOutException: Exception of type 'MQTTnet.Exceptions.MqttCommunicationTimedOutException' was thrown.

   at MQTTnet.Internal.TaskExtensions.<TimeoutAfter>d__1`1.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at MQTTnet.Client.MqttPacketDispatcher.<WaitForPacketAsync>d__4.MoveNext()<---
```

GieltjE:
This should fix #161

### DIFF
--- a/Frameworks/MQTTnet.NetStandard/Client/MqttPacketDispatcher.cs
+++ b/Frameworks/MQTTnet.NetStandard/Client/MqttPacketDispatcher.cs
@@ -32,8 +32,8 @@ namespace MQTTnet.Client
             
             if (_awaiters.TryRemove(key, out var awaiter))
             {
-                awaiter.SetResult(packet);
-                Task.Run(() => awaiter.SetResult(packet)); // Task.Run fixes a dead lock. Without this the client only receives one message.
+                Task task = Task.Factory.StartNew(() => awaiter.SetResult(packet)); // Task.Run fixes a dead lock. Without this the client only receives one message.
+                task.ContinueWith(t => { }, TaskContinuationOptions.OnlyOnFaulted);
                 return;
             }
 


JanEggers:
i dont think this is a solution it just swallows the error.

we could change SetResult to TrySetResult but I Think the real issue is that we dispatch exceptions like this:

https://github.com/GieltjE/MQTTnet/blob/a505e4e406465976f8d0db8ae996b9e9f9f76d92/Frameworks/MQTTnet.NetStandard/Client/MqttPacketDispatcher.cs#L14-L17

that way the task completion sources stay in the dictionary and if one message complets after we set the exception we get another error because we cannot set a result when the tcs was completed as error

JanEggers:
anyway are you able to create a unittest to repro the behavior that breaks the dispatcher?

JanEggers:
you should use TrySetResult and you can still use task.Run() to create a task.
also in case an exception was dispatched the awaiters should be cleared after that. and it seems we need some more unittests for the dispatcher

chkr1011:
Thank you for your contribution. Good work 👍 but I already fixed it in my local repo. So I will close this one.

@JanEggers I also applied your suggestions.

PR #318, opened 2018-06-20
improve session manager

JanEggers:
with this change i see 25% perf improvement in throughput

### DIFF
--- a/Source/MQTTnet/Server/MqttClientSessionsManager.cs
+++ b/Source/MQTTnet/Server/MqttClientSessionsManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using MQTTnet.Adapter;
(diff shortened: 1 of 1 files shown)


chkr1011:
Good work 👍 I also noticed that and was wondering. In the past I usually used custom locking and then moved to the ConcurrentDictionary. 

I also changed some more places with the same problem. And I refactored some async/await calls which area not required in my opinion (in terms of performance).

It would be great if you could do a quick review and letting me know what you think. Additionally a new performance test from your side would be nice. I want to see if the latest changes increase or decrease the performance in your use case.

JanEggers:
before
|                Method |     Mean |     Error |    StdDev |     Gen 0 |    Gen 1 | Allocated |
|  -------------------- |---------:|----------:|----------:|----------:|---------:|----------:|
|   Send_10000_Messages | 15.77 ms | 0.3055 ms | 0.4573 ms | 2046.8750 | 109.3750 |   8.67 MB |

after
|                Method |     Mean |     Error |    StdDev |     Gen 0 |    Gen 1 | Allocated |
|  -------------------- |---------:|----------:|----------:|----------:|---------:|----------:|
|   Send_10000_Messages | 14.78 ms | 0.2932 ms | 0.4478 ms | 1937.5000 | 187.5000 |   8.11 MB |


slightly better, good work!

Issue #325, opened 2018-06-22
build issues with 2.8.0 rc2


JanEggers:
new MqttFactory().CreateMqttClient(new MqttNetLogger(), new MqttClientConnectionContextFactory());

generates

error CS0012: The type 'IMqttClientAdapterFactory' is defined in an assembly that is not referenced. You must add a reference to assembly 'MQTTnet, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.

JanEggers:
nuget version seem correct but dlls are version 0.0.0.0 didnt figure ut why yet

JanEggers:
there is still an issue....

![image](https://user-images.githubusercontent.com/6747376/41835398-b8baa7e4-7857-11e8-8511-25c02272589a.png)

JanEggers:
// Assembly: MQTTnet.AspNetCore, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
// Assembly location: C:\Users\egja\.nuget\packages\mqttnet.aspnetcore\2.8.0-rc4\lib\netstandard2.0\MQTTnet.AspNetCore.dll

JanEggers:
@chkr1011 can you please provide a new rc so i can verify its fixed?

chkr1011:
@JanEggers Done. I will close this ticket hoping it is solved now 😄

PR #330, opened 2018-06-25
fixed another build issue

JanEggers:
sry did not run build scripts because i dont have the certs...

when removing the signing options I found this, maybe you should abort the script if a build error occours or make it show up in red so its easier to see....

contributes to #325


chkr1011:
Oh that was a stupid error 😄

PR #339, opened 2018-06-29
fixed parsing issue and added test

JanEggers:
sry for this fail but it seems that everythings thats not tested is broken :(.

anyway #325  is finally fixed 

I also added missing dependencies to the nuspec (by accident in your dev branch instead of mine)


chkr1011:
Good work 👍 But I assume this affects ASP.NET only. Should I release a new RC now (is the version issue fixed?) or should I wait?

JanEggers:
yes this only affects aspnet core and only with the new transport. the version issue is fixed so yes please create a new rc

PR #343, opened 2018-07-01
Server shutdown

### LINKED ISSUE #342: MqttManagedClient Disconnected Event Not Firing on Server.StopAsync
Hello,
This may be related to https://github.com/chkr1011/MQTTnet/issues/259. If so, feel free to merge them.
The difference is he does not mention that it is occurring on the IManagedMqttClient disconnected event. All my tests assume the Server and Client are on the same machine, so network traffic/firewall should not be a consideration.

I have 2 console apps based on the wiki. One is acting as the server and the other as client(s) where the client code is based on IManagedMqttClient.

I can start the server, and a client. If I then disconnect the client using Client.StopAsync(), the servers Server_ClientDisconnected responds immediately. If I reconnect the client, Server_ClientConnected responds immediately.

However, if I stop the server (via Server.StopAsync()) the client does not respond the same. A considerable amount of time passes before the client realizes the server is not connected and fires Client_Disconnected. I would assume that occasionally the client pings the server as in a keepalive and this is when the disconnect is detected. If I close the server application entirely, the clients then respond immediately.

This is problematic because since the client is unaware of the disconnection, messages are sent to a black whole (are not queued) as if a connection exists when it does not.

Note that in machine to machine tests, I found that the disconnected events did not fire on either side using the same StopAsync functions even though the messaging worked as expected. However, I am not convinced this was not either network or firewall issues. 

Thanks for a great library. I believe I can improve on this using IMqttClient by writing my own message queuing routine, but thought I'd mention this in case it was important enough to track down.

JanEggers:
fixes #342

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -25,8 +25,8 @@ namespace MQTTnet.Client
 
         private IMqttClientOptions _options;
         private CancellationTokenSource _cancellationTokenSource;
-        private Task _packetReceiverTask;
-        private Task _keepAliveMessageSenderTask;
+        internal Task _packetReceiverTask;
+        internal Task _keepAliveMessageSenderTask;
         private IMqttChannelAdapter _adapter;
         private bool _cleanDisconnectInitiated;
         private TaskCompletionSource<bool> _disconnectReason;
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -229,11 +229,8 @@ namespace MQTTnet.Client
 
             try
             {
-                if (clientWasConnected)
-                {
-                    await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
-                    await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);
-                }
+                await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
+                await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);
 
                 if (_adapter != null)
                 {
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -222,7 +222,7 @@ namespace MQTTnet.Client
 
             try
             {
-                if (!(exception is MqttCommunicationClosedGracefullyException))
+                if (clientWasConnected)
                 {
                     await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
                     await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -29,6 +29,7 @@ namespace MQTTnet.Client
         private Task _keepAliveMessageSenderTask;
         private IMqttChannelAdapter _adapter;
         private bool _cleanDisconnectInitiated;
+        private TaskCompletionSource<bool> _disconnectReason;
 
         public MqttClient(IMqttClientAdapterFactory channelFactory, IMqttNetLogger logger)
         {
@@ -54,6 +55,7 @@ namespace MQTTnet.Client
             try
             {
                 _cancellationTokenSource = new CancellationTokenSource();
+                _disconnectReason = new TaskCompletionSource<bool>();
                 _options = options;
                 _packetIdentifierProvider.Reset();
                 _packetDispatcher.Reset();
@@ -85,8 +87,10 @@ namespace MQTTnet.Client
             catch (Exception exception)
             {
                 _logger.Error(exception, "Error while connecting with server.");
-                await DisconnectInternalAsync(null, exception).ConfigureAwait(false);
-
+                if (_disconnectReason.TrySetException(exception))
+                {
+                    await DisconnectInternalAsync(null, exception).ConfigureAwait(false);
+                }
                 throw;
             }
         }
@@ -104,7 +108,10 @@ namespace MQTTnet.Client
             }
             finally
             {
-                await DisconnectInternalAsync(null, null).ConfigureAwait(false);
+                if (_disconnectReason.TrySetCanceled())
+                {
+                    await DisconnectInternalAsync(null, null).ConfigureAwait(false);
+                }
             }
         }
 
@@ -355,7 +362,10 @@ namespace MQTTnet.Client
                     _logger.Error(exception, "Unhandled exception while sending/receiving keep alive packets.");
                 }
 
-                await DisconnectInternalAsync(_keepAliveMessageSenderTask, exception).ConfigureAwait(false);
+                if (_disconnectReason.TrySetException(exception))
+                {
+                    await DisconnectInternalAsync(_keepAliveMessageSenderTask, exception).ConfigureAwait(false);
+                }
             }
             finally
             {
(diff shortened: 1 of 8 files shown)


Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -222,8 +222,11 @@ private void ThrowIfConnected(string message)
 
             try
             {
-                await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
-                await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);
+                if (!(exception is MqttCommunicationClosedGracefullyException))
```
JanEggers: I am not 100% shure if this is correct or if we dont want to wait if any MqttCommunicationException is thrown

chkr1011: Mhh I am also not 100 % sure. This wait was added by me because otherwise some exceptions will occur because the threads access the adapter. But the adapter is already disposed.

Source/MQTTnet/Server/MqttClientSession.cs
```diff
@@ -29,6 +29,8 @@ public class MqttClientSession : IMqttClientSession
         private MqttApplicationMessage _willMessage;
         private bool _wasCleanDisconnect;
         private IMqttChannelAdapter _adapter;
+        private Task<bool> _run;
```
chkr1011: In my opinion this should be renamed to _worker_ or _workerTask_ or whatever because _run_ is confusing (to me) 😄

Source/MQTTnet/Server/MqttClientSession.cs
```diff
@@ -122,17 +134,34 @@ public void FillStatus(MqttClientSessionStatus status)
                 {
                     _adapter.ReadingPacketStarted -= OnAdapterReadingPacketStarted;
                     _adapter.ReadingPacketCompleted -= OnAdapterReadingPacketCompleted;
+                    await Cleanup();
```
chkr1011: ConfigureAwait is missing here.

Source/MQTTnet/Server/MqttClientSession.cs
JanEggers: will do

Source/MQTTnet/Server/MqttClientSession.cs
JanEggers: will do

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -222,8 +222,11 @@ private void ThrowIfConnected(string message)
 
             try
             {
-                await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
-                await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);
+                if (clientWasConnected)
```
JanEggers: this condition also works instead of exception type checking

iamoldli: When server is closed,the SendKeepAliveMessagesAsync and ReceivePacketsAsync will all catch exceptions ,then then will all call the ```DisconnectInternalAsync``` method,in ```DisconnectInternalAsync``` method,under code whill deadlock
 await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
 await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);

I add under code,then the Disconnected event whill be call
image

my English is not good,please forgive me ~

jenscski: But if the disconnect i because of server crash/stop, the Disconnected event is called twice ...

iamoldli: I have find the reason, You can see #358

chkr1011:
Good work! I assume the discussion was completed right?

jenscski:
> I didnt have a look yet why the event is fired twice
The reason for double event is that of a lost connection with server network/server crash/server stop both the SendKeepAliveMessagesAsync and ReceivePacketsAsync will trigger the message when _cleanDisconnectInitiated is not set, which it isn't when connection with server is lost.

In this PR you are using CancellationToken.Register to clean up in the session code on the server, I have tested with a rewrite of the client using the same approach, and a ManualResetEvent to wait for proper disconnect. Should I post a PR for it?

JanEggers:
@chkr1011 @jenscski I pushed another change can you please verify if it is now correct?

jenscski:
@JanEggers  No double firing of events, but if the connection fails at AuthenticateAsync in ConnectAsync, then then _packetReceiverTask is not awaited. Not sure if it's important to wait for it, but probably it should be awaited.

JanEggers:
hmm we should add a tests that verifies behavior if connect fails mid air

JanEggers:
@jenscski it should also workas intended now. was a little hard to break it like you described but now its tested

jenscski:
@JanEggers  Looks like client behaves correctly now.

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -25,8 +25,8 @@ public class MqttClient : IMqttClient
 
         private IMqttClientOptions _options;
         private CancellationTokenSource _cancellationTokenSource;
-        private Task _packetReceiverTask;
-        private Task _keepAliveMessageSenderTask;
+        internal Task _packetReceiverTask;
+        internal Task _keepAliveMessageSenderTask;
```
jenscski: How about using reflection instead of making the private variables visible?

Source/MQTTnet/Server/MqttClientSession.cs
```diff
@@ -153,6 +182,8 @@ public void Stop(MqttClientDisconnectType type)
                 }
 
                 _willMessage = null;
+
+                _workerTask?.GetAwaiter().GetResult();
```
jenscski: This will cause a deadlock as Stop is called from RunInternalAsync, and Stop if waiting for RunInternalAsync to finish ...

Stop is called in ProcessReceivedPacket, when a MqttDisconnectPacket is received

chkr1011: Is there a way to scip the wait if the calling thread is the thread which must be awaited? Can we check this by the thread ID or task ID etc.?

chkr1011: Ah yes I remember such error. I was also fighting them and tried to avoid them with the wait calls. I will try to find another solution.

jenscski:
Is it possible to split this PR into one for client, and one for server?

So the client part could be merged, if it's ok.

Source/MQTTnet/Server/MqttClientSession.cs
JanEggers: > When a client is doing a disconnect, the _adapter is disconnected and disposed in Cleanup triggered by the cancellationToke.Register(...).

for me it sounds like a similar issue like in the client with the two disconnects overlapping so i guess we could also solve this with a tcs. for me it is more important to add tests for that case becasue I assume everything is fine if the tests are green. 

> I don't think the code is necessary at all
on my machine it was otherwise the server socket is closed before the disconnect message is send to all clients

jenscski: > ... add tests for that case becasue I assume everything is fine if the tests are green.

If you are interested, so here are my test, with a very temporary name, and some temporary code (the logging).
https://github.com/jenscski/MQTTnet/blob/45c125b86d675cc45bb88803100a91b4f411d495/Tests/MQTTnet.Core.Tests/MqttServerTests.cs#L265-L303

JanEggers: @jenscski thx for the testing method i added it to my pr and fixed the clean disconnect. not very elegant but hey it works... canyou please verify if you have the time?

chkr1011:
@JanEggers Is this finished now? I didn't follow up the whole conversation. If not please let me know when it is.

JanEggers:
@chkr1011 im done, if you are ok with the changes they can be merged

chkr1011:
I have one question left. After merging my resharper is showing this:

image

Any hints if this must be async or not?

JanEggers:
it is async but it must not be awaited in that context

FeroxRev:
Could you create a nuget RC release, so I can test this fix? Thanks!

JanEggers:
@chkr1011: yes please an rc would be very welcome

chkr1011:
Done

PR #354, opened 2018-07-05
Queue

### LINKED ISSUE #344: 2.8.0 memory leak in MqttClientSessionsManager._messageQueue
hi there the good news is that mqtt server with the new transport can handle 100k real msgs. the sad thing is that my application that is processing the events is not able to keep up. it handles 40k msgs...

that results in a memory leak in MqttClientSessionsManager._messageQueue.

so I think that this buffering should be handled similar to MqttClientPendingPacketsQueue 

also there should be a third option to block on enqueue so no messages are lost and the backpreasure propagates to the transport

JanEggers:
fixes #344

### DIFF
--- a/Source/MQTTnet/Server/AsyncQueue.cs
+++ b/Source/MQTTnet/Server/AsyncQueue.cs
@@ -9,7 +9,8 @@ namespace MQTTnet.Server
     public class AsyncQueue<T> : IDisposable
     {
         private readonly Queue<T> _queue = new Queue<T>();
-        private readonly AsyncAutoResetEvent _queueAutoResetEvent = new AsyncAutoResetEvent();
+        private readonly AsyncAutoResetEvent _queueEmptyEvent = new AsyncAutoResetEvent();
+        private readonly AsyncAutoResetEvent _queueNotFullEvent = new AsyncAutoResetEvent();
         protected readonly IMqttServerOptions _options;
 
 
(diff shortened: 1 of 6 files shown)


chkr1011:
Godd idea so far. I will merge this later after a 2.8.1 with the most important fixes is released.

What will happen if the server blocks? I mean on TCP level. The client will sent about 1000 messages and the server will block after 5 messages for example. Will the client also wait due to TCP communication magic or will a buffer overflow?

JanEggers:
as far as I know after the network buffers are full the client will be slowed down as well. its kind of like slow network

chkr1011:
Thank you. Good explanation. I remember that there is something with flow control but forgot it 😄 

But what is about the server. The server also has a publish method which directly adds the message to all subscribers. Will this method also block? I am not sure if this is correct. Probably yes because it is the desired behavior but on the other hand the call is blocked.

JanEggers:
it is a global queue so as long as the app cant keep up generating the messages is blocked / slowed down. that of course includes the server. but I guess It is less likely for the server to be blocked. what eats up time in my app is deserializing the payload and i guess it is not very likely to handle the packages you send by yourself.

chkr1011:
I am fine with the approach. I was only wondering if it makes sense with the server or if we need a "special" behavior which means "block. but not for server messages" because only one single blocking queue will block the whole server publish logic.

Source/MQTTnet/Server/AsyncQueue.cs
```diff
@@ -0,0 +1,114 @@
...
+                        case MqttPendingMessagesOverflowStrategy.DropOldestQueuedMessage:
+                            _queue.Dequeue();
+                            break;
+                        case MqttPendingMessagesOverflowStrategy.Block:
+                            wait = true;
+                            break;
+                    }
+                }
+
+                _queue.Enqueue(packet);
```
ralsu091: Is this intentional? If PendingMessagesOverflowStrategy is MqttPendingMessagesOverflowStrategy.Block, then it's possible to Add the item twice. (Line 73 and 79)

JanEggers: nope it is not

chkr1011:
Good work!

chkr1011:
@JanEggers Maybe we have a misunderstanding here. I executed all unit tests and a lot server related were no longer running.

Then I saw that the new AsyncQueue is used in the session manager. I am wondering if this is the correct place because it also contains the limits per client. So a limit of 50 per client will now limit the whole server to only 50 messages. In my opinion this has to be moved into the client pending messages queue!?

Also from performance perspective this is notable slower. My intent with the list and the thread was to ensure that all message are processed in the right order. Then the queue per client should drop messages etc.

chkr1011:
Ah I just saw that the client derives from the same queue...

JanEggers:
ouch I didnt run the tests, I will fix whatever breaks them but the change was intentional, without limiting the queue in the sessionmanager it gets flooded

chkr1011:
But then must be the setting changed. Or, the better way, a new one introduced. Because it says "MaxPendingMessagesPerClient". I would recommend a new one called "MaxPendingMessages" with a suitable default (whatever this is ... 100.000 ???).

JanEggers:
sure but first you should rollback the merge I need some time to investigate tests

chkr1011:
Do you know how? I never done this...

JanEggers:
image

reset branch to this commit 

if your git asks for a mode use hard

and then force push your local branch

JanEggers:
but be careful with force push you can break history so be shure to have it right

chkr1011:
OK thanks. I will try. In worst case we will start a new MQTT lib 😄

PR #355, opened 2018-07-07
Connection handler with websockets

JanEggers:
this enables support for websockets with the new transport


chkr1011:
Good work. I will release this with 2.8.1 as I assume this this no change with a righ risk on instabillity right?

PR #361, opened 2018-07-12
Using a own task for cleanup of worker threads.

### LINKED ISSUE #258: How to configure WSS and WS at the same?
How to configure WSS server side? Is still possible to have on WS and WSS?

```
var mqttServerOptions = new MqttServerOptionsBuilder()
                .WithConnectionBacklog(100)
                .WithDefaultEndpointPort(1883)
                .WithEncryptedEndpointPort(2883)
                .WithStorage(new RetainedMessageHandler())
                .WithSubscriptionInterceptor(
```
Thx

jenscski:
Made a attempt to fix #258 and #342 

Using an own task for cleanup of the two worker tasks in client, the usage of an own task for cleanup is due to an deadlock when the two worker threads start so wait for each other ...

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -25,10 +25,12 @@ namespace MQTTnet.Client
 
         private IMqttClientOptions _options;
         private CancellationTokenSource _cancellationTokenSource;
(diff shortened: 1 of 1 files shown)


jenscski:
This was a bad attempt to fix this deadlock

PR #364, opened 2018-07-18
Add proxy settings for WebSocket connections for .NET 451 and 461 and add option to specify the SSLProtocol and CertificateValidationCallback  in TLS options, SSLProtocol default is TLS1.2

PitySOFT:


JanEggers:
tests?

PitySOFT:
Hi, I can't make a test to check the proxy settings, as for the SSL Protocol I don't know how to test it.

I can only say that I use both changes in productions and works well.

Source/MQTTnet/Client/MqttClientOptionsBuilder.cs
```diff
@@ -76,6 +79,29 @@ public MqttClientOptionsBuilder WithTcpServer(string server, int? port = null)
             return this;
         }
 
+#if NET452 || NET461
+
```
chkr1011: Isn't this also supported for UWP?

Source/MQTTnet/Client/MqttClientOptionsBuilder.cs
```diff
@@ -86,33 +112,69 @@ public MqttClientOptionsBuilder WithWebSocketServer(string uri)
             return this;
         }
 
+
```
chkr1011: I already started to create a custom object for the parameters to avoid lots of parameters. My idea is to use it like this.
```
.WithTls(new WithTlsParameters{
    SslProtocol = Ssl3,
    CertificateValidationCallback = <whatever>,
    Certificates = LoadFromSomewhere()
});
````

All properties will have proper default values. 
I believe this is the right time to implement this because the large amout of overloads is a mess.

What do you think?

File: Source/MQTTnet/Client/MqttClientOptionsBuilder.cs [hunk 3]
```diff
@@ -86,33 +112,69 @@ public MqttClientOptionsBuilder WithWebSocketServer(string uri)
...
-                Certificates = certificates?.ToList()
+                Certificates = certificates?.ToList(),
+                SslProtocol = sslProtocol,
+                CertificateValidationCallback = certificateValidationCallback
             };
 
             return this;
         }
 
-        public MqttClientOptionsBuilder WithTls()
```
chkr1011: Isn't this a breaking change when being removed? The other methods require at least one parameter 'allowUntrustedCertificates'.  Or did I overlook something?

Source/MQTTnet/Client/MqttClientWebSocketOptions.cs
```diff
@@ -13,6 +13,10 @@ public class MqttClientWebSocketOptions : IMqttClientChannelOptions
 
         public CookieContainer CookieContainer { get; set; }
 
+#if NET452 || NET461
+        public MqttClientWebSocketProxyOptions MqttClientWebSocketProxy { get; set; } = new MqttClientWebSocketProxyOptions();
```
chkr1011: Same as above... UWP?

Source/MQTTnet/Client/MqttClientOptionsBuilder.cs
PitySOFT: I agree it's more clean this way.

Source/MQTTnet/Client/MqttClientOptionsBuilder.cs
PitySOFT: You are right, but if you make the method to use a custom object instead of parameters then this problem will go away.

Source/MQTTnet/Client/MqttClientWebSocketOptions.cs
PitySOFT: Same as above.

chkr1011:
I will merge this pull request and fix the overloads as mentioned. I will also check the UWP implementation.
Thank you for your contribution.

chkr1011:
I refactored the parameters and also the proxy usage. Please let me know if you have some doubts.

PitySOFT:
I suggest to use MqttClientWebSocketProxy instead of ProxyOptions because the proxy applies only to the websocket connection not to the standard one. So the name ProxyOptions may confuse some users that uses the normal protocol.

chkr1011:
Do you mean the options object? The class is of type WebSocketChannel so that the propery is only available when using WebSockets.

PitySOFT:
Sorry, my mistake. It's ok.

PR #368, opened 2018-07-20
Fix pause/resume of ClientKeepAliveMonitor

jenscski:
I think I have found a bug, and a fix for it.

I have really tried to make a unit test for it, but I was unsuccessful of doing so.

I was using a NodeMCU when I discovered the bug. When I just pulled the plug on the NodeMCU i did not get any disconnectes on the server. But with this change, I do get a session closed after the keep alive periode.

It's also possible to test it using MqttClient, but it requires a little bit of changing the code first, and therefore it's not possible using a unit test ...
1. Remove the send keep alive packages from MqttClient
2. At least one package must be sent from client to server, using MqttClient.PublishAsync to trigger the KeepAliveMonitor

### DIFF
--- a/Source/MQTTnet/Server/MqttClientSession.cs
+++ b/Source/MQTTnet/Server/MqttClientSession.cs
@@ -405,12 +405,12 @@ namespace MQTTnet.Server
 
         private void OnAdapterReadingPacketCompleted(object sender, EventArgs e)
         {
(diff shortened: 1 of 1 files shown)


chkr1011:
Hi, thank you for your contribution. You are right. This was a bug.
Best regards
Christian

PR #372, opened 2018-07-22
Small fix for inconsistent client name in log

jenscski:
All logging of client name is wrapped in ', but not this one line ...


chkr1011:
Good work 👍

PR #386, opened 2018-08-04
[WIP] Perf improvements

JanEggers:
hi there I cant stop fiddling with perf so here is another attempt to improve:

counters from MessageProcessingMqttConnectionContextBenchmark

before:

|             Method |     Mean |     Error |    StdDev |     Gen 0 |    Gen 1 | Allocated |
|-------------------- |---------:|----------:|----------:|----------:|---------:|----------:|
| Send_10000_Messages | 16.50 ms | 0.3514 ms | 0.7098 ms | 1890.6250 | 140.6250 |   7.94 MB |

after:

|              Method |     Mean |     Error |    StdDev |     Gen 0 |    Gen 1 | Allocated |
|-------------------- |---------:|----------:|----------:|----------:|---------:|----------:|
| Send_10000_Messages | 9.188 ms | 0.1794 ms | 0.2515 ms | 1281.2500 | 156.2500 |   5.52 MB |

main changes are:
- added body reader abstraction so the asp.net part can use its own implementation 
- use spans for reading body to avoid byte[] copy SpanBasedMqttPacketBodyReader 
- reuse body reader
- added new receive overload and provide packets with the packetcompleted event. that way the reader can read many packets on block if they are present in the buffer

x dont flush every packet to the underlying buffers so there are less network calls in total
this didnt work

if the throughput scales as well in reality im at 200k msgs/sec :)

let me know what you think


JanEggers:
added netcoreapp compilation,that gives a nice additional boost:

|              Method |     Mean |     Error |    StdDev |     Gen 0 |   Gen 1 | Allocated |
|-------------------- |---------:|----------:|----------:|----------:|--------:|----------:|
| Send_10000_Messages | 6.868 ms | 0.1353 ms | 0.2606 ms | 1617.1875 | 31.2500 |   1.76 MB |

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
```diff
@@ -60,15 +60,15 @@ private static bool TryReadBodyLength(ref ReadOnlySequence<byte> input, out int
             var index = 1;
             result = 0;
 
-            var temp = input.Slice(0, Math.Min(5, input.Length)).GetMemory();
+            var temp = input.Slice(0, Math.Min(5, input.Length)).GetMemory().Span;
```
JensNordenbro: Why Slice input to create a Span<T> only to make Memory<T> out of it only to make Span<T> of it?  Can´t you just use input.Slice directly?

JanEggers: i dont understand your point. 

I Slice the input directly, but that just gives me a shorter  ReadOnlySequence<byte> and that does not have an index operator so I need to get a span out of it

chkr1011:
Hi, 
I have no time for a review at the moment. But don't worry I will review it in a few days 😄 
Could you please have a look at the MQTTv5 branch? I changed the way how the serializer is choosen. Maybe your change requires some changes to support it. The idea is to read the first part of the _Connect_ packet to determine the serializer which should be choosen. Then the data is reverted and the serializer deserialized this initial part again to have a valid packet (for unit tests etc.)
Best regards
Christian

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
JensNordenbro: my bad :)

JanEggers:
Hi Chirtian, no worries... this isnt important right now Im just tinkering. I will try to rebase my changes as soon as im having time for it.

JanEggers:
I had a look at your changes. let me know when u integrt that in dev branch. I will rebase and adjust my code afterwards

JanEggers:
@chkr1011: can you please also merge the mqtt v5 stuff you did into dev so i can do another rebase

JanEggers:
and your code signing cert is no longer valid thats why the builds fail...

chkr1011:
I merged to develop and also fixed the certificate.

JanEggers:
thx

JanEggers:
@chkr1011 thx again christian, please do not yet review or merge, the code is working with the mqtt v5 changes but there is a negative perf impact somewhere.

I get 20 ms / 10000 packets on net461 and 11 ms on netcore before I did the changes It took 13 on net461 and 6 on netcore

Source/MQTTnet.AspnetCore/SpanBasedMqttPacketBodyReader.cs
```diff
@@ -0,0 +1,125 @@
+﻿using MQTTnet.Exceptions;
```
chkr1011: Is it possible to use this kind of Formatter also in the base lib and just use the old version for .NET 4.5.2 etc. (where not supported)?

chkr1011:
OK let me know as soon as I can have a review 😄  This new Span etc. stuff looks very interesting. Do you have a recommendation for a tutorial etc.? Everything I found is slightly theoretical and thus I don't understand it 😄

JanEggers:
have a look at:

https://adamsitnik.com/Span/

for me the most important method is slice as it provides access to a sub range without allocating a copy of that range.

Source/MQTTnet.AspnetCore/SpanBasedMqttPacketBodyReader.cs
JanEggers: If you are ok with a new dependency to System.Memory and System.Buffers i can move this to the core lib

chkr1011:
This article is much better than everything I found so far. Thanks!

Source/MQTTnet.AspnetCore/SpanBasedMqttPacketBodyReader.cs
chkr1011: After reading the article you mentioned I want to test the ArrayPool to get rid of this _singleByteBuffer_ and _fixedHeaderBuffer_ etc. If this all gives us at least a moderate performance or memory improvement I am fine with it.

JanEggers:
I am not able to resolve conflicts I will start over

PR #402, opened 2018-08-21
added AspNetMqttServerOptionsBuilder

### LINKED ISSUE #400: asp.net core 2.1 - ConfigureServices - How to build-up MQTT Server Options
Considering this code:

``` CSharp
            var mqttServerOptions = new MqttServerOptionsBuilder()
                .WithDefaultEndpoint()
                .WithDefaultEndpointPort(9876)
                .WithConnectionValidator((ctx) =>
                {
                    var mediator = serviceProvider.GetService<IMediator>();
                    if (await mediator.Send(new AuthenticateTerminalForMqttCommand(ctx.Username, ctx.Password)))
                    {
                        logger.LogInformation("Client {0} has connected.", ctx.ClientId);
                        ctx.ReturnCode = MqttConnectReturnCode.ConnectionAccepted;
                        return;
                    }

                    logger.LogInformation("Client {0} has failed to authenticate.", ctx.ClientId);
                    ctx.ReturnCode = MqttConnectReturnCode.ConnectionRefusedBadUsernameOrPassword;
                })
                .Build();
```
We cannot pull anything into `ConfigureServices()` but the `IServiceCollection` object.  So, with asp.net core 2.1, how do we provide a connection validator, among other items?  When I look at the main `Configure()` method, although we can do `app.UseMqttServer((mqttServer) => { ... custom code here ... });`, it does not seem to let us assign a custom connection validator, as the property is marked as ReadOnly at this point.

Any suggestions/ideas for how to overcome this want/need would be great.

JanEggers:
fixes #400

### DIFF
--- a/Source/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
+++ b/Source/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
@@ -21,13 +21,13 @@ namespace MQTTnet.AspNetCore
             return services;
         }
         
-        public static IServiceCollection AddHostedMqttServer(this IServiceCollection services, Action<MqttServerOptionsBuilder> configure)
-        {
-            var builder = new MqttServerOptionsBuilder();
-
-            configure(builder);
-            
-            services.AddSingleton<IMqttServerOptions>(builder.Build());
+        public static IServiceCollection AddHostedMqttServer(this IServiceCollection services, Action<AspNetMqttServerOptionsBuilder> configure)
+        {            
+            services.AddSingleton<IMqttServerOptions>(s => {
+                var builder = new AspNetMqttServerOptionsBuilder(s);
+                configure(builder);
+                return builder.Build();
+            });
 
             services.AddHostedMqttServer();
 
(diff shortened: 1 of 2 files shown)


Source/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
```diff
@@ -21,13 +21,13 @@ public static IServiceCollection AddHostedMqttServer(this IServiceCollection ser
             return services;
         }
         
-        public static IServiceCollection AddHostedMqttServer(this IServiceCollection services, Action<MqttServerOptionsBuilder> configure)
```
chkr1011: Is this a breaking change? Because the parameter is changed. If yes I would like to merge it for 2.9.0. Otherwise for 2.8.3.

dealproc: If it's a breaking change... you could possibly mark the existing one as `[Obsolete]` and still release as 2.8.3... and then folks need to change-up before [i think] 2.9 or would that be 3.0? (Trying to recall semver rules here)

JanEggers: it breaks binary compatibility but the source is compatible, if this is considered breaking I can provide another overload of that method

JanEggers:
I added back the old function and renamed the new one

PR #428, opened 2018-10-04
UnobservedTaskExceptions from a TaskCompletionSource in 2.8.3

MatthewTheThomas:
Hello,

With the 2.8.3 release of MQTTnet, our team noticed additional unobserved task exceptions being thrown from within the MQTT library. After looking into it, we found the issue to be a TaskCompletionSource that is never observed. 

This PR replaces the _disconnectReason TaskCompletionSource with a SemaphoreSlim since the tcs is never observed so each time it is disposed it will throw an UnobservedTaskException. The SemaphoreSlim is perhaps overkill when a bool would do it, but it should technically be more thread-safe (and it's less lines of code).

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -29,7 +29,7 @@ namespace MQTTnet.Client
         internal Task _keepAliveMessageSenderTask;
         private IMqttChannelAdapter _adapter;
         private bool _cleanDisconnectInitiated;
-        private TaskCompletionSource<bool> _disconnectReason;
+		private SemaphoreSlim _disconnectLimit = new SemaphoreSlim(1, 1);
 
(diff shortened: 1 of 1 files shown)


chkr1011:
Hi, thank you for your contribution. I released 2.8.4-rc1 which includes a fix. I assume that your RP is no longer required. 

Or do you see any issues in my fix?

Best regards
Christian

MatthewTheThomas:
@chkr1011 	I just checked using the 2.8.4-rc1 and it's exhibiting the same behaviors with regards to the UnobservedTaskExceptions caused by `_disconnectReason`.

chkr1011:
Ah OK now I got it. I fixed another issue 😄 In my opinion the semaphore is too heavy as you already said. I am testing an approach with an interlocked int at the moment.

chkr1011:
I released 2.8.4-rc2. Please let me know if this version works for you.

MatthewTheThomas:
@chkr1011 Perfect! That fixes it. Thanks!

PR #429, opened 2018-10-07
added ability to specify tls protocol version

### LINKED ISSUE #415: Encryption protocol support?
Hello,

It looks like the encryption protocol is hard-coded to TLS 1.2 in MqttTcpChannel.cs.  I'm curious why 1.0 and 1.1 aren't supported.  Also, when more protocols become available in the future, will it be possible to OR the enums and allow support for multiple?

JanEggers:
fixes #415


chkr1011:
Good work 👍 

Could you please also update the nuget spec with a one liner in the usual format? Then I don't have to collect them all before every release on my own 😄

PR #444, opened 2018-10-14
Referencing the MQTT Bindings for Azure Functions

keesschollaart81:
Hi, if you're interested, hereby a reference to my work on MQTT Bindings for Azure Functions where I used this library. Thanks for the excellent work. More background [over here](http://case.schollaart.net/2018/09/22/mqtt-and-azure-functions.html)


chkr1011:
Good work 👍 I read your documentation. Good project 😄

PR #471, opened 2018-11-05
updated nuget spec according to csproj

JanEggers:
Hi I just tried rc2 and it seems it does not pull in the dependencies specified in the csproj I updated them accordingly


JanEggers:
i could do my testing anyway I just needed to add the updated dependencies to my project.

PR #479, opened 2018-11-20
Fix for two issues with the managed message queue cap

PaulFake:
Hello,

Our QA team stumbled upon an issue with the storage queue allowing essentially an unlimited number of items regardless of the cap, when stress-testing the library against an unreliable network connection.  I examined the code and saw that it's possible for both the message queue and the storage queue to grow beyond the configured cap.  This can happen if there is an exception while publishing: the published message will be re-enqueued regardless of the number of items on the queue.  I also noticed that this same behavior can result in the DropOldestQueuedMessage strategy not behaving as expected, because after a message is re-enqueued, the items are no longer ordered by the time they entered the queue.  Therefore, the DropOldestQueuedMessage logic does not actually drop the oldest message, but rather just the first one in the queue.

A solution for both of these problems is to only peek the first item in the queue when publishing, rather than dequeue it.  In the nominal case, when the message has successfully published, we can then remove it from the queue.  If it fails to publish, we simply don't do anything and it will try again in the next iteration.  If a different thread knocks the current message off the queue (DropOldestQueuedMessage strategy) while it's being published, that's ok.  The code will check if it's already been removed before trying to remove it.

This method retains the original order of the queue when exceptions are encountered, and it prevents the queue from ever growing beyond the cap.


chkr1011:
Hi, thank you for your contribution. Good work 👍 
But I have one small point I want to discuss with you:
You added the IEquatable<ManagedMqttApplicationMessage> so that it supports equality checking. 
But you only check the Guid and no other properties. So to me it looks incomplete because other users may use it as well expecting a whole equality check.

So my proposal is to remove the IEquatable and just use the Guid in the function _RemoveFirstIfEqual_. I would prefer a predicate implemented as _Func<TItem, bool> so you can pass in a custom check for the Guid. 

What do you think about it?
Best regards
Christian

PaulFake:
Ok, whew.  I'm not good at Git and had to fight with it, but I think I got it all in ok.

chkr1011:
Good work. Not exactly my intention but it is OK. Probably I will refactor it a little but. Hope you agree.

PaulFake:
Sounds good.  When do you expect to build the next RC and/or release?

chkr1011:
I can do it now if you want to test it.

PaulFake:
Sure.  I can get our QA guys to test it on Monday.

PR #484, opened 2018-11-30
Feature msbuild conditions

fogzot:
The current solution and csproj files make it a bit difficult to hack on MQTTnet on non-Windows platforms where Mono is required for non-core framework support and UWP is unsupported. This patch aligns the conditions in the various csproj files to make sure that only the "officially" supported frameworks are enabled on non-Windows platforms and adds a solution that builds everything except UWP.


chkr1011:

Issue #496, opened 2018-12-13
Custom websocket request headers


cquesnel:
Hi,

I was wondering if there was a reason for which the requestHeaders for a websocket connection are not configurable through the options builder. From what I see all the code is there, and would only require an overload to `WithWebSocketServer` in `MqttClientOptionsBuilder` such as : 

```csharp
public MqttClientOptionsBuilder WithWebSocketServer(string uri, IDictionary<string, string> requestHeaders)
{
    _webSocketOptions = new MqttClientWebSocketOptions
    {
        Uri = uri,
        RequestHeaders = requestHeaders
    };

    return this;
}
```

I was asking because when using an AWS custom authorizer for validating connections, it needs some information passed through headers.

I can fork and start a pull request if you wish. 

Thanks for your help,
Charles

chkr1011:
Hi, 
yes please create a pull request. Please take care for backward compatibility when adding parameters to functions of the options builder.

This was not added because no one ever asked for it 😄 

Best regards
Christian

PR #497, opened 2018-12-13
Add custom WebSocket parameters

### LINKED ISSUE #495: Need to connect to web socket server with cookie
I need to connect to a websocket server with the outgoing initial HTTP request containing a cookie.  Is this possible?  I didn't see any options in MqttClientOptionsBuilder.

cquesnel:
Allows users to pass custom WebSocket parameters allowing to pass request headers and cookies.

Fixes #495 and #496

### DIFF
--- a/Source/MQTTnet/Client/MqttClientOptionsBuilder.cs
+++ b/Source/MQTTnet/Client/MqttClientOptionsBuilder.cs
@@ -1,7 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Net;
 using MQTTnet.Serializer;
 
 namespace MQTTnet.Client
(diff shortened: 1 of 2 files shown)


cquesnel:
@chkr1011 The build won't pass because of an expired certificate in appveyor

chkr1011:
Thank you for your contribution 👍 
Good work! But I have some things to change:

* Please use a default parameter instead of an overload. This is used for all other methods as well. So using a overload will break this pattern.
* I am not sure but is the Cookie container also available for UWP? Did you build for all supported platforms?

Best regards
Christian

cquesnel:
@chkr1011 

- I changed the method to use a default parameter
- For the cookie container, I simply exposed a field that was already present, so I'm pretty sure it's compatible. I did build the whole solution so there shouldn't be any problems there.

PR #533, opened 2019-01-17
 Add  MqttClient Connection Validator Event and  IMqttServerStorage supports injection dependencies

maikebing:


chkr1011:
Hi, thank you for your contribution.
I have some doubts because to me it seems that it covers two different things. Something in the area of storage and something in the area of client validation. Is it possible to split these changes? Or are they related and I don't get it?

I am also not happy with moving the  connection validator from the options to the server directly. What is the reason for this? Is there maybe something else missing?

Best regards
Christian

maikebing:
Added events for  the connection validator!
Existing  connection validator makes it difficult to use dependency injection.
So, I added events to make it easy to use dependent injected databases and other objects in ASP.Net Core.
IMqttServerStorage is also intended for easy use in asp.net core

chkr1011:
Sorry but I cannot accept is pull request. You moved only the ConnectionValidator because you need this. All other things are still in the options. So it breaks the consistency. 

If you want to use DI you should create your own MQTT Server class which has a proper constructor for all of your services. Then you can build the options and server as usual. You can also create a factory using the factory pattern to create your custom instance with proper DI.

Issue #527, opened 2019-01-08
Managed Client should dispose inner mqtt client


andreabalducci:
Last will message is delivered only when my application closes and not on disconnection of my managed client. 

Found the issue in ManagedClient not disposing wrapped client. 

As workaround I created the managed client without the factory, holding my reference to inner client and manually disposing after managed client.

SeppPenner:
@andreabalducci @chkr1011 Is this fixed already? I would say yes when looking at the merged pull request.

SeppPenner:
I will close this now. Feel free to open it again or contact me if this is still relevant.

PR #547, opened 2019-02-07
Fix Managed Client should dispose inner mqtt client #527

vbBerni:
Fixes the problem described in #527
Fix logging for SynchronizeSubscriptionsAsync
I changed the CreateClient method in the TestEnvironment to dispose all clients at the end

Also includes Test

### DIFF
--- a/Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
+++ b/Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
@@ -27,9 +27,11 @@ namespace MQTTnet.Extensions.ManagedClient
 
         private CancellationTokenSource _connectionCancellationToken;
         private CancellationTokenSource _publishingCancellationToken;
+        private Task _maintainConnectionTask;
 
         private ManagedMqttClientStorageManager _storageManager;
 
+        private bool _disposed = false;
         private bool _subscriptionsNotPushed;
 
         public ManagedMqttClient(IMqttClient mqttClient, IMqttNetChildLogger logger)
@@ -68,6 +70,8 @@ namespace MQTTnet.Extensions.ManagedClient
 
         public async Task StartAsync(IManagedMqttClientOptions options)
         {
+            ThrowIfDisposed();
+
             if (options == null) throw new ArgumentNullException(nameof(options));
             if (options.ClientOptions == null) throw new ArgumentException("The client options are not set.", nameof(options));
 
@@ -76,7 +80,7 @@ namespace MQTTnet.Extensions.ManagedClient
                 throw new NotSupportedException("The managed client does not support existing sessions.");
             }
 
-            if (_connectionCancellationToken != null) throw new InvalidOperationException("The managed client is already started.");
+            if (!_maintainConnectionTask?.IsCompleted ?? false) throw new InvalidOperationException("The managed client is already started.");
 
             Options = options;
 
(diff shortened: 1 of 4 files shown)


Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
```diff
@@ -193,8 +193,9 @@ public Task UnsubscribeAsync(IEnumerable<string> topics)
 
         public void Dispose()
         {
-            _connectionCancellationToken?.Dispose();
-            _publishingCancellationToken?.Dispose();
+            _mqttClient?.Dispose();
```
chkr1011: I am wondering why this is done before the Stop calls. I assume Stop should happen first. Otherwise it may happen that an already disposed object is used.

vbBerni: Yes this was very uncommon but it was required because if I cancel the _connectionCancellationToken first the MaintainConnectionAsync’s finnaly block called _mqttClient.DisconnectAsync and therefore did a regular disconnect.

vbBerni:
I added a more typical dispose behavior. Also all methods throw a ObjectDisposedExcetion now.
The Dispose and the StopAsync Methods now wait until the internal MaintainConnection-Task is finished.

vbBerni:
Any update on this?

Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
```diff
@@ -94,24 +98,32 @@ public async Task StartAsync(IManagedMqttClientOptions options)
...
             StopPublishing();
             StopMaintainingConnection();
 
             _messageQueue.Clear();
 
-            return Task.FromResult(0);
+            if (_maintainConnectionTask != null)
+            {
+                await Task.WhenAny(_maintainConnectionTask);
+                _maintainConnectionTask = null;
```
chkr1011: Why is here a "WhenAny" if there is only one? Is there a task missing in the parameter list?

Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
```diff
@@ -193,8 +211,31 @@ public Task UnsubscribeAsync(IEnumerable<string> topics)
...
+
+            _disposed = true;
+
+            StopPublishing();
+            StopMaintainingConnection();
+
+            if (_maintainConnectionTask != null)
+            {
+                Task.WaitAny(_maintainConnectionTask);
+                _maintainConnectionTask = null;
```
chkr1011: Why is here a "WhenAny" if there is only one? Is there a task missing in the parameter list?

chkr1011:
Sorry but I don't have much time at the moment. But I was able to perform a quick review. If this is fixed I will merge it.

Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
vbBerni: The big advantage of Task.WhenAny is that it only checks for completion but ignores exceptions within the task. If I use await _maintainConnectionTask it could cause an exception.

Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
vbBerni: Dispose is not async. Therefor I can’t use Task.WhenAny to ensure a task was finished. The synchronous alternative is Task.WaitAny

chkr1011:
Good work 👍

Issue #524, opened 2019-01-06
ApplicationMessageReceived isn't Async


ramonsmits:
ApplicationMessageReceived isn't Async which breaks the async flow and requires the event implementation to do `.GetAwaiter().GetResult();`.

Consider adding the ability to register a `Func<Task>` to flow the async context and also the cancellation token.

```c#
var mqttOptions = new MqttClientOptionsBuilder()
    //...
    .WithApplicationMessageReceiver(MyReceiver)  //   WithApplicationMessageReceiver(Func<MqttApplicationMessage, CancellationToken, Task>)
    //...
    .Build();
```

```c#
Task MyReceiver(MqttApplicationMessage message, CancellationToken cancellationToken)
{
     // Do async stuff
}
```

chkr1011:
Hi,
you can make the event handler async by making the method _async void_. But version 3 of this library will have support for a real async handler.

Best regards
Christian

ramonsmits:
While technically you can do `async void` that will result in data loss as the invoking of the event will then immediately return and [send the PubAck as mentioned in the other issue](https://github.com/chkr1011/MQTTnet/issues/523).

This basically results in `AtMostOnce` behavior is unreliable and as said results in data loss when errors occur.

GieltjE:
Isn't the connection between the client and server synchronous? If that is the case how much sense does it make to have an aysnc messagereceived event?

ramonsmits:
@GieltjE c# async await doesn't have a direct correlation with a process being synchronous but more with blocking IO which often happens.

dpsenner:
@ramonsmits the event handler posted in your previous comment is a worst case scenario for async code because of `GetAwaiter().GetResult()` which should instead return a `Task` to `await` on. This would not only fix deadlocks but also fix any re-entrant issues of event handlers that do async publish. As far as I can tell, the proper way to fix this is probably by adding an async api that goes through all the layers. This unfortunately also means that c# events and event handlers need a replacement because they are unsuited for async code since they usually return void or only await the first attached event handler. See [this](https://stackoverflow.com/questions/12451609/how-to-await-raising-an-eventhandler-event) for further information.

dpsenner:
I checked the implementation and the issue appears to stem from `ApplicationMessageDispatcher` having

* `public void Dispatch(Exception exception)`
* `public void Dispatch(MqttBasePacket packet)`

whereas it should probably read as:

* `public Task Dispatch(Exception exception)`
[...]

PR #550, opened 2019-02-15
Fix the MqttPacketDispatcher to not break the async flow

dpsenner:
This has the bonus side effect that the dispatcher does no longer spawn cpu intensive tasks

This patch relates to #524, but does not yet fix the api to allow async processing of application messages.

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -230,7 +230,7 @@ namespace MQTTnet.Client
             var clientWasConnected = IsConnected;
 
             InitiateDisconnect();
-            
+
             IsConnected = false;
 
             try
@@ -242,7 +242,7 @@ namespace MQTTnet.Client
 
                 await WaitForTaskAsync(_packetReceiverTask, sender).ConfigureAwait(false);
                 await WaitForTaskAsync(_keepAliveMessageSenderTask, sender).ConfigureAwait(false);
-                
+
                 _logger.Verbose("Disconnected from adapter.");
             }
             catch (Exception adapterException)
@@ -409,7 +409,7 @@ namespace MQTTnet.Client
                     _logger.Error(exception, "Unhandled exception while receiving packets.");
                 }
 
-                _packetDispatcher.Dispatch(exception);
+                await _packetDispatcher.Dispatch(exception).ConfigureAwait(false);
 
                 if (!DisconnectIsPending())
                 {
@@ -444,8 +444,7 @@ namespace MQTTnet.Client
                 return ProcessReceivedPubRelPacket(pubRelPacket, cancellationToken);
             }
 
-            _packetDispatcher.Dispatch(packet);
-            return Task.FromResult(0);
+            return _packetDispatcher.Dispatch(packet);
         }
 
         private Task ProcessReceivedPublishPacketAsync(MqttPublishPacket publishPacket, CancellationToken cancellationToken)
@@ -500,7 +499,7 @@ namespace MQTTnet.Client
             _packetReceiverTask = Task.Factory.StartNew(
                 () => ReceivePacketsAsync(cancellationToken),
                 cancellationToken,
-                TaskCreationOptions.LongRunning, 
+                TaskCreationOptions.LongRunning,
                 TaskScheduler.Default).Unwrap();
         }
 
--- a/Source/MQTTnet/Client/MqttPacketDispatcher.cs
+++ b/Source/MQTTnet/Client/MqttPacketDispatcher.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Concurrent;
+using System.Linq;
 using System.Threading.Tasks;
 using MQTTnet.Packets;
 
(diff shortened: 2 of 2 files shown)


dpsenner:
I tested the change against a dummy application that publishes subscriptions to other topics and that application appears to work just fine.

Source/MQTTnet/Client/MqttPacketDispatcher.cs
```diff
@@ -8,18 +9,29 @@ namespace MQTTnet.Client
...
 
-            _awaiters.Clear();
+            // await the tasks to complete
+            try
+            {
+                await Task.WhenAll(_awaiters.Values.Select(t => t.Task)).ConfigureAwait(false);
+            }
+            catch (Exception)
+            {
+                // swallow any exception that happens during exception handling to prevent
```
ramonsmits: why not just throw? At the minimum people should do proper exception handling but when they don't is should just crash and burn. If not, this will definitely result in super unwanted hidden errors.

dpsenner: I am happy that you are raising concern about this line. Not having this in place may result however in collateral damage. As far as I am concerned this needs thorough insights by whoever is most familiar with this. @chkr1011, would you please have a look?

I do also see an issue about the previous `Task.Run` that was put in place to avoid deadlocks. A proper async/await application does not need `Task.Run` because it completes as soon as the task completes. On the contrary, when an async/await library like MQTTnet needs `Task.Run` to avoid deadlocks it is a strong indicator that there's something fishy going on somewhere. Something along the lines of:

* `Task t; t.GetAwaiter().GetResult()`
* `Task t; t.Wait()`

I propose to remove ANY `Task.Run` invocation and fix any emerging issues. This is roughly what I planned to summarize in a separate issue as already mentioned in #551 (more specifically [this comment](https://github.com/chkr1011/MQTTnet/pull/551#issuecomment-464443340)). Such a change probably requires a refactor of the alive/send/receive mechanisms to something like:

```csharp
private async Task RunAsync(CancellationToken token) {
    var awaitPacketsToSendCancellation = new CancellationTokenSource(); // wrap in using
    // add more cancellation token sources for each task
    Task awaitPacketsToSendTask = AwaitPacketsToSendAsync(awaitPacketsToSendCancellation.Token);
    Task receiveTask = ReceiveAsync(token);
    Task awaitCancellationTask = token.AwaitCancellationAsync(); // an extension method that creates a     TaskCompletionSource and sets the result when token.Register() is called
    Task awaitSendAliveCheckTask = TimeSpan.FromSeconds(5).AwaitSendAliveAsync(token); // an extension method that does roughly Task.Delay(timeSpan) and handles OperationCancelledException gracefully
    while (!token.IsCancellationRequested) {
    Task completedTask = await Task.WhenAny(sendTask, receiveTask, awaitCancellationTask, awaitSendAliveCheckTask);
    if (completedTask == awaitPacketsToSendTask) {
        var packetsToSend = await awaitPacketsToSendTask;
        await SendPacketsAsync(packetsToSend, token);
        awaitPacketsToSendTask = AwaitPacketsToSendAsync(token);
    }
    else if (completedTask == receiveTask) { /* process received packets and restart receiveTask */ }
    else if (completedTask == awaitSendAliveCheckTask) { /* check the elapsed time since the last successful send, eventually sending an alive and then start a new await send alive check task */ }
    else if (completedTask == awaitCancellationTask) { /* terminate gracefully if token.IsCancellationRequest == true by cancelling all child tokens and terminate when the connection etc was teared down */ }
    else { /* throw */ }
}
```

@ramonsmits What do you think?

ramonsmits: Task.Run must indeed not be used to resolve deadlock issues however there can be valid reasons for using it with CPU bound would still benefit from parallelism. However, this is applicable to the actual processing of the message not really the framework. 

To me is seems seems a bit strange to manage as individual tasks but I'm not familiar enough with this code base but the whole `else if` nesting doesn't feel really very clean.

dpsenner: At the end of the day, if MQTTnet calls `Task.Run` the processing of messages ends happens in parallel across multiple threads even when the library has to maintain the absolute ordering of messages even across multiple topics.

Thanks for you valuable insights regarding the pseudo-code posted earlier. It may be possible to start several tasks and each task loops until it is cancelled. I have not implemented that pattern until now but only followed the above implementation pattern - mainly because it works. In any case the code should gracefully handle the following situation:

1. The broker sends a PUB
1. The PUB is received by the receive task
1. The receive task parses the message and calls the new api, allowing the application to processes the PUB; the code awaits until the async operation completes
1. the application code would then do several application logic stuff and eventually do several PUB by doing mqttClient.PublishAsync() and await that task to complete.
1. The await packets to send task would at this point complete and send the packets to be sent
1. Only when this completes, the processing of the PUB completes

An interesting question that emerges is: when should PUBACK for QoS 1 (or PUBREL for Qos 2) be sent? There are two candidates:

1. It be sent before the application code is allowed to handle the PUB
1. It is sent afterwards when the application code gracefully processed the PUB

The second solution makes the application re-entrant for the broker and that may not be desired. At the same time it would cause the broker to resend a PUB until the client gracefully handles the PUB.

ramonsmits: Lets not discuss multiple things here :-), I don't think the discussion on PUBACK belongs here. Shall we create a separate issue for that?

dpsenner:
I am closing the PR. All these changes need more discussion in an issue and that issue can then serve as a synchronization point for related future pull requests.

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -500,7 +499,7 @@ private void StartReceivingPackets(CancellationToken cancellationToken)
             _packetReceiverTask = Task.Factory.StartNew(
                 () => ReceivePacketsAsync(cancellationToken),
                 cancellationToken,
-                TaskCreationOptions.LongRunning, 
+                TaskCreationOptions.LongRunning,
```
ramonsmits: TaskCreationOptions.LongRunning is pretty useless, just use `Task.Run` as `LongRunning` forces the creation of a thread which isn't specifically for all processing.

See http://blog.i3arnon.com/2015/07/02/task-run-long-running/ for more info.

PR #551, opened 2019-02-15
Add async handler interfaces for connected and disconnected to allow async code on connect

dpsenner:
This commit aligns connect and disconnect with the way how application message was implemented.

### DIFF
--- a/Source/MQTTnet.Extensions.ManagedClient/IManagedMqttClient.cs
+++ b/Source/MQTTnet.Extensions.ManagedClient/IManagedMqttClient.cs
@@ -13,7 +13,12 @@ namespace MQTTnet.Extensions.ManagedClient
         int PendingApplicationMessagesCount { get; }
         IManagedMqttClientOptions Options { get; }
 
+        IMqttClientConnectedHandler ConnectedHandler { get; set; }
+        [Obsolete("Use ConnectedHandler instead.")]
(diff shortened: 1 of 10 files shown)


chkr1011:
Hi,
thank you for your contribution. This is exactly what is missing for the next alpha release. Good work!

But there are some other events missing. For the upcoming release 3.0.0 I want to remove all classic event handlers. Do you want to replace the others as well (following the same pattern)? If you don't have  time etc. I can do it as well.

Best regards
Christian

dpsenner:
My actual plan is to make mqttnet work with full async/await support in a single-thread. There are many things to refactor to achieve this goal and we'll most likely have to collaborate. I'm glad you see this as a step towards the future and as such, feel free to merge it upstream. We can do several smaller pull requests and work us gradually into the future.

chkr1011:
What exactly do you mean with single thread? Apart from the server (where something is still not async) the client is running in one thread mostly. Only the keep alive is a separate thread (but async). Only the message queue from the managed client is remaining.

dpsenner:
It feels like expanding that story in this pull request is off-topic. I have the intention to create an issue where we can discuss the topic. I am going to need at least an hour to write down something that makes sense not only in my head. Do you prefer the discussion to happen in here?

chkr1011:
I would prefer a new Ticket.

PR #553, opened 2019-02-18
Add extension methods that provide a simpler api for IMqttClient and IManagedMqttClient

dpsenner:
This PR adds extension methods on connect/disconnect that provide a non-async/await api in cases where the handlers of connect/disconnect are not async.


chkr1011:
Thank you for your contribution. Good work! 👍

PR #555, opened 2019-02-19
When no handlers are configured the client should not blow up

### LINKED ISSUE #554: Elvis operator will result in null reference exceptions with await
Hi

While trying out the dev branch I ran into this:

` await ConnectedHandler?.HandleConnectedAsync(new MqttClientConnectedEventArgs(authenticateResult));`

using **?** together with await will result in a _NullReferenceException_.

Reference: https://stackoverflow.com/questions/40436213/why-elvis-operator-doesnt-work-with-async-await

BennyM:
I fixed #554 and added tests for it.

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -101,7 +101,11 @@ namespace MQTTnet.Client
                 _logger.Info("Connected.");
 
                 Connected?.Invoke(this, new MqttClientConnectedEventArgs(authenticateResult));
-                await ConnectedHandler?.HandleConnectedAsync(new MqttClientConnectedEventArgs(authenticateResult));
+                if(ConnectedHandler != null)
+                {
+                    await ConnectedHandler.HandleConnectedAsync(new MqttClientConnectedEventArgs(authenticateResult));
+                }
+               
 
                 return authenticateResult;
             }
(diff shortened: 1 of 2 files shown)


Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -101,7 +101,11 @@ public async Task<MqttClientAuthenticateResult> ConnectAsync(IMqttClientOptions
                 _logger.Info("Connected.");
 
                 Connected?.Invoke(this, new MqttClientConnectedEventArgs(authenticateResult));
-                await ConnectedHandler?.HandleConnectedAsync(new MqttClientConnectedEventArgs(authenticateResult));
+                if(ConnectedHandler != null)
+                {
```
chkr1011: Please move the handler to a variable first. Because it may happen that the property is checked in the if statement and then the value is set to null. This is only a theoretical thing but also a common pattern for event handlers (before ? was available).

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -265,7 +269,10 @@ private async Task DisconnectInternalAsync(Task sender, Exception exception)
 
                 _logger.Info("Disconnected.");
                 Disconnected?.Invoke(this, new MqttClientDisconnectedEventArgs(clientWasConnected, exception));
-                await DisconnectedHandler?.HandleDisconnectedAsync(new MqttClientDisconnectedEventArgs(clientWasConnected, exception));
+                if(DisconnectedHandler != null)
+                {
```
chkr1011: Same as above.

PR #592, opened 2019-04-15
performance improvements span based serializers

JanEggers:
I cannot compare with before because the current dev branch is not able to run the benchmark.

see commit "fixed another issue", seems the mqtt v5 changes introduced an issue in the aspnet part of the codebase.

this is what it looks like after my changes:

|              Method | Runtime |     Toolchain |     Mean |     Error |    StdDev |     Gen 0 |    Gen 1 |   Gen 2 | Allocated |
|-------------------- |-------- |-------------- |---------:|----------:|----------:|----------:|---------:|--------:|----------:|
| Send_10000_Messages |     Clr |       Default | 25.30 ms | 0.4913 ms | 0.6213 ms | 2562.5000 |  93.7500 | 31.2500 |   10.7 MB |
| Send_10000_Messages |    Core | .NET Core 2.1 | 12.19 ms | 0.3646 ms | 1.0225 ms | 1718.7500 | 546.8750 | 31.2500 |   1.45 MB |

vs 2019 includes an improved profiler it seems with it I can spot many more allocations per message I would like to get rid of. most of them require changes of apis so I left them like they are and just focused on reducing the number of byte[] allocations.

ValueTask instead of Task was one option I saw benifit in my first attempt but to keep the scope of the change smaller I left that like it is. 

image

@chkr1011 pls do a review asap. I dont want the branch to get stale again.

### DIFF
--- a/Source/MQTTnet.AspnetCore/Client/MqttClientConnectionContextFactory.cs
+++ b/Source/MQTTnet.AspnetCore/Client/MqttClientConnectionContextFactory.cs
@@ -21,7 +21,10 @@ namespace MQTTnet.AspNetCore.Client
                     {
                         var endpoint = new DnsEndPoint(tcpOptions.Server, tcpOptions.GetPort());
                         var tcpConnection = new TcpConnection(endpoint);
-                        return new MqttConnectionContext(new MqttPacketFormatterAdapter(options.ProtocolVersion), tcpConnection);
+
+                        var writer = new SpanBasedMqttPacketWriter();
+                        var formatter = new MqttPacketFormatterAdapter(options.ProtocolVersion, writer);
+                        return new MqttConnectionContext(formatter, tcpConnection);
                     }
                 default:
                     {
--- a/Source/MQTTnet.AspnetCore/Client/Tcp/SocketSender.cs
+++ b/Source/MQTTnet.AspnetCore/Client/Tcp/SocketSender.cs
@@ -5,6 +5,10 @@ using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Net.Sockets;
 
+#if NETCOREAPP2_1
+using System.Runtime.InteropServices;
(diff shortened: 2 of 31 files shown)


chkr1011:
Hi, 
ValueTasks look interesting. Didn't know that this exists...
What do you plan for API changes? I want to release 3.0 soon and keep the API stable for a while.

JanEggers:
> What do you plan for API changes? I want to release 3.0 soon and keep the API stable for a while.

thought so, that is why I didnt change em. basic idea is to get as close as possible to zero allocations / packet. I understand that we need some of the stuff on the screenshot but I wonder if we can get rid of the collections and hashsets. In general I want to speed up processing of messages I have the broker running on an embedded device and currently im at 50% i need to reduce that as much as possible.

> What do you think about a .net standard 2.0 library version?
I dont know if we would gain anything from that. We need AspNetCore (Kestrel) to get the pipelines based tcp / websocket server. I dont see a benifit in implementing that ourself. What is your motivation? 

> at least for users which are able to use latest and greatest versions.

Im not sure how microsoft / netstandard will evolve. I raised an Issue about a new features in System.Memory.dll (about a SequenceReader) but they said they will not release it as seperate nuget only part of netcore 3.0 and that will not be compatible with netfx / netstandard 2.1. for me thats a big downside as I thought with dotnet core the monolith framework was properly broken down, but right now it looks like the same mess as netfx, just cross platform.

chkr1011:
Yes I was hoping the pipelines etc. are available but if this is not possible then we don't need it. 

When it comes to allocations I also tried to initialize all properties (in mqtt5) with null by default but at some points it was not possible or too much effort. But we can try it again. The decoder can create them on demand (if user properties etc. are available). 

Did you saw the latest added question?

I added some TODOs in MqttConnectionContext. Do you have an idea how to fill them properly?

PR #594, opened 2019-04-17
Less allocations

JanEggers:
image

whats the point of https://github.com/chkr1011/MQTTnet/blob/develop/Source/MQTTnet/Client/Receiving/MqttApplicationMessageReceivedHandlerDelegate.cs

why not use events like before? if the intention was to not block the server if the eventhander is expensive / blocking we still need to add task.run otherwise it still blocks.

I would like to use events to distribute the packets to sessions as it seems the foreach over the concurrentdictionary allocates lists and arrays.

this bit:

https://github.com/chkr1011/MQTTnet/blob/420b35e37d9eea2cd0126ab74b7f5e7d36055c73/Source/MQTTnet/Server/MqttClientSessionsManager.cs#L207-L213

before:

|              Method |     Mean |     Error |   StdDev |     Gen 0 |    Gen 1 |   Gen 2 | Allocated |
|-------------------- |---------:|----------:|---------:|----------:|---------:|--------:|----------:|
| Send_10000_Messages | 12.19 ms | 0.3646 ms | 1.0225 ms | 1718.7500 | 546.8750 | 31.2500 | 1.45 MB

after:

|              Method |     Mean |     Error |   StdDev |     Gen 0 |    Gen 1 |   Gen 2 | Allocated |
|-------------------- |---------:|----------:|---------:|----------:|---------:|--------:|----------:|
| Send_10000_Messages | 11.81 ms | 0.3674 ms | 1.083 ms | 1500.0000 | 375.0000 | 31.2500 |   1.22 MB |


chkr1011:
Please don't just delete newly added classes like "MqttClientPublishResult". This class is still missing some properties like the reason string and the user properties from the original ACK packet. Removing that allows no access to the results from the ACK packet.

I was wondering if it is possible to reuse the "ReceivedMqttPacket" maybe because this is generated for every received message but in the end it is used only for deserialization. After that it can be maybe reused. But I am not sure. That are just thoughts.

JanEggers:
> Please don't just delete newly added classes like "MqttClientPublishResult"

oki il revert that. I just didnt know the intention of the class....

chkr1011:
This approach tries to cover multiple things.
1. Allow using async methods and wait for the task to keep message ordering
2. Still choose between sync and async if required (wrapped in every delegate class)
3. Allow implementing some interfaces and set same class as handler for different things. Just a style thing than using += and -= etc.
This should not be used in server internals. Why do you think so?

The problem with events is proper attach and detach. Also the Enqueue... method was async for some time. NOW I agree that it may can be changed to an event.

JanEggers:
with valuetask:

image


|              Method |     Mean |     Error |    StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|-------------------- |---------:|----------:|----------:|---------:|---------:|------:|----------:|
| Send_10000_Messages | 10.40 ms | 0.2157 ms | 0.6361 ms | 765.6250 | 265.6250 |     - |  547.9 KB |

chkr1011:
Looks promising but is ValueTask also supported in 4.5.2 and UWP?

JanEggers:
yup just requires a reference to System.Threading.Tasks.Extensions

chkr1011:
How long do you want to work on this? I want to release 3.0.0 soon so I would prefer 1) no critical and fundamental changes but 2) still all required API changes (if required) to avoid changed APIs next release. But I don't know how long you will need and whats on your list (if there is any list).

JanEggers:
I put back MqttClientPublishResult so this is ready for review. I am just tinkering and the gain isnt great enough that this kind of changes block the release of 3.0.0

VladimirAkopyan:
Grad to see your guys using  ValueTask
I will write a blog post about this library at about the time you release new version!

chkr1011:
Is there any progress on this? I would like to close it. We can fur sure reopen it if there is any progress but I always think there is a "new" merge request but there is not 😄

JanEggers:
ill close it for now

PR #693, opened 2019-07-09
fixed #691

### LINKED ISSUE #691: Sending Large packets with AspnetCore based connection throws System.ArgumentException: Destination is too short
when sending packets bigger than 4096 bytes this error occours.

this is the source of evil:

https://github.com/chkr1011/MQTTnet/blob/1847af13d2023341dbb01d41730212f544b57f98/Source/MQTTnet.AspnetCore/MqttConnectionContext.cs#L165

JanEggers:
fixed #691 

@chkr1011  I would be very happy if you could review this and create a prerelease package asap. 

We switched one of our apps to dotnet core and hit this issue.

### DIFF
--- a/Source/MQTTnet.AspnetCore/MqttConnectionContext.cs
+++ b/Source/MQTTnet.AspnetCore/MqttConnectionContext.cs
@@ -162,11 +162,12 @@ namespace MQTTnet.AspNetCore
                 var buffer = formatter.Encode(packet);
                 var msg = buffer.AsMemory();
                 var output = _output;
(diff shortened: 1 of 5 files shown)


JanEggers:
@davidfowl is there a better way to emulate the behavior of the SlabMemoryPool used in AspNetCore?

davidfowl:
This is actually fixed in later versions of pipelines and 3.0. If you ask for more than the pool can give m, it’ll allocate outside of the pool

chkr1011:
Good work 👍 I will create a preview nuget now.

PR #702, opened 2019-07-15
X509Certificate convert to X509Certificate2 in ClientCertificate

rudacs:
Fix to convert from X509Certificate to X509Certificate2.


chkr1011:

PR #703, opened 2019-07-15
Exposing MqttClientConnection in incoming message

rudacs:
Exposing MqttClientConnection (Certificate, Username and Password data) on receiving the message (Broker)

### DIFF
--- a/Source/MQTTnet/Server/MqttApplicationMessageInterceptorContext.cs
+++ b/Source/MQTTnet/Server/MqttApplicationMessageInterceptorContext.cs
@@ -2,15 +2,18 @@
 {
     public class MqttApplicationMessageInterceptorContext
     {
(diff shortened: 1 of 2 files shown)


chkr1011:
Thank you for your contribution. I can not merge this because it breaks some rules of that lib. The MQTT connection is an internal thing and should not be exposed to public API. 

But I think I understand the idea behind that. Please try to solve your use case by using the new session items. There you can store every information you need. Also the whole connection validator context if you like.

SeppPenner:
@rudacs You can find a description of the `SessionItems` object here: https://github.com/chkr1011/MQTTnet/wiki/Server#storing-data-in-the-session. It is avaiable from the version [3.0.6-beta1](https://www.nuget.org/packages/MQTTnet/3.0.6-beta1).

PR #709, opened 2019-07-17
Fixed storage queue race condition

PaulFake:


chkr1011:
Is there a chance to add a UnitTest for that to make sure it will not break again in the future?

PaulFake:
@chkr1011 That's a good idea.  I'll look into it next week.

PaulFake:
@chkr1011 Unit testing this issue is an interesting problem.  It's not something that can reliably be reproduced, so no test can conclusively prove there *isn't* a race condition in the code.  The best we can do is a kind of heuristic approach that publishes for a few minutes and monitors the storage queue to ensure it goes down to zero (or something along these lines).  Is this good enough for you?

chkr1011:
For me it would be fine to check the overall flow (roughly points). 
1. Setup everyting.
2. Connect with server.
3. STOP server when client is connected.
4. Publish message.
5. Check for existence in storage after 500 ms.
6. Start server.
7. Wait for connect.
8. Check if the message is properly removed from the storage.

This can be done using the _TestEnvironment_.

chkr1011:

PR #712, opened 2019-07-19
fix typo in icon

mseyock:
Fixed a typo inside the icon. 
MQQT => MQTT


chkr1011:
Could you please have a look at the nuget page? To it looks like that there is something wrong with the shadow and the corners. I assume we need the "low detail" version of the icon without the shadow!?

mseyock:
@chkr1011 i will take a look at the nuget version. I'll remove the shadow at the beginning of the next week and will show you the result before creating a merge request. Thanks

chkr1011:
I have no idea why but since the typo fix the issue for nuget packages is gone. Maybe also a caching issue. I assume there is no change required anymore.

PR #718, opened 2019-07-23
Fixed AsyncLock::WaitAsync cancellation bug

PaulFake:
There isn't any code in MQTTnet that actually uses a cancellation token in WaitAsync, so this is more of a preventative thing than a bug fix.  The original code just checks if the task was completed, not whether it was cancelled.  If the process is cancelled immediately before the call to WaitAsync, it'll return as normal (https://referencesource.microsoft.com/#mscorlib/system/threading/SemaphoreSlim.cs,612) rather than throw a cancellation exception.  This change will ensure we only return the releaser if the wait actually ran to completion rather than exited early due to cancellation.  I've tested this, and it properly throws a cancellation exception later on.  (Credit for this fix goes to my co-worker @mwinterb)

### DIFF
--- a/Source/MQTTnet/Internal/AsyncLock.cs
+++ b/Source/MQTTnet/Internal/AsyncLock.cs
@@ -23,7 +23,7 @@ namespace MQTTnet.Internal
         public Task<IDisposable> WaitAsync(CancellationToken cancellationToken)
         {
             var task = _semaphore.WaitAsync(cancellationToken);
-            if (task.IsCompleted)
+            if (task.Status == TaskStatus.RanToCompletion)
             {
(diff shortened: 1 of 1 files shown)


PaulFake:
I'll look at the build failure tomorrow

SeppPenner:
> Starting test execution, please wait...
Failed   Create_Managed_Client_With_Logger
Error Message:
 Assert.IsFalse failed. 
Stack Trace:

There is some failing test. I don't know if it's failing because of your change though.

PaulFake:
I ran that test locally with my change compiled, and it passed three times

chkr1011:

PR #729, opened 2019-07-31
Add Server certificate password and client certificate builders

cslutgen:
Exposed interfaces in builder pattern to provide support to configure an optional password to be used when loading a server certificate from a file that is password protected.

Added support to enable Client Certificate exchange via builder pattern.

### DIFF
--- a/Source/MQTTnet/Server/IMqttServerCredentials.cs
+++ b/Source/MQTTnet/Server/IMqttServerCredentials.cs
@@ -1,0 +1,6 @@
+using System;
+
+public interface IMqttServerCredentials
+{
+    String Password { get; }
+}
--- a/Source/MQTTnet/Client/Options/MqttClientOptionsBuilder.cs
+++ b/Source/MQTTnet/Client/Options/MqttClientOptionsBuilder.cs
@@ -139,6 +139,13 @@ namespace MQTTnet.Client.Options
             return this;
         }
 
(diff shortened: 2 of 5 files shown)


Source/MQTTnet/Server/IMqttServerCredentials.cs
```diff
@@ -0,0 +1,6 @@
+using System;
+
+public interface IMqttServerCredentials
+{
+    String Password { get; }
```
chkr1011: Why did you call this _Password_? I am also wondering why this is added to an interface? Now it looks a little bit strange because of _Password.Password_ etc. Isn't it enough to just add a _CertificatePassword_ property of type _string_?

cslutgen: I was attempting to keep the same semantics as used on the IMqttClientCredentials. In using an interface, it allows the user to provide a concrete class that can keep the credentials secured. This was the same logic as I understood used to keep the IMqttClientCredentials secured by whatever class the user wanted to use, i.e. a class wrapping SecureString, etc.

Regarding the property name, I would be happy to change the name to TlsEndpointOptions.CertificatePassword if that is what you are suggesting?

Thus the call to use it would become options.TlsEndpointOptions.CertificatePassword.Password

cslutgen: @chkr1011 I uploaded an addition commit to change the property name to CertificateCredentials.Password. Let me know if this is amenable to you.

PR #740, opened 2019-08-12
Fix Client SslStream CRL verification

cslutgen:
Logic appears to be inverted. When IgnoreCertificateRevocationErrors, SslStream was requested to checkCertificateRevocation which were then ignored in the callout validator.

Also, it looks like a managed object dispose was missed, a little cleanup.


chkr1011:
Thank you for your contribution. Good work 👍 I will release 3.0.7 soon which will include this fix.

PR #742, opened 2019-08-13
Fixed null reference error with certificate credentials.

SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/741.

### DIFF
--- a/Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
+++ b/Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
@@ -48,7 +48,7 @@ namespace MQTTnet.Implementations
                     throw new ArgumentException("TLS certificate is not set.");
                 }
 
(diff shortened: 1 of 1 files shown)


SeppPenner:
Ok, this made two tests fail as well:

```log
Starting test execution, please wait...
  X Create_Managed_Client_With_Logger [855ms]
  Error Message:
   Assert.IsFalse failed. 
  Stack Trace:
  X Send_Long_Body [964ms]
  Error Message:
   Assert.IsTrue failed. 
  Stack Trace:
     at MQTTnet.Tests.Server_Tests.Send_Long_Body() in C:\projects\mqttnet\Tests\MQTTnet.Core.Tests\Server_Tests.cs:line 884
Test Run Failed.
```

chkr1011:
I implemented a fix for this. The problem is that we need to call another constructor. Your fix will lead to a _null_ password which is not valid when using the overload with a password parameter.

SeppPenner:
> I implemented a fix for this. The problem is that we need to call another constructor. Your fix will lead to a null password which is not valid when using the overload with a password parameter.

I was already quite sure that this wasn't the correct fix 😄

Issue #745, opened 2019-08-16
MQTTv5 - The AssignedClientId is ignored in ClientConnected, ClientDisconnected, and is never sent when connection is successful


bcrosnier:
### Describe the bug
I've been trying to use server-assigned ClientIds, as designed in MQTT v5.0 3.1.3.1 Client Identifier (ClientID).

I'd like your advice on normative statement MQTT-3.1.3-7, stating:
> It MUST then process the CONNECT packet as if the Client had provided that unique ClientID, and MUST return the Assigned Client Identifier in the CONNACK packet [MQTT-3.1.3-7].

At time of writing, while it is possible to set `AssignedClientIdentifier` in `MqttConnectionValidatorContext`, which then [replaces `connectPacket.ClientId`](https://github.com/chkr1011/MQTTnet/blob/master/Source/MQTTnet/Server/MqttClientSessionsManager.cs#L308), `HandleClientAsync()` will still use [the original `clientId`](https://github.com/chkr1011/MQTTnet/blob/master/Source/MQTTnet/Server/MqttClientSessionsManager.cs#L243) (which is always empty, which is we we're using `AssignedClientIdentifier`).

This empty `clientId` is then used when calling `HandleClientConnectedAsync()` and `HandleClientDisconnectedAsync()`.

This makes it awkward to properly identify clients from a `ClientConnectedHandler` or a ClientDisconnectedHandler`.

Furthermore, in the event of a successful connection, the client is never informed of its `AssignedClientIdentifier`: the `CONNACK` packet [does not, in fact, have any of the v5 Properties](https://github.com/chkr1011/MQTTnet/blob/master/Source/MQTTnet/Server/MqttClientConnection.cs#L144).

Interestingly enough, on a non-successful response, the v5 Properties are properly sent: [`DataConverter.CreateConnAckPacket()` is used](https://github.com/chkr1011/MQTTnet/blob/master/Source/MQTTnet/Server/MqttClientSessionsManager.cs#L251) instead of `new MqttConnAckPacket()`.

To sum up:
- When using `AssignedClientIdentifier`, the `ClientId` available in `ClientConnectedHandler` and `ClientDisconnectedHandler` is always empty, even if `AssignedClientIdentifier` is set. Is this normal?
- The `CONNACK` packet from a successful authentication does not have any of the v5 Properties, whereas the one from a non-successful authentication does. Is this normal?

[...]

PR #746, opened 2019-08-19
Better handling of AssignedClientIdentifier - chkr1011/MQTTnet#745

bcrosnier:
Allow AssignedClientIdentifier in ClientConnected, ClientDisconnected and ConnectedHandler.
Fixes chkr1011/MQTTnet#745

### DIFF
--- a/Source/MQTTnet/Server/MqttClientSessionsManager.cs
+++ b/Source/MQTTnet/Server/MqttClientSessionsManager.cs
@@ -240,9 +240,9 @@ namespace MQTTnet.Server
                     return;
                 }
 
-                clientId = connectPacket.ClientId;
-
                 var connectionValidatorContext = await ValidateConnectionAsync(connectPacket, channelAdapter).ConfigureAwait(false);
+
+                clientId = connectPacket.ClientId;
 
                 if (connectionValidatorContext.ReasonCode != MqttConnectReasonCode.Success)
                 {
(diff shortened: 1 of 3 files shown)


bcrosnier:
As written in #745 - this changes the public signature of `MqttClientConnection.RunAsync()` to pass the `MqttConnectionValidatorContext`, so that it can create the `MqttConnAckPacket` using its own `IMqttDataConverter`.

There may be better ways to do this that I didn't think of.

This should be compatible with MQTT v3, as `ValidateConnectionAsync()` explicitly checks for v5 before using the `AssignedClientIdentifier` - and the `IMqttDataConverter` should already be the one matching the protocol version.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -240,10 +240,10 @@ private async Task HandleClientAsync(IMqttChannelAdapter channelAdapter, Cancell
                     return;
                 }
 
-                clientId = connectPacket.ClientId;
-
                 var connectionValidatorContext = await ValidateConnectionAsync(connectPacket, channelAdapter).ConfigureAwait(false);
 
+                clientId = connectPacket.ClientId;
```
chkr1011: I am not sure if this is correct. The Client ID from a CONNECT packet might be null. But the assigned client ID is set in another property. Isn't it required to use either ClientId OR AssignedClientId (in can the first one is null/empty)? Or am I missing something?

bcrosnier: You're right: either the client ID or the assigned client ID must be used, and the former only if the latter is null or empty.
In this case, however, this particular check is done by the call to ValidateConnectionAsync() above, which may (or may not) have modified connectPacket.ClientId [on line 308](https://github.com/chkr1011/MQTTnet/blob/1a41415dd165560cff97bb4ab52592133a013c55/Source/MQTTnet/Server/MqttClientSessionsManager.cs#L308).

PR #763, opened 2019-09-30
Fix for #762 - thread pool leak

### LINKED ISSUE #762: Thread pool leak
### Describe the bug
`_messageQueue.PeekAndWait()` (line 352, ManagedMqttClient.cs) causes thread leak due to lock on `ManualResetEvent` when a connection is already dead. Multiple reconnections can cause a thread pool exhaustion (especially on Xamarin.Android)

- ManagedClient

### To Reproduce
1. Using this version of MQTTnet '3.0.8'.
2. Start/stop the client repeatedly
3. See the thread stacks full of `BlockingQueue.PeekAndWait()`

### Expected behavior
Valid cancellation of `PeekAndWait`

### Screenshots
image

This can be easily fixed by using a `ManualResetEventSlim` which supports a `CancellationToken`

zawodskoj:
Fixes #762

### DIFF
--- a/Source/MQTTnet/Internal/BlockingQueue.cs
+++ b/Source/MQTTnet/Internal/BlockingQueue.cs
@@ -8,7 +8,7 @@ namespace MQTTnet.Internal
     {
         private readonly object _syncRoot = new object();
         private readonly LinkedList<TItem> _items = new LinkedList<TItem>();
-        private readonly ManualResetEvent _gate = new ManualResetEvent(false);
+        private readonly ManualResetEventSlim _gate = new ManualResetEventSlim(false);
 
         public int Count
         {
@@ -32,7 +32,7 @@ namespace MQTTnet.Internal
             }
         }
 
-        public TItem Dequeue()
+        public TItem Dequeue(CancellationToken cancellationToken = default)
         {
             while (true)
             {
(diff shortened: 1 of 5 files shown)


zawodskoj:
There are strange failed test. Should we leave it as is?

SeppPenner:
> There is strange failed test. Should we leave it as is?

Which one is it? I haven't run the build process yet.

zawodskoj:
> Which one is it? I haven't run the build process yet.

```
 X Create_Managed_Client_With_Logger [862ms]
  Error Message:
   Assert.IsFalse failed. 
  Stack Trace:
```

SeppPenner:
Ok, I will check this later.

Source/MQTTnet/Internal/BlockingQueue.cs
```diff
@@ -8,7 +8,7 @@ public class BlockingQueue<TItem>
     {
         private readonly object _syncRoot = new object();
         private readonly LinkedList<TItem> _items = new LinkedList<TItem>();
-        private readonly ManualResetEvent _gate = new ManualResetEvent(false);
+        private readonly ManualResetEventSlim _gate = new ManualResetEventSlim(false);
```
chkr1011: ManualResetEventSlim is Disposable. So please implement the Dispose interface. Also parent classes might need to call Dispose then as well.

zawodskoj: Ok, i'll do it later

chkr1011: I have some time at the moment. I will do. Hope you did not start already 🙈

PR #782, opened 2019-10-23
Reduce subscription publishing loglevel in ManagedMqttClient

cstichlberger:
- the Info level for publishing subscriptions is quite noisy
 - previously this wasn't a problem because subscription processing was buffered, now the message is logged for nearly every single un/subscription that is made
 - all other frequently recurring log events are also on level Verbose
 - include some additional info when a log entry is made


SeppPenner:
@chkr1011 What do you think? This seems to be a good idea if you ask me.

PR #797, opened 2019-11-15
Add an x509 certificate parameter

### LINKED ISSUE #796: Setup encryption with X509Certificate directly
I would like a way to use an installed, non-exportable, certificate. I don't see an overload that takes an X509Certificate directly, but it looks like one is just being built internally from the bytes provided in `.WithEncryptionCertificate()`.

- Server

I would like to add an overload to MqttServerOptionsBuilder.WithEncryptionCertificate() that takes an X509Certificate directly

I'm happy to implement this if it's an acceptable change to the API

borigas:
Closes #796 

Adds a way to set the server's certificate using an `X509Certificate2` directly instead of needing to export it to `byte[]`.

### DIFF
--- a/Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
+++ b/Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
@@ -43,13 +43,17 @@ namespace MQTTnet.Implementations
 
             if (options.TlsEndpointOptions?.IsEnabled == true)
             {
-                if (options.TlsEndpointOptions.Certificate == null)
+                if (options.TlsEndpointOptions.Certificate == null && options.TlsEndpointOptions.X509Certificate == null)
                 {
                     throw new ArgumentException("TLS certificate is not set.");
                 }
 
                 X509Certificate2 tlsCertificate;
-                if (string.IsNullOrEmpty(options.TlsEndpointOptions.CertificateCredentials?.Password))
+                if (options.TlsEndpointOptions.X509Certificate != null)
+                {
+                    tlsCertificate = options.TlsEndpointOptions.X509Certificate;
+                }
+                else if (string.IsNullOrEmpty(options.TlsEndpointOptions.CertificateCredentials?.Password))
                 {
                     // Use a different overload when no password is specified. Otherwise the constructor will fail.
                     tlsCertificate = new X509Certificate2(options.TlsEndpointOptions.Certificate);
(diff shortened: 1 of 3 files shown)


borigas:
I don't have UWP installed, so I haven't tested explicitly. I think it should behave the same though. The change is just that it exposes the `X509Certificate2` to the public API instead of building it internally from a `byte[]`

chkr1011:
The X509Certificate  is not available in UWP (it has class "Certificate" instead) and the nuget cannot be referenced. So we have to find another solution. That is also the reason why I decided to use byte[] in the current version.

borigas:
Apologies, I hadn't realized that there was a UWP specific implementation of `MqttTcpServerAdapter` and assumed because `X509Certificate2` was being used already in the standard MqttTcpServerAdapter, it was ok everywhere.

Do you have suggestions on the best way to add support for using a non-exportable certificate from the Windows certificate manager? This is a pretty high priority item for me, so I would love to find a solution.

A couple ideas. Let me know if you have preferences or better ideas:
1. Try/catch the UWP version and do something like https://github.com/chkr1011/MQTTnet/blob/08d1a57aa22f79a8f0be65d094cc6cb6eaf203c9/Source/MQTTnet/Implementations/MqttTcpServerAdapter.Uwp.cs#L84?
2. Could we add a selector function?

Sorry if those are dumb ideas. I'm not familiar with the UWP API limitations, so I'm not sure if they'd actually solve the problem

borigas:
I installed UWP this morning to test this better and everything compiles and tests pass. Again, I'm not very familiar with UWP limitations so I could be missing something, but it seems like this works

chkr1011:
There is also a merge request from @stepkillah which adresses this issue. It works by adding some compiler flags which are "converting" the options to use either byte[] (for UWP) or X509Certificate. I have to check if this will also work for netstandard1.3.

borigas:
My understanding is #798 / #800 address an equivalent issue, but for the client side. This change is for the server side.

Can you help me understand exactly what the suggested changes are? Would you like me to put the shared changes inside compiler flags to hide them from UWP?

borigas:
I stuffed the new X509Certificate changes into non-uwp flags. Hopefully that's what you were suggesting

borigas:
Tests all pass locally and before the compiler directives passed on the build server. Doesn't make sense to me that compiler directives would unit tests. Is this something worth me investigating or is the CI flaky?

SeppPenner:
@chkr1011 Do you want to merge this to develop as well?

borigas:
@chkr1011 what's needed to get this merged? Also what's the timeline for the next release? I need to consume these changes in my project soon, and I'm hoping to avoid needing to setup a myget feed for it

borigas:
@chkr1011 can you tell me why this was closed without merging?

PR #798, opened 2019-11-17
Replace raw bytes with x509 certificate to allow specify passwords and flags

stepkillah:
Currently, there is no way to specify `Password` and `StorageFlags` for certificates used by MQTT client. This PR should extend certificates support.

Not tested with UWP

### DIFF
--- a/Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
+++ b/Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
@@ -85,7 +85,7 @@ namespace MQTTnet.Extensions.WebSocket4Net
             {
                 foreach (var certificate in _webSocketOptions.TlsOptions.Certificates)
                 {
-                    certificates.Add(new X509Certificate(certificate));
+                    certificates.Add(certificate);
                 }
             }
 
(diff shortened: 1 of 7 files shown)


chkr1011:
I had to revert this because the X502.. certificate is not available in UWP. So we have to find another solution.

stepkillah:
@chkr1011 can it be solved by adding some `#if UWP` flags? I can add those and create another pr?

SeppPenner:
Isn't this a duplicate of https://github.com/chkr1011/MQTTnet/pull/797?

stepkillah:
@SeppPenner I thought #797 is more for server implementation

SeppPenner:
You're absolutely right, my mistake. However, the solution would be the same for both (e.g. using UWP compiler flag).

stepkillah:
Ok, I'll add uwp compiler flags, and create another pr

SeppPenner:
I'm not sure if the compiler flag helps, but I would assume yes. I haven't used UWP that much by now. This is something @chkr1011 needs to answer...

chkr1011:
I also prefer the compiler flag at the moment. The reason is that there is no matching API in any way. So the data must be different in any way and there is no compatibility (apart from raw byte[] perspective). Maybe byte[] is OK but we can also consider replacing it with "Certificate" class for UWP. So that certificate becomes "Certificate" for UWP (from options to lowest level) and "X509Certificate" for all other platforms (from options to lowest level). I am not sure if X509Certificate is also supported in netstandard1.3. But this is a thing we only have to verify later. If not it will not compile anyway.

PR #800, opened 2019-11-18
Replace raw bytes with x509 certificate to allow specify passwords and flags #2

stepkillah:
Currently, there is no way to specify Password and StorageFlags for certificates used by MQTT client. This PR should extend certificates support.

Not tested with UWP

Added UWP compiler flags, so for now - for UWP it behaves in the same way as before

### DIFF
--- a/Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
+++ b/Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
@@ -85,7 +85,12 @@ namespace MQTTnet.Extensions.WebSocket4Net
             {
                 foreach (var certificate in _webSocketOptions.TlsOptions.Certificates)
                 {
+#if WINDOWS_UWP
                     certificates.Add(new X509Certificate(certificate));
+#else
+                    certificates.Add(certificate);
+#endif
+                    
                 }
             }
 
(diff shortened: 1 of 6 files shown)


stepkillah:
All tests are passed locally

Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
```diff
@@ -85,7 +85,12 @@ public async Task ConnectAsync(CancellationToken cancellationToken)
             {
                 foreach (var certificate in _webSocketOptions.TlsOptions.Certificates)
                 {
+#if WINDOWS_UWP
                     certificates.Add(new X509Certificate(certificate));
```
chkr1011: I am not sure but this looks wrong to me. In all other places you are using byte[] for UWP but here you use X509Certificate directly.

stepkillah: for UWP `certificate` variable it's byte array, but in this particular file was no platform implementation for uwp, it using `X509Certificate` for all platforms

stepkillah:
@chkr1011 any updates on this one? I don't think there should be some changes, at least from point of view that uwp should behave as before

PR #814, opened 2019-12-02
Added samples of two WinForm projects

grammyleung:
One is a simple server, the other is a simple test including server and clients of publisher and subscriber.


chkr1011:
Thank you for your contribution but I don't think we need this at all. We already have lots of demo apps. But you can request a change at the docs or wiki and create a link to your repo instead.

SeppPenner:
@chkr1011 I will add these to the wiki section soon. (If @grammyleung agress, I will host the projects from my page if necessary).

SeppPenner:
The projects are now hosted from me on my github page and can be found under https://github.com/chkr1011/MQTTnet/wiki/Examples or https://github.com/SeppPenner/MQTTnet.TestApp.SimpleServer and https://github.com/SeppPenner/MQTTnet.TestApp.WinForm.

PR #826, opened 2019-12-17
Added IMqttRetainedMessagesManager

fogzot:
* Added interface IMqttRetainedMessagesManager to allow for different
retained messages manager implementations. The interface copies almost
exactly the current MqttRetainedMessagesManager implementation.

* Added IMqttRetainedMessagesManager.Start() to configure the provided
IMqttRetainedMessagesManager outside the constructor. This method returns
Task because some implementations that use external storage may be slow on
initialization (e.g., connect to a database).

* Modified MqttRetainedMessagesManager to implement new interface.

### DIFF
--- a/Source/MQTTnet/Server/MqttServer.cs
+++ b/Source/MQTTnet/Server/MqttServer.cs
@@ -19,7 +19,7 @@ namespace MQTTnet.Server
         private readonly IMqttNetChildLogger _logger;
 
         private MqttClientSessionsManager _clientSessionsManager;
-        private MqttRetainedMessagesManager _retainedMessagesManager;
+        private IMqttRetainedMessagesManager _retainedMessagesManager;
         private CancellationTokenSource _cancellationTokenSource;
 
         public MqttServer(IEnumerable<IMqttServerAdapter> adapters, IMqttNetChildLogger logger)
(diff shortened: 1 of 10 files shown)


fogzot:
This replaces  pull request #810.

Source/MQTTnet/Server/MqttServer.cs
```diff
@@ -121,7 +121,8 @@ public async Task StartAsync(IMqttServerOptions options)
 
             _cancellationTokenSource = new CancellationTokenSource();
 
-            _retainedMessagesManager = new MqttRetainedMessagesManager(Options, _logger);
+            _retainedMessagesManager = Options.RetainedMessagesManager ?? new MqttRetainedMessagesManager();
```
chkr1011: I suggest to move that into the options class only. So that the property is initialized with the default manager. If the value is null here we can throw a proper exception.

Please consider adding a new exception like "MqttConfigurationException : MqttException" which can be used for other invalid config scenarios as well. I would like to use such an exception for other places like when MQTTv5 features are used but V3.1.1 is selected in client options.

fogzot: Done in new patch. (You probably want to check the exception message.)

fogzot: Also, MqttException does not exists. I use Exception as the other exceptions in MQTTnet do.

PR #835, opened 2019-12-23
Topic/#641/version 5.0 authentication examples

quickstar:
Hi
@PremiumBurger and me are currently writing our thesis about the topic "**Client-managed anonymous authentication and authorization for MQTT**". In order to create a working example of the concepts described in the thesis we really need the enhanced authentication as described in the official specification in [chapter 4.12](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Enhanced_authentication).

This pull request extends the current MQTTnet broker implementation with the ability to perform such an enhanced authentication.


quickstar:
Btw. looks like Appveyor isn't happy about an expired certificate...

jimsch:
To make this easier to review can you clean up the leading white space issues so that they do not show up as diffs?  I will continue to work through this over the next week.

quickstar:
> To make this easier to review can you clean up the leading white space issues so that they do not show up as diffs? I will continue to work through this over the next week.

Hi @jimsch 
Thanks for your feedback! I've changed the indentation to whitespaces.

jimsch:
@quickstar, I was going to start pulling this into my personal development tree and found out that you made your changes against 'master' and not against 'develop'.  It is my understanding from trying to follow the project that this is where they are doing all of the changes and then they merge over to master when they do version drops.  This means that your set of changes are out of date relative to where I am currently working.

jimsch:
Yet another request.  https://github.blog/2016-08-15-change-the-base-branch-of-a-pull-request/ shows how to change the target of the pull request.  Comparing to deploy will help clean out things which you did not change.

And thanks for the merge in so that I can look at the changes.  Have you done the testing or did you just merge?

Source/MQTTnet/Server/MqttClientConnection.cs
```diff
@@ -163,13 +167,30 @@ private async Task<MqttClientDisconnectType> RunInternalAsync()
                     Interlocked.Increment(ref _sentPacketsCount);
                     _lastPacketReceivedTimestamp = DateTime.UtcNow;
 
+                    if (packet is MqttAuthPacket clientAuthPacket)
+                    {
+                        if (_serverOptions.EnhancedAuthenticationBrokerHandler == null)
+                        {
+                            continue;
```
jimsch: I think this should force a disconnect rather than ignoring the message.

quickstar: Yes, I think this makes sense. It's a case which should never occur, though. How can I disconnect at this point?

quickstar:
> Have you done the testing or did you just merge?

Unittests are working, and our project seems still to work after the rebase. But it definitely needs further review and testing!

Tests/MQTTnet.Core.Tests/MQTTv5/ExtendedAuth/InteractiveEnhancedAuthBrokerHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+                {
+                    AuthenticationMethod = AuthMethod.InteractiveAuthName,
+                    AuthenticationData = Encoding.UTF8.GetBytes("nonce2")
+                }
+            };
+        }
+
+        public MqttBasePacket StartChallenge(MqttConnectPacket connectPacket)
+        {
+            if (connectPacket.Properties.AuthenticationMethod != AuthMethod.InteractiveAuthName)
```
jimsch: From a pure programming point of view, the fact that I cannot setup the CONNACK packet in this function with all of the failure information is unfortunate.  It would be nice to set errors such as BadAuthenticationMethod into the packet even if I then have the same packet coming into the Connection Validator function.  At that point I can just check if the reason code is not success and return immediately.

quickstar: > From a pure programming point of view, the fact that I cannot setup the CONNACK packet in this function with all of the failure information is unfortunate.

I probably don't understand your point. The StartChallenge method returns just a MqttBasePacket, which means you can setup whatever you like. But setting up a CONNACK packet in this method would be a little unfortunate because, the whole point of this authentication flow is to initiate an enhanced auth.

> It would be nice to set errors such as BadAuthenticationMethod into the packet even if I then have the same packet coming into the Connection Validator function.

It was an intended decision, that if you return null the default implementation kicks in. If you don't want this behavior just return your custom tailored packet, which can be a CONNACK packet as well.

Source/MQTTnet/Server/MqttClientConnection.cs
```diff
@@ -159,13 +167,33 @@ private async Task<MqttClientDisconnectType> RunInternalAsync(MqttConnectionVali
                     Interlocked.Increment(ref _sentPacketsCount);
                     _lastPacketReceivedTimestamp = DateTime.UtcNow;
 
+                    if (packet is MqttAuthPacket clientAuthPacket)
+                    {
+                        if (_serverOptions.EnhancedAuthenticationBrokerHandler == null)
+                        {
+                            Session.WillMessage = null;
+                            disconnectType = MqttClientDisconnectType.Clean;
+                            await StopAsync();
```
quickstar: What do you think about this implementation? Is this the way to go, in order to cleanly disconnect?

chkr1011: I am not sure if this is a "clean" situation. The problem is that the client is sending data which is not supported by the server. So in my mind this is a disconnect which is "not clean".

Tests/MQTTnet.Core.Tests/MQTTv5/Client_Tests.cs
```diff
@@ -12,12 +15,51 @@
...
 using MQTTnet.Server;
 using MQTTnet.Tests.Mockups;
+using MQTTnet.Tests.MQTTv5.ExtendedAuth;
 
 namespace MQTTnet.Tests.MQTTv5
 {
     [TestClass]
     public class Client_Tests
     {
+        [TestMethod]
```
chkr1011: Please have a look at the "TestEnvironment" class and check if you can make use of that for testing. It already wraps ports and also does a lot of validations and error checking.

Source/MQTTnet/Server/MqttClientConnection.cs
quickstar: >So in my mind this is a disconnect which is "not clean".

Makes absolutely sense, I'm going to change this to NotClean!

Source/MQTTnet/Server/MqttRetainedMessagesManager.cs
```diff
@@ -21,7 +21,11 @@ public Task Start(IMqttServerOptions options, IMqttNetChildLogger logger)
             if (logger == null) throw new ArgumentNullException(nameof(logger));
             _logger = logger.CreateChildLogger(nameof(MqttRetainedMessagesManager));
             _options = options ?? throw new ArgumentNullException(nameof(options));
+#if NET452
+            return new Task(() => {});
```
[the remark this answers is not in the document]
quickstar: I had to change this, because the develop branch didn't compile on my machine.

Task.FromResult(0) also just creates a new Task with the result value of 0. Therefore i think it doesn't really makes any difference. But in order to stay consistent I'm going to change it!

Tests/MQTTnet.Core.Tests/MQTTv5/ExtendedAuth/InteractiveEnhancedAuthBrokerHandler.cs
jimsch: I wrote my comment to soon.  I was doing it based on just reading the code rather than having done the integration to my code base and then debugging my process.  Sorry about that.

I am going to change this to a completely different request.  There is now a new item which was defined call SessionItems passing this object into the two functions allows for a clean way to pass data between the call to the connect validator and these two functions.  It also allows for internal state to be kept to deal with multiple round trips such as a Kerberos authentication.

Source/MQTTnet/Server/MqttClientConnection.cs
```diff
@@ -159,13 +167,33 @@ private async Task<MqttClientDisconnectType> RunInternalAsync(MqttConnectionVali
...
+                            disconnectType = MqttClientDisconnectType.NotClean;
+                            await StopAsync();
+                            break;
+                        }
+
+                        var responsePacket = _serverOptions.EnhancedAuthenticationBrokerHandler.HandleAuth(clientAuthPacket, Session.Items);
+                        if (authPacket is MqttConnAckPacket connAckPacket)
+                        {
+                            connAckPacket.IsSessionPresent = !Session.IsCleanSession;
+                        }
```
jimsch: I don't know if I am supposed to deal with this or if it should be here.  If I return a connAckPacket with a failed reason code then the session needs to be shutdown.

jimsch: Similar question arises for a re-authentication where one returns a DISCONNECT packet in v5 and just disconnects the session in v3.

Source/MQTTnet/Server/MqttRetainedMessagesManager.cs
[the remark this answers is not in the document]
gfoidl: In general I would use [Task.CompletedTask](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.completedtask), and for .NET versions where it is not available (i.e. before 4.6) just use a cached task like `private readonly static Task s_completed = Task.FromResult(0);` so it safes the allocation of a fresh task-object each time.

I can submit a PR to change all uses of `Task.FromResult(0)` to make it allocation free if you want.

Issue #687, opened 2019-07-02
ManagedClient WithCleanSession


det-iot:
I'm trying to use the clean session with the managed client

```
var options = new ManagedMqttClientOptionsBuilder()
    .WithAutoReconnectDelay(TimeSpan.FromSeconds(5))
    .WithClientOptions(new MqttClientOptionsBuilder()
        .WithClientId(Guid.NewGuid().ToString())
        .WithTcpServer(configuracion.Servidor)
        .WithCredentials(configuracion.Username, configuracion.Password)
        .WithKeepAlivePeriod(TimeSpan.FromSeconds(60))
        .WithCleanSession(false) //sesion persistente
        .Build())
    .Build();
var mqttClient = new MqttFactory().CreateManagedMqttClient();

await mqttClient.SubscribeAsync(new TopicFilterBuilder()
    .WithTopic(configuracion.Topic)
    .WithExactlyOnceQoS()
    .Build());

await mqttClient.StartAsync(options);
```
 I have tried to connect to 2 different HiveMQ brokers with the following error:

```
Error: System.NotSupportedException: The managed client does not support existing sessions.
   at MQTTnet.Extensions.ManagedClient.ManagedMqttClient.StartAsync(IManagedMqttClientOptions options)
   at MQTT_client.MQTT.ManageMQTTClient()
```

If I use other client, I can connect successfully using cleansesion to that HiveMQ broker.

So, is it not supported to use cleansession with managed client? or it could be done in a different way using the managed client?

I'm using the 3.0.2 MQTTnet version

SeppPenner:
As far as I can see, you're right: It doesn't yet support this, does it? @chkr1011

chkr1011:
No that is not supported yet but already on the feature list.

det-iot:
Thank you for the information. I will wait for the feature. I think it would be very useful.

det-iot:
@chkr1011 do you have an ETA of supporting this feature? I will need to choose in a short time a MQTT client which support clean session.

chkr1011:
No not yet. Also will this feature come with some drawbacks because it cannot cover all use cases. But you can always use the regular _MqttClient_. The _ManagedMqttClient_ is just a small extension which reconnects automatically etc. With that base client you can do everything (including reuse of existing sessions).

det-iot:
Hello,

[...]

Issue #492, opened 2018-12-11
Managed client clean session support


rcollada:
Hi, I just would like to understand the rational behind the fact that the Managed client does not support Clean Session. Even if I implement my own managed client, because I need to recover the session I wonder if I'm missing something there.
Regards.

chkr1011:
Hi,
the problem is that it requires some more complex subscription handling. 

Lets say you just pass _true_ via connecting. Then the managed client will still subscribe to all topics automatically (like at the moment). But then he will get all retained messages for that topic. If the session exists the subscriptions also exist at the server. So there is no need from the client to do this. But if the client ignores this fact he subscribes and then receives all retained messages.

This is not a problem in general but some users may don't want this. Other may accept it. So it should be configurable in my opinion or supported out of the box. So this means the subscriptions must be handled including a flag saying "this subscription is approved by the server". It has to be set only after a successful _SubAck_. Then after a reconnect the client must inspect the _ConnectAck_ if the session is really existing. If the session is still new all subscriptions must be invalidated and subscribed again.

But if this flow works well it can happen that a client ID is used twice (by accident). If the second client creates a new session with other subscriptions the original client may find another session after taking over the connection and thinks that the session exists (and thus is his one). So the client will keep a broken set of subscriptions. 

I hope this gives you a quick overview about the reasons which causes the not supported existing sesions (yet).

Best regards
Christian

kmaesdillen:
Do have any clue when this will be supported?

chkr1011:
At the moment this is on low priority. I am mostly busy with MQTT v5 support.
After this is done I will prepare some other tasks. But it is still possible to use the core Client and build a managed client with the exact required features.

RainerMueller82:
Hi, this is exactly what I need. Nevertheless, for me it's ok to just set cleansession to false, I don't use retain messages on the same topic. So subscribing again will not result in unwanted behaviour I think. I'll try that.

SeppPenner:
Is this still a valid feature request?

rcollada:
I downloaded the managed client and modified a little bit to support clean session false, and it have been working perfectly in a small prod server.
It would be great if the feature could be included in the nugget package but in the mean time if you need a workaround that's what you can do.

chkr1011:
I also have this topic on my list but had no time to start it. So please be patient and wait some more time :smile:

t03apt:
@chkr1011 
[...]

PR #836, opened 2019-12-23
Support existing session with ManagedMqttClient.

PMExtra:
#492 #687

### DIFF
--- a/Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
+++ b/Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
@@ -91,10 +91,6 @@ namespace MQTTnet.Extensions.ManagedClient
             if (options == null) throw new ArgumentNullException(nameof(options));
             if (options.ClientOptions == null) throw new ArgumentException("The client options are not set.", nameof(options));
 
-            if (!options.ClientOptions.CleanSession)
-            {
-                throw new NotSupportedException("The managed client does not support existing sessions.");
-            }
 
             if (!_maintainConnectionTask?.IsCompleted ?? false) throw new InvalidOperationException("The managed client is already started.");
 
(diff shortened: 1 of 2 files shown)


PMExtra:
By the way. I'm really think it would be better if we use `switch` statement instead of multiple `if-return` in `TryMaintainConnectionAsync`. But I won't push the change in this merge request because I want the minimal change for the feature and for simply code review.

If you want, you can replace it with `ManagedMqttClient.cs:318`:

```csharp
                switch (connectionState)
                {
                    case ReconnectionResult.NotConnected:
                        StopPublishing();
                        await Task.Delay(Options.AutoReconnectDelay, cancellationToken).ConfigureAwait(false);
                        break;

                    case ReconnectionResult.Reconnected:
                        await PublishReconnectSubscriptionsAsync().ConfigureAwait(false);
                        StartPublishing();
                        break;

                    case ReconnectionResult.Recovered:
                        StartPublishing();
                        break;

                    case ReconnectionResult.StillConnected:
                        await PublishSubscriptionsAsync(Options.ConnectionCheckInterval, cancellationToken).ConfigureAwait(false);
                        break;
                }
```

chkr1011:
The reason (for using if) is that I just prefer them over switch case because if is not limited to static values I need less code (no case: no break;). I know they are a little bit faster than if statements but I am sure that this is not critical for this project.

PMExtra:
@chkr1011 In this case, I think `switch` is more naturally and easy to read. The statements are just depends on a state of the enum type that is a known finite set.
On the other hand, I recommend you to read the new `case .. when ...` pattern matching feature in C# 7.0, and the `switch expression` feature in C# 8.0. They surprise me so I share with you if you haven't know.

Anyway, it's just a little suggestion, I don't mind if you have your own code style. Thank you for creating MQTTnet, it helped me a lot.

PR #838, opened 2019-12-31
Allow access to full packet on failed connect

jimsch:
The server can return user properties that might be useful for the client when a failed connect occurs.  Change the code so that the full connect ack packet is available on a failed connect.

Also, when building a connect ack, the wrong set of user properties is being returned.  Return the response set not the request set of properties.


Source/MQTTnet/Adapter/MqttConnectingFailedException.cs
```diff
@@ -5,12 +5,13 @@ namespace MQTTnet.Adapter
 {
     public class MqttConnectingFailedException : MqttCommunicationException
     {
-        public MqttConnectingFailedException(MqttClientConnectResultCode resultCode)
+        public MqttConnectingFailedException(MqttClientAuthenticateResult resultCode)
             : base($"Connecting with MQTT server failed ({resultCode.ToString()}).")
```
chkr1011: Is the message still working when using the new object or does it only print the class name "MqttClientAuthenticateResult"?

jimsch: No, it was supposed to have the correct field referenced.  It apparently got lost when I setup the pull request from my development branch.  Now correctly references the ResultCode field

Source/MQTTnet/Adapter/MqttConnectingFailedException.cs
```diff
@@ -5,12 +5,13 @@ namespace MQTTnet.Adapter
 {
     public class MqttConnectingFailedException : MqttCommunicationException
     {
-        public MqttConnectingFailedException(MqttClientConnectResultCode resultCode)
-            : base($"Connecting with MQTT server failed ({resultCode.ToString()}).")
+        public MqttConnectingFailedException(MqttClientAuthenticateResult resultCode)
```
chkr1011: You already added a new "Result" property. So please name the parameter also "result" because it no longer the result code only.

jimsch: done

PR #839, opened 2020-01-04
bugfixes

JanEggers:
see commit messages


Source/MQTTnet/Internal/Disposable.cs
```diff
@@ -44,12 +44,13 @@ public void Dispose()
             {
                 return;
             }
+
+            _isDisposed = true;
```
JanEggers: without this change the test ManagedClients_Will_Message_Send failed 

reason: the managed client sends a disconnect while it is disposed.

im not sure about this change. i

chkr1011:
Next time please also fill the change log for the nuget. Otherwise I might forget something.

PR #840, opened 2020-01-05
Setup for the Subscription intercepter to return a reason code

jimsch:
Allow for the intercept code to set a reason code rather than using a default which may be wrong.   The spec says that both the success and failure codes can be changed.

I am not totally happy about the way things are setup as something that sets Accept session to false and then true will get the wrong reason code.  Should it be changed to have a success and failure version of the result code and return the last one set?  For most uses what is there is going to be fine I think.


Source/MQTTnet/Server/MqttSubscriptionInterceptorContext.cs
```diff
@@ -20,8 +22,37 @@ public MqttSubscriptionInterceptorContext(string clientId, TopicFilter topicFilt
         /// </summary>
         public IDictionary<object, object> SessionItems { get; }
 
-        public bool AcceptSubscription { get; set; } = true;
+        private bool acceptSubscription = true;
+        public bool AcceptSubscription
+        {
+            get => acceptSubscription;
+            set {
+                if (!value && resultCode < MqttSubscribeReasonCode.UnspecifiedError) resultCode = MqttSubscribeReasonCode.UnspecifiedError;
```
chkr1011: Please stick to the code style used in this project. There is nothing bad in your implementation but we just use curly braces in every case. So please adopt the style here as well. And below also 😄

Source/MQTTnet/Server/MqttSubscriptionInterceptorContext.cs
```diff
@@ -20,8 +22,37 @@ public MqttSubscriptionInterceptorContext(string clientId, TopicFilter topicFilt
         /// </summary>
         public IDictionary<object, object> SessionItems { get; }
 
-        public bool AcceptSubscription { get; set; } = true;
+        private bool acceptSubscription = true;
```
chkr1011: Please move the fields to the top of the file and use _ as prefix. This is the default code style of this project.

Source/MQTTnet/Server/MqttSubscriptionInterceptorContext.cs
```diff
@@ -20,8 +22,37 @@ public MqttSubscriptionInterceptorContext(string clientId, TopicFilter topicFilt
...
+            }
+        }
 
         public bool CloseConnection { get; set; }
+
+        public static MqttSubscribeReasonCode ConvertToSubscribeReasonCode(MqttQualityOfServiceLevel qualityOfServiceLevel)
+        {
+            switch (qualityOfServiceLevel)
+            {
+            case MqttQualityOfServiceLevel.AtMostOnce: return MqttSubscribeReasonCode.GrantedQoS0;
```
chkr1011: Also please check the style here.

jimsch: I am not clear what the style of the project is in this case, can you please let me know.   This piece of code was a straight cut and paste the only difference being the indentation level (which I have changed).  In other places I have seen it so that the return is always on a different line and braces are added for each case.

chkr1011:
Please update the nuget release notes according to your changes. Then I will merge your request.

Source/MQTTnet/Server/MqttSubscriptionInterceptorContext.cs
```diff
@@ -20,8 +25,41 @@ public MqttSubscriptionInterceptorContext(string clientId, TopicFilter topicFilt
         /// </summary>
         public IDictionary<object, object> SessionItems { get; }
 
-        public bool AcceptSubscription { get; set; } = true;
+        public bool AcceptSubscription
+        {
+            get => _acceptSubscription;
+            set {
+                if (!value && _resultCode < MqttSubscribeReasonCode.UnspecifiedError)
```
chkr1011: Is this really required? I want to avoid implicit code in setters of properties.

jimsch: There i a problem that a new variable has been introduced and you now need to keep the result code error and the accept subscription state synchronized.  Keeping the backwards compatibility on AcceptSubscription means that old code which just sets that value needs to modify the returned error code as well.  I agree that is is not ideal, but I am not sure how else to address it.

chkr1011:
Please merge the latest master branch into your feature branch.

Source/MQTTnet/Server/MqttSubscriptionInterceptorContext.cs
chkr1011: Ok got your point. You mean that "AcceptSubscription" cannot be true and Reason code something like "denied" at the same time right? I am not 100 % sure right now but I assume we have to declare the "AcceptSubscription" as obsolete and only use the result code. Then no special setter code is required. But for backward compatibility we have to keep the accept subscription for a small amount of versions. Maybe we can check the "AcceptSubscription" for true only and return a proper "best matching" result code. Or we can set the "AcceptSubscription" to a nullable boolean!? What do you think?

jimsch: I had been thinking about moving to a single variable as well.  I did keep the setter on AcceptSubscription.  I think that AcceptSubscription should get the getter and lose the setter in the future.  I tried to just obsolete that one item and was told that it needed to move to a language version of 8.2 in order for the obsolete attribute to be placed on just the setter.  I was reluctant to obsolete the entire attribute as keeping the error checking code in one location seems to be a better answer as far as I was concerned.

SeppPenner:
@jimsch Can you fix the merge conflicts, so @chkr1011 can merge this, please?

jimsch:
I have merged forward to current.  I don't know what is causing any of the errors but the subscribe tests pass on my machine just fine.   The intercept test fails, but I don't think that is my code.

SeppPenner:
@JanEggers @chkr1011 Can you check if this is ok? I also have no idea why the tests fail...

chkr1011:
I will close this due to inactivity but I will port the changes to the latest version.

Issue #851, opened 2020-01-27
check "Keep Alive" while other messages are sent


cdobler:
### Describe the bug
If no messages are sent during a defined keep alive interval, the client sends a PINGREQ package to the server and whether a PINGRESP package is returned. If no response is received from the server, the connection is closed.

If regular messages are sent from the client, no pings are sent since each time a message is sent, an internal timer is reset (*_sendTracker*).

In my setup, there is a gateway between the client and the server. If the connection between the gateway and the server breaks down, the client does not recognize this. Moreover, I send a message every second, i.e. no pings are sent. As a result, the client needs to wait for a TCP timeout to realize that the connection is down.

In a test, I stopped sending regular messages. As a result, the client started sending pings and recognized a broken connection immediately due to the failing response from the server.

I have two ideas in mind to solve this problem:
- Introduce a second timer (*_receiveTracker*) and send a ping in case one of the two timers expires.
- Add a configuration option to send pings in defined intervals without using the *_sendTracker*

If one of those ideas is ok, I could implement the required changes and start a merge request.

### Which project is your bug related to?
- Client (MqttClient)

### To Reproduce
1. Using this version of MQTTnet '3.0.8'. (I used 2.8.5 before which was affected as well)
2. See description above

### Expected behavior
The client should realize that the connection to the server broke down without having to wait for a TCP timeout.

PR #852, opened 2020-01-28
Added receiveTracker to detect whether messages are received from broker.

cdobler:
Implementation of first proposed solution: "Introduce a second timer (_receiveTracker) and send a ping in case one of the two timers expires."
#851


chkr1011:
Please fix conflict. Then I will merge.

SeppPenner:
No problems anymore, I will merge this now :)

Issue #812, opened 2019-11-28
Concurrent processing


ramonsmits:
### Describe the feature request
Issue https://github.com/chkr1011/MQTTnet/issues/732 talk parallel processing which is also my request.

This is especially important under high load if the processor is also doing IO. Assume for example that due to IO processing duration is 200ms than only 5 messages per second is possible. 

Allow concurrent processing if the consumer requests this but have the default set to 1 which would result in sequential processing.

### Which project is your feature request related to?

- Client

### Describe the solution you'd like

Allow concurrent processing. The amount of allows concurrent processing should be limited by using for example `SemaphoreSlim`.

### Describe alternatives you've considered

There is a bad workaround possible by doing something like:
```c#
static SemaphoreSlim concurrencyLimit = new SemaphoreSlim(Environment.ProcessorCount * 4);

static async Task Handler(MqttApplicationMessageReceivedEventArgs ea)
{
    await concurrencyLimit .WaitAsync().ConfigureAwait(false);
    // Do NOT await, so task is forked but results in QoS 1 issues it task fails.
    Task.Run(async () =>
    {
        try
        {
            // Do your stuff here
        }
        catch
        {
        }
        finally
        {
            concurrencyLimit .Release();
        }
    });
}
```

But this solution has issues with QoS 1 as now the event handler ACK's already before actually completed.

chkr1011:
I am already working on a new client which is mostly just a wrapper on top of a socket. It does not implement and use cases etc. So you can deal with all messages on your own and how you need it for your use cases. I don't think I can implement this in the client because some need the order (so no tasks) and some need it as fast as possible ignoring the order. I hope this helps. I hope to release that new client soon.

ramonsmits:
@chkr1011 I understand that for some order is important so having the default be a concurrency limit of 1 to get sequential thus ordered processing would probably be a good idea.

> because some need the order (so no tasks)

For my understanding, what do you mean with the comment "so no tasks"?

PMExtra:
Please reference my forked commit: https://github.com/PMExtra/MQTTnet/commit/d7704b1a7f8dfd422c17eeca86c53306953ba3a5

I've made a concurrent solution, but i have no idea about UnitTest. I've tested it in my project, it seems to work fine.

ramonsmits:
@PMExtra Not entirely sure if your solution is thread-safe and also, it doesn't have any maximum concurrency control meaning you could potentially have thousands of concurrent tasks.

[...]

PR #875, opened 2020-03-18
Suppport parallel receiving (concurrent processing).

PMExtra:
Reference #812 

This is just a preliminary solution, without UnitTest for now.  I have no idea about how to test it.

I want more people to discuss about it, and we can keep improve this feature until merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ MQTTnet is a high performance .NET library for MQTT based communication. It prov
 * Uniform API across all supported versions of the MQTT protocol
 * Interfaces included for mocking and testing
 * Access to internal trace messages
-* Unit tested (~210 tests)
+* Unit tested (~250 tests)
 * No external dependencies
 
 \* Tested on local machine (Intel i7 8700K) with MQTTnet client and server running in the same process using the TCP channel. The app for verification is part of this repository and stored in _/Tests/MQTTnet.TestApp.NetCore_.


Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -522,7 +530,18 @@ private async Task TryProcessReceivedPacketAsync(MqttBasePacket packet, Cancella
 
                 if (packet is MqttPublishPacket publishPacket)
                 {
-                    await TryProcessReceivedPublishPacketAsync(publishPacket, cancellationToken).ConfigureAwait(false);
+                    if (Options.ParallelReceiving)
+                    {
+                        var task = new Task(async () => await TryProcessReceivedPublishPacketAsync(publishPacket, cancellationToken), cancellationToken);
```
chkr1011: Please consider using the ThreadPool.Enqueue.. method because we might get a lot of messages.

PMExtra: Do you mean `ThreadPool.QueueUserWorkItem(WaitCallback callBack)`?

I don't know what's the difference between them. But I think the default implementation of TPL may be sufficient.

Another way, `ThreadPool` is not supported in `netstandard1.3`. Shall we consider to upgrade netstandard version?
According to https://docs.microsoft.com/en-us/dotnet/standard/net-standard , I'm not sure we should support the outdated .NET Standard version.

chkr1011: Mhh we still have to support netstandard1.3. But I have another question. Why do you need that feature? You can start a new task also in your message handler? What is the benefit in moving that into this library directly?

PMExtra: Because of the QOS Ack/Rec

PMExtra: @chkr1011 Do you have any idea?

chkr1011: I am thinking about a different solution. My idea is to add a new method/property to the event args of the handler. Then you can "disable" the internal auto ACK and do the ACK at a different time. Thats the way how other libs like in Azure do this.  So a new property would indicate the AcknowledgeMode = Manual|Auto and a new "Acknowledge" method will allow the ACK at a different time. If will do nothing for QoS 0 but the proper stuff for QoS 1 and 2.

PMExtra: Awesome.
I will also use the parallel receiving in my self project that is a RPC library of MQTT protocol based on MQTTnet.
And in my project, I'm planning 3 options of AcknowledgeMode: OnCall / OnResult / Manual.
It could based on manual acknowledge mode in MQTTnet.

PMExtra: It's different with MQTTnet.Extensions.Rpc.
It's based on mqttv5 and broker-side ACL ready, support broadcast and cluster.
It's open-source but still developing.
I'm glad to contribute it into MQTTnet.Extensions.Rpc if you approve it.

PR #880, opened 2020-03-25
Reduce the performance cost of Memory<T>.getSpan()

xljiulang:


chkr1011:
@JanEggers Do you think that makes sense? I am no expert in memory/span stuff.

JanEggers:
@chkr1011 the change is fine, do you know why appveyor fails to build?

chkr1011:
@JanEggers Yes some tests are still failing. I had no time to check them. They also fail on local machine so it is possible to fix them. But most of them seem to be race conditions to me. Probably something wrong is merged. Do you have some time to check them as well?

PR #884, opened 2020-03-29
Added check if the server is started or not.

SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/846.


SeppPenner:
I'm not sure if this is a good idea, it was just a proposal and easy to implement...

Source/MQTTnet/Server/MqttServer.cs
```diff
@@ -190,6 +190,8 @@ public Task ClearRetainedApplicationMessagesAsync()
             return _retainedMessagesManager?.ClearMessagesAsync();
         }
 
+        public bool IsStarted => _cancellationTokenSource != null;
```
chkr1011: Please also move this to the server interface because you have to cast the object from the MQTT factory.

chkr1011:
Please add an entry to the nuspec file which describes this change.

SeppPenner:
I changed the pull request.

Source/MQTTnet/Server/MqttServer.cs
chkr1011: Also please move the property to the property section of the file. Be best is at corresponding other properties for a similar purpose.

And please merge master and fix conflicts.

PR #885, opened 2020-03-29
Removed checks for user properties in MQTT v3 and instead ignore them.

SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/848. For me, the argumentation under the issue seems valid.


chkr1011:
Please add an entry at the nuspec file so that this change is documented.

SeppPenner:
@chkr1011 I have updated the .nuspec file.

PR #886, opened 2020-03-30
GetUserProperty extension.

PMExtra:
See files changed, there are some use cases in UnitTest.

### DIFF
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -22,6 +22,7 @@
 * [Server] Added interceptor for unsubscriptions.
 * [MQTTnet.Server] Added interceptor for unsubscriptions.
 * [MQTTnet.AspNetCore] improved compatibility with AspNetCore 3.1
+* [MqttApplicationMessage] Added GetUserProperty<T> convenience method (thanks to @PMExtra).
     </releaseNotes>
     <copyright>Copyright Christian Kratky 2016-2019</copyright>
     <tags>MQTT Message Queue Telemetry Transport MQTTClient MQTTServer Server MQTTBroker Broker NETStandard IoT InternetOfThings Messaging Hardware Arduino Sensor Actuator M2M ESP Smart Home Cities Automation Xamarin</tags> 
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -22,7 +22,7 @@
 * [Server] Added interceptor for unsubscriptions.
 * [MQTTnet.Server] Added interceptor for unsubscriptions.
 * [MQTTnet.AspNetCore] improved compatibility with AspNetCore 3.1
-* [MqttApplicationMessage] Added GetUserProperty<T> convenience method (thanks to @PMExtra).
+* [Core] Added MqttApplicationMessage.GetUserProperty&lt;T&gt;() convenience method (thanks to @PMExtra).
     </releaseNotes>
     <copyright>Copyright Christian Kratky 2016-2019</copyright>
     <tags>MQTT Message Queue Telemetry Transport MQTTClient MQTTServer Server MQTTBroker Broker NETStandard IoT InternetOfThings Messaging Hardware Arduino Sensor Actuator M2M ESP Smart Home Cities Automation Xamarin</tags> 
--- a/Source/MQTTnet/Extensions/UserPropertyExtension.cs
+++ b/Source/MQTTnet/Extensions/UserPropertyExtension.cs
@@ -1,0 +1,18 @@
+using System;
+using System.Linq;
+
+namespace MQTTnet.Extensions
+{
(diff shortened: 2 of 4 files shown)


chkr1011:
Thank you for your contribution! Please also update the nuspec and add an entry with your changes.

PMExtra:
Is this right?

Source/MQTTnet/Extensions/UserPropertyExtension.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Linq;
+
+namespace MQTTnet.Extensions
+{
+    public static class UserPropertyExtension
+    {
+        public static string GetUserProperty(this MqttApplicationMessage message, string propertyName, StringComparison comparisonType = StringComparison.OrdinalIgnoreCase)
+        {
+            return message.UserProperties.SingleOrDefault(up => up.Name.Equals(propertyName, comparisonType))?.Value;
```
chkr1011: Please add parameter null validation. Message can be null. In case the user properties are null I suggest to return null as well.

Build/MQTTnet.nuspec
```diff
@@ -22,6 +22,7 @@
 * [Server] Added interceptor for unsubscriptions.
 * [MQTTnet.Server] Added interceptor for unsubscriptions.
 * [MQTTnet.AspNetCore] improved compatibility with AspNetCore 3.1
+* [MqttApplicationMessage] Added GetUserProperty<T> convenience method (thanks to @PMExtra).
```
chkr1011: Please choose either Core, Client, Server, or a project name like MQTTnet.Server. In your case I would say "Core" fits best.

Source/MQTTnet/Extensions/UserPropertyExtension.cs
PMExtra: Do you means that?
```
message?.UserProperties?.SingleOrDefault...
```

PMExtra:
Please check it.

Source/MQTTnet/Extensions/UserPropertyExtension.cs
chkr1011: Yes and no 😄 Please throw argument exception if message is null (in both new methods) because this is never allowed. But the user properties may be null when MQTT 3.1.1 is used etc. So here we return null.

PMExtra:
I have a change that make more friendly exception. What do you think?
```
public static T GetUserProperty<T>(this MqttApplicationMessage message, string propertyName, StringComparison comparisonType = StringComparison.OrdinalIgnoreCase)
{
    var value = GetUserProperty(message, propertyName, comparisonType);
    try
    {
        return (T) Convert.ChangeType(value, typeof(T));
    }
    catch (Exception ex)
    {
        throw new InvalidOperationException($"Cannot convert value '{value}' of user property '{propertyName}' to type '{typeof(T).Name}'", ex);
    }
}
```

chkr1011:
For conversion that is OK but also please add the null check for application message.

PMExtra:
What should return if message is null? default(T)?

In my opinion, I suggest throw an exception if message is null.

And if the property is non-exists or null, `GetUserProperty<int>` should throw an exception, `GetUserProperty<int?>` should return `null`.

chkr1011:
Please read my comment above 😄 

Yes and no 😄 Please throw argument exception if message is null (in both new methods) because this is never allowed. But the user properties may be null when MQTT 3.1.1 is used etc. So here we return null.

PMExtra:
So, what behaviors are expected if GetUserProperty is null:

1. `GetUserProperty<int>`
2. `GetUserProperty<int?>`

In my opinion, I suggest 1 will throw an exception, and 2 will return null.

And I suggest `throw new InvalidOperationException` if `UserProperties` is null (MQTT v3)

chkr1011:
In both cases it should throw something like a new "UserPropertyNotFoundException : MqttNetException".

PMExtra:
If so, we should make another `GetUserPropertyOrDefault` methods.

PMExtra:
So I don't think we should make anything like "UserPropertyNotFoundException".

And I just found we can't ensure an ApplicationMessage is v3 or v5, because it can be `new` or `Build()` instead of received. Although we usually call `GetUserProperty` with received message.

So I suggest follow your previous opinion, return null if `message.UserProperties` is null.

chkr1011:
Yes that makes sense.

PMExtra:
I think it's ready to merge now, please check.

SeppPenner:
I have resolved the issues with the .nuspec file @chkr1011.

chkr1011:
I merged and tested it but unfortunately it does not work on UWP apps. I removed the type conversion for now. Please do some research how to implement this for UWP. You can again add it with support for all supported platforms.

PR #889, opened 2020-03-31
Support MqttClientOptionsBuilder.WithConnectionUri()

PMExtra:
A convenience way to configure server, port, username and password, example:
```
var options = new MqttClientOptionsBuilder()
    .WithConnectionUri("mqtt://user:password@127.0.0.1")
    .Build();
```

`mqtt://` and `mqtts://` are defined by MQTT official. (https://mqtt.org/wiki/URI-Scheme)
`ws://` and `wss://` are defined by [RFC8307](https://tools.ietf.org/html/rfc8307)

`tcp://` is commonly used.

The above 5 schemes are supported now.

I also have some other non-standardized idea like:
- [ ] `ssl://`
- [ ] `tls://`
- [ ] `mqtt+ws://`
- [ ] `mqtt+wss://`
- [ ] `mqtt+tcp://`
- [ ] `mqtt+ssl://`
- [ ] `mqtt+tls://`

Some of them could be added if you approve it.

### DIFF
--- a/Build/MQTTnet.nuspec
+++ b/Build/MQTTnet.nuspec
@@ -22,6 +22,7 @@
 * [Server] Added interceptor for unsubscriptions.
 * [MQTTnet.Server] Added interceptor for unsubscriptions.
 * [MQTTnet.AspNetCore] improved compatibility with AspNetCore 3.1
+* [Client] Support WithConnectionUri to configure client (thanks to @PMExtra).
     </releaseNotes>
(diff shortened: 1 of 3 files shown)


chkr1011:
Good idea. In my opinion the already added ones are enough.

SeppPenner:
In my opinion, it's ok to support the standard only as well.

chkr1011:
@PMExtra Please leave a one liner in the nuspec and describe the new feature, Please choose one of the usual areas in braces.

PR #891, opened 2020-04-03
Introduce asynchronous message processing

### LINKED ISSUE #587: MQTTnet.Exceptions.MqttCommunicationTimedOutException (3.0.0-alpha3) with MqttQualityOfServiceLevel.AtLeastOnce
While testing 3.0.0-alpha4 I get the following exception:
```
MQTTnet.Exceptions.MqttCommunicationTimedOutException: Exception of type 'MQTTnet.Exceptions.MqttCommunicationTimedOutException' was thrown.
   at MQTTnet.PacketDispatcher.MqttPacketAwaiter`1.WaitOneAsync(TimeSpan timeout)
   at MQTTnet.Client.MqttClient.SendAndReceiveAsync[TResponsePacket](MqttBasePacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.PublishAtLeastOnceAsync(MqttPublishPacket publishPacket, CancellationToken cancellationToken)
```

Getting this after I transitioned to the async methods although I'm not certain that v3 is the cause.

Easily reproducable by just PublishAsync a bunch of messages with `MqttQualityOfServiceLevel.AtLeastOnce` set. There are blocking and eventually result in the timeouts.

This is with RabbitMQ as the broker.

### LINKED ISSUE #648: Publish from MessageReceivedHandler doesn't work
When I Publish message from handler

`client.UseApplicationMessageReceivedHandler(async e =>
 {
      await client.PublishAsync(message);
 });`

 I am getting MqttCommunicationTimedOutException from MqttPacketAwaiter before receiving acknowledgement that message was Published.

Despite I have set very long communication timeout - it is throwed much earlier.
I am using QOS 1.

Workaround that we've currently decided to use - to publish asynchronously:

`client.UseApplicationMessageReceivedHandler(e =>
 {
      Task.Run(async () => await client.PublishAsync(message));
 });`

### LINKED ISSUE #829: When client is receiving "long processing" topics, it will be disconnect at some point.
In my application I want to receive images, do some modifications and save it. To simulate a "worst case" scenarios I put a Thread.Sleep(500) delay in the MqttClient.ApplicationReceiveHandler.

When publishing that topic about 100 times in a for loop, the subscribing client droppes the connection at some point. The server log shows "client xxx has exceeded timeout, disconnecting."

I tried to set the .WithKeepAlivePeriode up to 5 minutes. And then it worked!

**It seems, when processing a long receiving queue, the KeepAlive signal could not be processed as long as the topic queue is not dequeued (processed).**

PSanetra:
This PR fixes #648, #587 and #829 by introducing asynchronous message processing. 

The current code calls the `ApplicationMessageReceivedHandler` synchronously in the packet receiver loop. This blocks KeepAlive and Ack packets from being processed while a message is still being processed.

This PR also fixes the test `MQTTnet.Tests/Server_Tests/Same_Client_Id_Connect_Disconnect_Event_Order` as it failed on my machine due to a race condition in the MqttClientSessionsManager.

Fixes #648
Fixes #587
Fixes #829

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -34,6 +34,9 @@ namespace MQTTnet.Client
         private CancellationTokenSource _backgroundCancellationTokenSource;
         private Task _packetReceiverTask;
         private Task _keepAlivePacketsSenderTask;
+        private Task _publishPacketReceiverTask;
+
+        private AsyncQueue<MqttPublishPacket> _publishPacketReceiverQueue;
 
         private IMqttChannelAdapter _adapter;
         private bool _cleanDisconnectInitiated;
@@ -87,6 +90,9 @@ namespace MQTTnet.Client
                     _logger.Verbose($"Trying to connect with server '{options.ChannelOptions}' (Timeout={options.CommunicationTimeout}).");
                     await _adapter.ConnectAsync(options.CommunicationTimeout, combined.Token).ConfigureAwait(false);
                     _logger.Verbose("Connection with server established.");
+
+                    _publishPacketReceiverQueue = new AsyncQueue<MqttPublishPacket>();
+                    _publishPacketReceiverTask = Task.Run(() => ProcessReceivedPublishPackets(backgroundCancellationToken), backgroundCancellationToken);
 
                     _packetReceiverTask = Task.Run(() => TryReceivePacketsAsync(backgroundCancellationToken), backgroundCancellationToken);
 
@@ -230,6 +236,9 @@ namespace MQTTnet.Client
             _backgroundCancellationTokenSource?.Dispose();
             _backgroundCancellationTokenSource = null;
 
+            _publishPacketReceiverQueue?.Dispose();
+            _publishPacketReceiverQueue = null;
+
             _adapter?.Dispose();
             _adapter = null;
         }
@@ -300,9 +309,12 @@ namespace MQTTnet.Client
             try
             {
                 var receiverTask = WaitForTaskAsync(_packetReceiverTask, sender);
+                var publishPacketReceiverTask = WaitForTaskAsync(_publishPacketReceiverTask, sender);
                 var keepAliveTask = WaitForTaskAsync(_keepAlivePacketsSenderTask, sender);
 
-                await Task.WhenAll(receiverTask, keepAliveTask).ConfigureAwait(false);
+                await Task.WhenAll(receiverTask, publishPacketReceiverTask, keepAliveTask).ConfigureAwait(false);
+
+                _publishPacketReceiverQueue.Dispose();
             }
             catch (Exception e)
             {
@@ -522,7 +534,7 @@ namespace MQTTnet.Client
 
                 if (packet is MqttPublishPacket publishPacket)
                 {
-                    await TryProcessReceivedPublishPacketAsync(publishPacket, cancellationToken).ConfigureAwait(false);
+                    EnqueueReceivedPublishPacket(publishPacket);
                 }
                 else if (packet is MqttPubRelPacket pubRelPacket)
                 {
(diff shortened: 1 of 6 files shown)


chkr1011:
I also tried to fix the UnitTest at the same time so that my change is already there since a couple of minutes 😄 Please merge master into your branch. I am not sure which fix is better but I think your strategy might be better. I added a comment which describes the problem at the SpinWait.

Regarding async processing. There is already an open pull request with that feature. In the comments I proposed to add some parameter in the context of the message handler like other SDKs do in Azure (AcknowlegeMode = Auto|Manual etc.).

PSanetra:
@chkr1011 Interesting approach :smile:

I think the strategy without the SpinWait is cleaner. It is also possible that you would need to declare the `IsFinalized` property as volatile, because it is accessed from different Threads in parallel. (Not sure how often this would cause issues in the real world.)

Regarding PR #875: I think that PR is related but does not fully solve the issue I try to solve in this PR. PR #875 tries to implement concurrent processing which can violate the order of message processing. This might be ok in some use cases, but I think the default should be to maintain the order of message processing.

This PR maintains the order of message processing and just starts one additional `Task` which dequeues messages from an `AsyncQueue`. Inside the packet receiver loop the publish packets are just getting enqueued and then the loop continues. This solves the issues I mentioned.

I will rebase this PR onto the master and keep the strategy of this PR for the unit test fix.

chkr1011:
Ah OK so only processing of incoming application messages is moved to a dedicated thread right? That is OK for me because as you already said the order of messages is important.

So we have to update the other pull request as discussed there. But this will not affect this PR.

I never used the volatile keyword so I have to read some docs first 😄 But for that special case I do not think it is required because the unit test is running successful and the SpinWait only may run a little bit longer than required.

chkr1011:
@PSanetra Please let me know if this pull request is fully updated. Then I will start my review.

PSanetra:
@chkr1011 yes I have rebased it again onto the current master. It should now be ready for review.

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -34,6 +34,9 @@ public class MqttClient : Disposable, IMqttClient
         private CancellationTokenSource _backgroundCancellationTokenSource;
         private Task _packetReceiverTask;
         private Task _keepAlivePacketsSenderTask;
+        private Task _publishPacketReceiverTask;
+
+        private AsyncQueue<MqttPublishPacket> _publishPacketReceiverQueue;
```
chkr1011, line 39: I assume this must be disposed as soon as the client gets disposed.

PSanetra:
@chkr1011 the messages in the queue are not lost if they are received via a QoS 1 or QoS 2 subscription as they would be resend after a client restart. QoS 0 messages are semantically allowed to get lost.

Regarding the unbound queue and possible memory leak due to too many messages: 
I think in MQTT 3.1.1 there is no mechanism in the protocol which limits the maximum in-flight messages a broker should send to the client. But I guess there should always be a way to configure this in the broker even if this is not part of the MQTT 3.1.1 specification. (E.g. there is such an option in [vernemq](https://docs.vernemq.com/configuration/options#inflight-messages) for QoS 1 and 2 messages which defaults to 20 in-flight messages)

With MQTT 5 it is possible to configure a [receive maximum](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901083) for the connection. The default is 65535. Sadly this is only specified for QoS 1 and QoS 2 publications, too.

Maybe it would be an option to keep processing QoS 0 messages synchronously and QoS 1 and 2 messages asynchronously?! Otherwise this would mean that the issues #648 and #829 will still apply to processing of QoS 0 messages. Another question: Isn't the broker sending the packets anyway even if the client is not handling them? I guess that would mean that the packets are queued by the OS and could maybe result in such a client crash, too?

Another option would be to just drop new messages if the queue would be getting too big. This would be semantically ok for QoS 0 messages. QoS 1 and QoS 2 messages would be resend by the broker as the client would not send PUBACK and PUBREC packets for those messages back to the broker. I think this option is my favorite one.

chkr1011:
You are right. I was only thinking about QoS 0. So QoS 1 and 2 are covered. Regarding the flooding of the inflight queue: In my opinion the receival will stop if the receive buffer is full. When I am not wrong this is handled by TCP automatically so sending will stop because the buffer on the other side is full. So there is already a change for package loss even with the current implementation. 

I agree that we have to add some limit at broker level. Either a general broker setting (MQTT <= 3.1.1) or an individual value per connection/session (MQTT >= 5.0.0).

TLDR: I think we can proceed with your changes...

chkr1011:
One last thing. Please add a one liner in the nuspec describing the change like the already existing entries.

Source/MQTTnet/Client/MqttClient.cs
PSanetra: Done

PSanetra:
@chkr1011 I have added release notes to the nuspec file.

Should I begin with implementing a packet drop mechanism for the `_publishPacketReceiverQueue` or should we just rely on the broker to not send more in-flight messages than the client could handle?

chkr1011:
@PSanetra In my opinion we need such thing at client and server. Then the users have full control.

chkr1011:
@PSanetra Please check the master branch. After merging this PR I get errors in some unit tests like:
 
  Test method MQTTnet.Tests.Client_Tests.Send_Reply_In_Message_Handler threw exception: 
System.Exception: Client(s) had 1 errors ($[2020-04-04T21:39:36.9102228Z] [client] [12] [MqttClient] [Error]: Error while handling application message.
System.NullReferenceException: Object reference not set to an instance of an object.
   at MQTTnet.Client.MqttClient.PublishAtLeastOnceAsync(MqttPublishPacket publishPacket, CancellationToken cancellationToken) in C:\projects\mqttnet\Source\MQTTnet\Client\MqttClient.

PSanetra:
@chkr1011 I can not reproduce the failing test locally and it seems like the test does not fail anymore in the current build. Did you find the cause?

chkr1011:
I ended up avoiding to set the adapter to null when disconnecting. It might happen that a running task (especially for strange scenarios in unit tests) is finishing after disconnect and dispose. This thread will access the adapter  for result conversion but it is already set to null.

But there are still other errors in the build log. They do not happen on my machine but often at the build server.

PR #895, opened 2020-04-10
Add an x509 certificate parameter

### LINKED ISSUE #796: Setup encryption with X509Certificate directly

borigas:
First, thanks for your maintaining this project. I appreciate it

I'm resubmitting #797 as I don't understand why it was closed without comment. If it's not functionality that's useful to others, please let me know. Based on #800 with similar cert changes but for client side being submitted shortly after my server side cert changes, it seems likely others might be interested. Either way, it's functionality that I need, so I know I would appreciate it being merged so I can stop maintaining a separate fork

Closes #796

Adds a way to set the server's certificate using an X509Certificate2 directly instead of needing to export it to byte[].

Thanks again for your time

### DIFF
--- a/Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
+++ b/Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
@@ -44,13 +44,17 @@ namespace MQTTnet.Implementations
 
             if (options.TlsEndpointOptions?.IsEnabled == true)
             {
-                if (options.TlsEndpointOptions.Certificate == null)
+                if (options.TlsEndpointOptions.Certificate == null && options.TlsEndpointOptions.X509Certificate == null)
                 {
                     throw new ArgumentException("TLS certificate is not set.");
                 }
 
                 X509Certificate2 tlsCertificate;
-                if (string.IsNullOrEmpty(options.TlsEndpointOptions.CertificateCredentials?.Password))
+                if (options.TlsEndpointOptions.X509Certificate != null)
+                {
+                    tlsCertificate = options.TlsEndpointOptions.X509Certificate;
+                }
+                else if (string.IsNullOrEmpty(options.TlsEndpointOptions.CertificateCredentials?.Password))
                 {
                     // Use a different overload when no password is specified. Otherwise the constructor will fail.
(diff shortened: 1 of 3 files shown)


Source/MQTTnet/Implementations/MqttTcpServerAdapter.cs
```diff
@@ -44,13 +44,17 @@ public Task StartAsync(IMqttServerOptions options)
 
             if (options.TlsEndpointOptions?.IsEnabled == true)
             {
-                if (options.TlsEndpointOptions.Certificate == null)
+                if (options.TlsEndpointOptions.Certificate == null && options.TlsEndpointOptions.X509Certificate == null)
```
chkr1011, line 47: Please throw an exception if both is set. In such cases we cannot decide which certificate must be used.

Source/MQTTnet/Server/MqttServerOptionsBuilder.cs
```diff
@@ -2,6 +2,7 @@
 using System.Net;
 using System.Net.Security;
 using System.Security.Authentication;
+using System.Security.Cryptography.X509Certificates;
```
chkr1011, line 5: Does this also work for UWP? The code below is wrapped in an compiler statement but namespace is not.

borigas:
@chkr1011 thanks for taking a look. I think I addressed your feedback. Please take another look when you get a chance. Thanks!

borigas:
@chkr1011 is anything needed to merge this?

chkr1011:
I am not happy with the duplication of the certificate properties. But I will merge it for now and see how to improve the strategy. Having two properties now is a code smell to me. But at the moment I have no perfect other solution. So please expect this to be changed.

chkr1011:
@borigas I found a solution which seems to be fair enough. Please test it and let me know it that works for you as well.

borigas:
Changes in c779b39 look good to me. Thanks!

PR #898, opened 2020-04-13
Fix hanging issue under UWP.

xgstation:
Reference: https://github.com/chkr1011/MQTTnet/issues/897
Introduced local variables for IAsyncAction for StreamSocket.ConnectAsync in MqttTcpChannel.Uwp.cs
Using 1st solution as it retains async/await.

### DIFF
--- a/Source/MQTTnet/Implementations/MqttTcpChannel.Uwp.cs
+++ b/Source/MQTTnet/Implementations/MqttTcpChannel.Uwp.cs
@@ -85,7 +85,7 @@ namespace MQTTnet.Implementations
                     socketProtectionLevel = SocketProtectionLevel.Tls10;
                 }
 
-                await _socket.ConnectAsync(new HostName(_options.Server), _options.GetPort().ToString()).AsTask().ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


JanEggers:
the proposed fix looks like a workaround can you please try .ConfigureAwait(false) instead.

xgstation:
Yes that also worked out.

SeppPenner:
Looks better now. @JanEggers @chkr1011 What do you think?

Source/MQTTnet/Implementations/MqttTcpChannel.Uwp.cs
```diff
@@ -85,7 +85,7 @@ public async Task ConnectAsync(CancellationToken cancellationToken)
                     socketProtectionLevel = SocketProtectionLevel.Tls10;
                 }
 
-                await _socket.ConnectAsync(new HostName(_options.Server), _options.GetPort().ToString(), socketProtectionLevel);
```
JanEggers, line 88: why did you remove the socketProtectionLevel

xgstation: Sorry that wasn't my intent, just fixed it.

PR #904, opened 2020-04-21
fixed netcoreapp3.1 build #903

### LINKED ISSUE #903: IEndpointRouteBuilder' does not contain a definition for 'MapMqtt'
### Describe the bug
I configured mqttnet in. Net core, and the following error occurred when configuring startup: 'iendpointrutebuilder' does not contain a definition for 'mapmqtt' and the best extension method overload 'connectionroutebuilder extensions. Mapmqtt (connectionsroutebuilder, pathstring)' requires a receiver of type 'connectionsroutebuilder'
My. Net core version is 3.1, mqttnet version is 3.0.9

image

### Code example

 
```csharp
  public void Configure(IApplicationBuilder app,  ILoggerFactory loggerFactory)
        {
            app.UseAbp(options => { options.UseAbpRequestLocalization = false; }); // Initializes ABP framework.

            app.UseCors(_defaultCorsPolicyName); // Enable CORS!

            app.UseStaticFiles();

            app.UseRouting();

            app.UseAuthentication();

            app.UseAbpRequestLocalization();

            app.UseEndpoints(endpoints =>
            {
                endpoints.MapMqtt("/mqtt");
            });
              
            app.UseEndpoints(endpoints =>
            {
                endpoints.MapHub<AbpCommonHub>("/signalr");
                endpoints.MapControllerRoute("default", "{controller=Home}/{action=Index}/{id?}");
                endpoints.MapControllerRoute("defaultWithArea", "{area}/{controller=Home}/{action=Index}/{id?}");
            });
          
            // Enable middleware to serve generated Swagger as a JSON endpoint
            app.UseSwagger();
            // Enable middleware to serve swagger-ui assets (HTML, JS, CSS etc.)
            app.UseSwaggerUI(options =>
            {
                options.SwaggerEndpoint(_appConfiguration["App:ServerRootAddress"].EnsureEndsWith('/') + "swagger/v1/swagger.json", "demo_dzt API V1");
                options.IndexStream = () => Assembly.GetExecutingAssembly()
                    .GetManifestResourceStream("demo_dzt.Web.Host.wwwroot.swagger.ui.index.html");
            }); // URL: /swagger
        }
```
image

JanEggers:
fixes #903 

@chkr1011 please test this when I run build.ps1 the created package depends on version 2.5.1 of mqttnet instead of 0.0.1 so i could not test the include properly

### DIFF
--- a/Build/MQTTnet.AspNetCore.nuspec
+++ b/Build/MQTTnet.AspNetCore.nuspec
@@ -15,12 +15,19 @@
     <copyright>Copyright Christian Kratky 2016-2020</copyright>
     <tags>MQTT Message Queue Telemetry Transport MQTTClient MQTTServer Server MQTTBroker Broker NETStandard IoT InternetOfThings Messaging Hardware Arduino Sensor Actuator M2M ESP Smart Home Cities Automation Xamarin</tags>
     <dependencies>
-      <dependency id="MQTTnet" version="$nugetVersion" />
 
-      <dependency id="Microsoft.AspNetCore.Connections.Abstractions" version="3.1.3" />
-      <dependency id="Microsoft.AspNetCore.Http.Connections" version="1.1.0"  />
-      <dependency id="Microsoft.AspNetCore.WebSockets" version="2.2.1" />
-      <dependency id="Microsoft.Extensions.Hosting.Abstractions" version="3.1.3" />
+
+      <group targetFramework="netstandard2.0">
+        <dependency id="MQTTnet" version="$nugetVersion" />
+        <dependency id="Microsoft.AspNetCore.Connections.Abstractions" version="3.1.3" />
+        <dependency id="Microsoft.AspNetCore.Http.Connections" version="1.1.0"  />
+        <dependency id="Microsoft.AspNetCore.WebSockets" version="2.2.1" />
+        <dependency id="Microsoft.Extensions.Hosting.Abstractions" version="3.1.3" />
+      </group>
+      <group targetFramework="netcoreapp3.1">
+        <dependency id="MQTTnet" version="$nugetVersion" />
+        <dependency id="System.IO.Pipelines" version="4.7.1" />
+      </group>
     </dependencies>
   </metadata>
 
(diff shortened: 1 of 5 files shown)


chkr1011:
@JanEggers I am confused now 😄  For it also does not even compile. Why are you trying to reference the binaries in the TestApp from a nuget instead the project? I don't see the point in doing this.

chkr1011:
I fixed the build for now. I don't know if this is the final solution but now I will try to fix the issue you mentioned. But is this really an issue? Isn't this the minimum version? Or is this the exact version?

JanEggers:
> Why are you trying to reference the binaries in the TestApp from a nuget instead the project?

this was the best I could come up with to verify that the nuget contains the netcoreapp build. if you have a better idea or if we just can asume it works once the nuspec and build.ps1 is fixed you can revert back to the project reference.

chkr1011:
How did you find out that the build is missing? You added the new line so that the netcoreapp3.1 is added now. Is that already the fix? By the way: I finally fixed the build using the CI...

JanEggers:
> How did you find out that the build is missing? 
#903
> You added the new line so that the netcoreapp3.1 is added now. Is that already the fix?
i assume that fixed it but i was not able to test it

> By the way: I finally fixed the build using the CI...

nice

chkr1011:
OK I will release 3.0.10 as beta. Then we can test. I am too lazy to setup a test nuget source etc. 😄 To me your change at the nuspec should be enough. But if this works. What is about net core 3.0 etc?

JanEggers:
> What is about net core 3.0 etc?

meh not supported! even microsoft supports only 3.1 and 2.1, 3.1 is now added and 2.1 should work with the netstandard2.0 build. im really looking forward dotnet5. they promise to unify this mess so there is just dotnet 5 and not some standard and some runtime and you have to target runtime because the standard misses out on features

SeppPenner:
> im really looking forward dotnet5. they promise to unify this mess so there is just dotnet 5 and not some standard and some runtime and you have to target runtime because the standard misses out on features

Same here... :D Hopefully makes things easier.

Issue #932, opened 2020-05-22
TestParallelWrites failing


chkr1011:
@JanEggers I saw that the Unit Test "TestParallelWrites" fails from time to time when being executed 20 times in a row (sometimes more sometimes less). I don't think that it is related to my latest change (replacing the semaphore with async lock). 

The errors usually indicate that the buffer length management  is broken.

```
 TestParallelWrites
   Source: MqttConnectionContextTest.cs line 40
   Duration: 87 ms

  Message: 
    Test method MQTTnet.AspNetCore.Tests.MqttConnectionContextTest.TestParallelWrites threw exception: 
    System.ArgumentOutOfRangeException: Number was less than the array's lower bound in the first dimension. (Parameter 'destinationIndex')
  Stack Trace: 
    Array.Copy(Array sourceArray, Int32 sourceIndex, Array destinationArray, Int32 destinationIndex, Int32 length, Boolean reliable)
    Array.Copy(Array sourceArray, Int32 sourceIndex, Array destinationArray, Int32 destinationIndex, Int32 length)
    MqttPacketWriter.Write(Byte[] buffer, Int32 offset, Int32 count) line 140
    MqttPacketWriter.Write(ArraySegment`1 buffer) line 196
    MqttPacketWriter.WriteVariableLengthInteger(UInt32 value) line 82
    MqttV310PacketFormatter.Encode(MqttBasePacket packet) line 47
    MqttPacketFormatterAdapter.Encode(MqttBasePacket packet) line 51
    MqttConnectionContext.SendPacketAsync(MqttBasePacket packet, TimeSpan timeout, CancellationToken cancellationToken) line 166
    <<TestParallelWrites>b__1>d.MoveNext() line 52
    --- End of stack trace from previous location where exception was thrown ---
    MqttConnectionContextTest.TestParallelWrites() line 56
    ThreadOperations.ExecuteWithAbortSafety(Action action)
```

Do you have any idea what is wrong?

JanEggers:
I will have a look

JanEggers:
i had alook on your latest changes why did you add back the pipelines dependency to mqttnet.aspnetcore? also why did you upgrade to connections 1.1 I specificly downgraded all dependencies as much as possible. if the chage was inteded pleas also update the nupkg accordingly otherwise we will have a mess in packaging again. I wonder if there is really any need for the nuspec at all or if all informations can be derived from csproj. (thats what i would prefere to avoid conflicts)

[...]

PR #935, opened 2020-05-27
fixed parallel writes #932

JanEggers:
you broke it with https://github.com/chkr1011/MQTTnet/commit/a58ab607ad3e4f8e56ebeca7980e64dd16b28851

can you please create a pr when changing something in the aspnetcore part? I know its tidious but the review process really helps avoid errors.

I also cranked up the numbers in the test but it was still just failing 1/5 dont know how to make it fail more reliable.


JanEggers:
also I dont know if it make sense to push a new package since 3.0.10 and 3.0.11 contain the bug

JanEggers:
and just FYI the parallel stacks debug window in vs helps track down this kind of issues really fast

chkr1011:
Oops 👎 Sorry for that. I will make pull requests. But still wondering why that happens while simply refactoring... anyway. I will have a look at the Parallel Stacks window. Never used that before.

PR #937, opened 2020-06-09
Added checks for null references in the MqttTcpChannel.

SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/766.

### DIFF
--- a/Source/MQTTnet/Implementations/MqttTcpChannel.cs
+++ b/Source/MQTTnet/Implementations/MqttTcpChannel.cs
@@ -127,6 +127,11 @@ namespace MQTTnet.Implementations
                 // Workaround for: https://github.com/dotnet/corefx/issues/24430
                 using (cancellationToken.Register(Dispose))
                 {
+                    if (this._stream == null)
+                    {
+                        throw new ObjectDisposedException(nameof(_stream));
+                    }
+
                     return await _stream.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
                 }
             }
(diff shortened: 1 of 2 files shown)


SeppPenner:
I don't know it this is what you meant with "so we should properly implement a dispose guard that first checks wether the channel is disposed and throw an objectdisposed exception." @JanEggers...

Source/MQTTnet/Implementations/MqttTcpChannel.cs
```diff
@@ -127,6 +127,11 @@ public async Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancellat
                 // Workaround for: https://github.com/dotnet/corefx/issues/24430
                 using (cancellationToken.Register(Dispose))
                 {
+                    if (this._stream == null)
```
JanEggers, line 130: fetch the field into a local variable and perfom the check and the ReadAsync on the local variable otherwise it is not threadsafe

Source/MQTTnet/Implementations/MqttTcpChannel.cs
```diff
@@ -157,6 +162,11 @@ public async Task WriteAsync(byte[] buffer, int offset, int count, CancellationT
...
+                    {
```
JanEggers, line 166: same here

Source/MQTTnet/Implementations/MqttTcpChannel.cs
SeppPenner: Makes sense 👍

SeppPenner: Is resolved now.

Source/MQTTnet/Implementations/MqttTcpChannel.cs
SeppPenner: Is resolved now.

PR #938, opened 2020-06-09
Added interceptor for undelivered messages.

SeppPenner:
Makes https://github.com/chkr1011/MQTTnet/pull/864 obsolete.

### DIFF
--- a/Source/MQTTnet/MqttApplicationMessage.cs
+++ b/Source/MQTTnet/MqttApplicationMessage.cs
@@ -29,5 +29,7 @@ namespace MQTTnet
         public byte[] CorrelationData { get; set; }
 
         public List<uint> SubscriptionIdentifiers { get; set; }
+
+        public uint DeliveryCount { get; set; }
     }
 }
--- a/Source/MQTTnet/MqttApplicationMessage.cs
+++ b/Source/MQTTnet/MqttApplicationMessage.cs
@@ -29,7 +29,5 @@ namespace MQTTnet
         public byte[] CorrelationData { get; set; }
 
         public List<uint> SubscriptionIdentifiers { get; set; }
-
-        internal uint DeliveryCount { get; set; }
     }
 }
--- a/Source/MQTTnet/MqttApplicationMessage.cs
+++ b/Source/MQTTnet/MqttApplicationMessage.cs
@@ -30,6 +30,6 @@ namespace MQTTnet
 
         public List<uint> SubscriptionIdentifiers { get; set; }
 
-        public uint DeliveryCount { get; set; }
+        internal uint DeliveryCount { get; set; }
     }
 }
--- a/Source/MQTTnet/Server/MqttClientSessionsManager.cs
+++ b/Source/MQTTnet/Server/MqttClientSessionsManager.cs
@@ -228,12 +228,15 @@ namespace MQTTnet.Server
 
                 foreach (var clientSession in _sessions.Values)
                 {
-                    clientSession.EnqueueApplicationMessage(
+                    var isSubscribed = clientSession.EnqueueApplicationMessage(
                         applicationMessage,
                         sender?.ClientId,
                         false);
 
-                    deliveryCount++;
+                    if (isSubscribed)
+                    {
+                        deliveryCount++;
+                    }
                 }
 
                 if (deliveryCount == 0)
--- a/Source/MQTTnet/Server/MqttClientSessionsManager.cs
+++ b/Source/MQTTnet/Server/MqttClientSessionsManager.cs
@@ -238,12 +238,14 @@ namespace MQTTnet.Server
 
                 if (deliveryCount == 0)
                 {
-                    if (_options.UndeliveredMessageInterceptor == null)
+                    var undeliveredMessageInterceptor = _options.UndeliveredMessageInterceptor;
+
+                    if (undeliveredMessageInterceptor == null)
                     {
                         return;
                     }
 
-                    await _options.UndeliveredMessageInterceptor.InterceptApplicationMessagePublishAsync(new MqttApplicationMessageInterceptorContext(sender?.ClientId, sender?.Session?.Items, applicationMessage));
+                    await undeliveredMessageInterceptor.InterceptApplicationMessagePublishAsync(new MqttApplicationMessageInterceptorContext(sender?.ClientId, sender?.Session?.Items, applicationMessage));
                 }
             }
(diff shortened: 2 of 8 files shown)


SeppPenner:
@chkr1011 Just let me know if you need any changes, please :) I guess @cshark-inator won't adjust his pull request soon, so I added the changes here.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -224,13 +224,25 @@ async Task TryProcessNextQueuedApplicationMessageAsync(CancellationToken cancell
...
                         applicationMessage,
                         sender?.ClientId,
                         false);
                 }
+
+                if (applicationMessage.DeliveryCount == 0)
+                {
+                    if (_options.UndeliveredMessageInterceptor == null)
+                    {
+                        throw new OperationCanceledException(nameof(_options.UndeliveredMessageInterceptor));
```
JanEggers, line 241: return here dont throw if the user does not care the server should not blow up

Source/MQTTnet/MqttApplicationMessage.cs
```diff
@@ -29,5 +29,7 @@ public class MqttApplicationMessage
         public byte[] CorrelationData { get; set; }
 
         public List<uint> SubscriptionIdentifiers { get; set; }
+
+        public uint DeliveryCount { get; set; }
```
JanEggers, line 33: this should not be on the public api

SeppPenner: Ok, I will change this.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
SeppPenner: Makes sense. I wasn't sure how to properly handle this :)

Source/MQTTnet/MqttApplicationMessage.cs
SeppPenner: How would you avoid this? It`s used in the `MqttClientSession` and `MqttClientSessionManager`. Can we store this in the session somehow?

Source/MQTTnet/Server/MqttClientSessionsManager.cs
SeppPenner: This is fixed now.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -224,13 +224,25 @@ async Task TryProcessNextQueuedApplicationMessageAsync(CancellationToken cancell
...
+
                 foreach (var clientSession in _sessions.Values)
                 {
                     clientSession.EnqueueApplicationMessage(
                         applicationMessage,
                         sender?.ClientId,
                         false);
                 }
+
+                if (applicationMessage.DeliveryCount == 0)
```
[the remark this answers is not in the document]
SeppPenner: That should be possible. However, `EnqueueApplicationMessage` doesn't return anything, it's a `void` type method. I will implement this and let you check again.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -224,13 +224,25 @@ async Task TryProcessNextQueuedApplicationMessageAsync(CancellationToken cancell
...
+                {
+                    if (_options.UndeliveredMessageInterceptor == null)
```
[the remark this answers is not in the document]
SeppPenner: Ok, this is fixed now.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -224,12 +224,28 @@ async Task TryProcessNextQueuedApplicationMessageAsync(CancellationToken cancell
...
+                var deliveryCount = 0;
+
                 foreach (var clientSession in _sessions.Values)
                 {
                     clientSession.EnqueueApplicationMessage(
                         applicationMessage,
                         sender?.ClientId,
                         false);
+
+                    deliveryCount++;
```
JanEggers, line 236: this makes no sense we at least have to check if this session is subscribed to that message

SeppPenner: How can I do that? 😅

SeppPenner: @JanEggers By the way, that's what I meant with the other comment already:

> However, EnqueueApplicationMessage doesn't return anything, it's a void type method. I will implement this and let you check again.

JanEggers: ahh :) 

you could add a task completion source to each message in ApplicationMessagesQueue
return the task and complete it once the session checked it.

then wait for all sessions in the server and count the transmissions.

I would cancel this PR as its a waste of performance and makes no sense in general.

If you want to know a message was delivered from A to B. B has to send a new message aka response to A

SeppPenner: @chkr1011 What do you think? Should we close this?

> I would cancel this PR as its a waste of performance and makes no sense in general.

> If you want to know a message was delivered from A to B. B has to send a new message aka response to A

That's true. Honestly makes no sense after thinking about this a bit more :) I agree with you, we should close this. Let's wait for @chkr1011 to have his opinion here as well.

Note: I just implemented @cshark-inator's request to get a bit more things done here :)

chkr1011: In my opinion this should be fairly easy. The method _EnqueueApplicationMessage_ can easily return a boolean because it uses it internally as well (checkSubscriptionsResult.IsSubscribed). We can use this to increase the delivery count. If this results in 0 we can invoke the handler.

In my opinion this is not a killer feature but if someone needs that approach we should add it. There is be no noticeable performance impact because we are only increasing an integer number.

@JanEggers I assume you are referring to a delivery count which includes the transmit to the client. I only refer to that fact that a message was subscribed and will be processed in the near future. A use case could be identification of "dead" messages which are not subscribed by anyone. So a valid feature request in my opinion.

SeppPenner: Ok, sounds good. I will check this and make some changes there.

SeppPenner:
I have updated the pull request. Please take a look again and tell me if there are any changes needed.

PR #940, opened 2020-06-11
CrossPlatformSocket: Added SocketOptionName.ReuseAddress = true.

SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/494 (At least as far as I understood it).
Suggestion from @zvrba.

Might also fix https://github.com/chkr1011/MQTTnet/issues/872.

### DIFF
--- a/Source/MQTTnet/Implementations/CrossPlatformSocket.cs
+++ b/Source/MQTTnet/Implementations/CrossPlatformSocket.cs
@@ -86,6 +86,7 @@ namespace MQTTnet.Implementations
         {
             if (localEndPoint is null) throw new ArgumentNullException(nameof(localEndPoint));
 
+            _socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, true);
             _socket.Bind(localEndPoint);
         }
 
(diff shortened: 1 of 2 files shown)


Source/MQTTnet/Implementations/CrossPlatformSocket.cs
```diff
@@ -86,6 +86,7 @@ public void Bind(EndPoint localEndPoint)
         {
             if (localEndPoint is null) throw new ArgumentNullException(nameof(localEndPoint));
 
+            _socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReuseAddress, true);
```
chkr1011, line 89: I don't think we need this. There is already a property which sets the socket option. The property is also set when the server starts. Probably the user who has problems with that only needs to enable this in the server options.

SeppPenner: Ok, can you tell him how to enable this, please? https://github.com/chkr1011/MQTTnet/issues/872.

I will close the pull request soon.

SeppPenner: Will be closed as I have given a hint to the original person who had this problem.

SeppPenner:
Not needed any longer, see https://github.com/chkr1011/MQTTnet/issues/872#issuecomment-644053219 for more information.

PR #941, opened 2020-06-11
Added missing disconnect reason codes.

SeppPenner:
I added the missing disconnect related reason codes from http://docs.oasis-open.org/mqtt/mqtt/v5.0/cs01/mqtt-v5.0-cs01.html#_Toc502667250. (MQTT5).

What else do I need to do to get https://github.com/chkr1011/MQTTnet/issues/847 working?

Fixes https://github.com/chkr1011/MQTTnet/issues/847.

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -309,6 +309,7 @@ namespace MQTTnet.Client
         async Task DisconnectInternalAsync(Task sender, Exception exception, MqttClientAuthenticateResult authenticateResult)
         {
             var clientWasConnected = IsConnected;
+            var reasonCode = MqttClientDisconnectReason.NormalDisconnection;
 
             TryInitiateDisconnect();
             _isConnected = false;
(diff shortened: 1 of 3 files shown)


Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -326,6 +327,7 @@ async Task DisconnectInternalAsync(Task sender, Exception exception, MqttClientA
             catch (Exception adapterException)
             {
                 _logger.Warning(adapterException, "Error while disconnecting from adapter.");
+                reasonCode = MqttClientDisconnectReason.UnspecifiedError;
```
chkr1011, line 330: Why do you use "UnspecifiedError" here and below "ImplementationSpecificError"? I am just curious why this is not the same 😄

SeppPenner: No idea :D Which one should I use? I wasn't really sure about that...

chkr1011: I also have no idea 😄 But to me they should be at least equal. My proposal is "UnspecifiedError" for both cases.

PR #954, opened 2020-07-09
improve exceptions in case of corrupt packages

### LINKED ISSUE #788: cpu 100+%
English is poor
After using for a period of time, the usage rate of cpu will always exceed 100%.

**Platform:**

image

**Top:**

image

**Trace File:**

[trace.zip](https://github.com/chkr1011/MQTTnet/files/3802912/trace.zip)

JanEggers:
and close pipes after exception occoured to avoid lingering connections should fix #788

@chkr1011 i was not able to reproduce the high cpu load. but i added some testcases and unified the api behavior when dealing with corrupt packages.

I dont quite like the error message that is checking for at least 7 bytes if you have a better idea let me know.

the only thing I found was that if for some reason the clientSessionmanager would call ReceiveAsync after an exception the asp.net connection would still try to read data, thats why explicitly close them now when an exception was thrown. this should however in practise not make a difference as the clientSessionmanager should exit its loop if the first exception is thrown and thus let kestrel cleanup the connection

### DIFF
--- a/Source/MQTTnet/Formatter/MqttPacketFormatterAdapter.cs
+++ b/Source/MQTTnet/Formatter/MqttPacketFormatterAdapter.cs
@@ -120,6 +120,14 @@ namespace MQTTnet.Formatter
         {
             if (receivedMqttPacket == null) throw new ArgumentNullException(nameof(receivedMqttPacket));
 
+            if (receivedMqttPacket.Body.Length < 7)
+            {
+                // 2 byte protocol name length
+                // at least 4 byte protocol name
+                // 1 byte protocol level
+                throw new MqttProtocolViolationException("Mqtt Connect packet must have at least 7 bytes");
+            }
+
             var protocolName = receivedMqttPacket.Body.ReadStringWithLengthPrefix();
             var protocolLevel = receivedMqttPacket.Body.ReadByte();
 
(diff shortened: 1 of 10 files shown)


Source/MQTTnet/Formatter/MqttPacketFormatterAdapter.cs
```diff
@@ -120,6 +120,14 @@ MqttProtocolVersion ParseProtocolVersion(ReceivedMqttPacket receivedMqttPacket)
         {
             if (receivedMqttPacket == null) throw new ArgumentNullException(nameof(receivedMqttPacket));
 
+            if (receivedMqttPacket.Body.Length < 7)
+            {
+                // 2 byte protocol name length
+                // at least 4 byte protocol name
+                // 1 byte protocol level
+                throw new MqttProtocolViolationException("Mqtt Connect packet must have at least 7 bytes");
```
chkr1011, line 128: Please consider writing packets in upper case all the time. I usually do this in all trace messages so that it looks like "MQTT CONNECT packet must ...". The idea behind this is to have a more natural reading flow etc.

chkr1011:
@JanEggers The 7 byte checks looks OK to me. I also have no better idea. The comment is explaining the number perfectly. Without the comment I would have doubts but this is OK. Maybe you can out the 7 into a cost with a name like "MinimumConnectPacketLength" etc.

PR #974, opened 2020-08-12
Update EndpointRouterExtensions.cs

osioReitech:


SeppPenner:
@JanEggers Is this okay? I guess, your're the most experienced one with the extension methods :D

JanEggers:
@osioReitech what is the issue you are trying to solve?

the api requires endpointrouting that is only present in aspnet core 3.1 that is why the build breaks with:

Extensions\EndpointRouterExtensions.cs(9,41): error CS0246: The type or namespace name 'IEndpointRouteBuilder' could not be found (are you missing a using directive or an assembly reference?)

osioReitech:
Good afternoon. Sorry for no further explanation. According to the documentation, this extension method should be accessible from .NET Core 3.0.

But unfortunately that didn't work on my version of .NET Core 3.0

Documentation link: https://github.com/chkr1011/MQTTnet/wiki/Server#aspnet-core-30-since-mqtt-version-309

SeppPenner:
I will update the wiki (Seems to be only working with 3.1+). Sorry for the inconvenience...

PR #993, opened 2020-09-21
Tried to add a MqttServerOptions fluent method for the client disconn…

SeppPenner:
…ected handler.

@chkr1011 and @JanEggers: I wanted to add the disconnected handler within the `fluent calls` like `WithApplicationMessageInterceptor` and so on. Can you guide me on what's missing, please? It seems like there is something I didn't understand there... (The disconnected handler is somewhere else already getting called).

### DIFF
--- a/Source/MQTTnet/Server/IMqttServerOptions.cs
+++ b/Source/MQTTnet/Server/IMqttServerOptions.cs
@@ -27,5 +27,7 @@ namespace MQTTnet.Server
         IMqttRetainedMessagesManager RetainedMessagesManager { get; }
 
         IMqttServerApplicationMessageInterceptor UndeliveredMessageInterceptor { get; set; }
+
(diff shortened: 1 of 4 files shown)


chkr1011:
@SeppPenner To me your strategy looks good. But I have no idea why you renamed the "handler" to "callback". What exactly do you mean with "called somewhere else"?

SeppPenner:
@chkr1011 I have no idea either. Changed it back...

The disconnected handler is already called in `SafeNotifyClientDisconnectedAsync()` in https://github.com/chkr1011/MQTTnet/blob/d63143bdc9d45c62c48dd39718e0c8616c278528/Source/MQTTnet/Server/MqttClientSessionsManager.cs#L331 if I see it correctly, so there is nothing else required?

chkr1011 (approved):

chkr1011:
All fine 👍 Sorry for the long delay.

SeppPenner:
Nice :)

Issue #990, opened 2020-09-15
MqttClientSessionsManager should use lock


zhou881117:
log by ClientConnectedHandler ClientDisconnectedHandler ClientSubscribedTopicHandler：
Client connected 4194328.
Client connected 4194328.
Client connected 4194328.
Client disconnected 4194328.
Client disconnected 4194328.
Client Subscribed 4194328 **/devices/update/**/4194328.
info: Microsoft.Hosting.Lifetime[0]

the client can not receive subscribed topic

client is a rtos device，sometimes  subscribe  no effect when connected

JanEggers:
can you create a test / proof of what is wrong? MqttClientSessionsManager is using concurrentdictionary so it should be good.

zhou881117:
wait for a monent
i test 

first  the client device is reconnecting
second start the server 

at this monent, the device connected ,then subscribe the topic,most probable,subscibed not effect.

i will debug the  code.

zhou881117:
mqttnet ver:3.0.8

bug Scene:
first：a rtos  device connecting to server  by a loop
second： start the  asp .net core 3.1 server 
result：the client can not receive subscribed topic

i solve it by a lock HashSet，code:
 public class Startup
    {
        private HashSet<string> clientIds = new HashSet<string>();

...
 public void ConfigureServices(IServiceCollection services)
...
var mqttServerOptions = new MqttServerOptionsBuilder().WithConnectionValidator(t =>
            {
                bool res = deviceManager.OnLogin(t.Username, t.Password, t.ClientId);               
                if (res)
                {
                    Console.WriteLine($"Authorized: ClientId: {t.ClientId}");
                    lock (clientIds)
                    {
                        if (clientIds.Contains(t.ClientId))
                        {
                            t.ReasonCode = MQTTnet.Protocol.MqttConnectReasonCode.ServerUnavailable; //.NotAuthorized;
                        }
                        else
                        {
                            clientIds.Add(t.ClientId);
                            t.ReasonCode = MQTTnet.Protocol.MqttConnectReasonCode.Success; //.NotAuthorized;
                        }
                    }

                }
                else
                {
                    Console.WriteLine($"NotAuthorized: userName: {t.Username} password:{t.Password}");
                    t.ReasonCode = MQTTnet.Protocol.MqttConnectReasonCode.NotAuthorized;
                }
            });
...
   public void Configure(IApplicationBuilder app, IWebHostEnvironment env,
...
[...]

PR #1004, opened 2020-10-11
improved behavior when multiple connections fight over a session

JanEggers:
improvements / fixes for #990 

there is still an issue in the client. right now the client throws argumentnullexception when receiving a diconnect packet.

i added the required options class but there is an enum collision.

Does a conversion for the reason codes exist?

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -561,7 +561,7 @@ namespace MQTTnet.Client
                 {
                     await SendAsync(new MqttPingRespPacket(), cancellationToken).ConfigureAwait(false);
                 }
-                else if (packet is MqttDisconnectPacket disc)
+                else if (packet is MqttDisconnectPacket disconnectPacket)
                 {
                     // Also dispatch disconnect to waiting threads to generate a proper exception.
                     _packetDispatcher.Dispatch(packet);
@@ -569,7 +569,7 @@ namespace MQTTnet.Client
                     await DisconnectAsync(new MqttClientDisconnectOptions() 
                     {
                         // todo conversion
-                        ReasonCode = disc.ReasonCode
+                        ReasonCode = (MqttClientDisconnectReason)(disconnectPacket.ReasonCode ?? MqttDisconnectReasonCode.UnspecifiedError)
                     }, cancellationToken).ConfigureAwait(false);
                 }
                 else if (packet is MqttAuthPacket authPacket)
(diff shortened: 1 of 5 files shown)


Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -561,12 +561,16 @@ async Task TryProcessReceivedPacketAsync(MqttBasePacket packet, CancellationToke
                 {
                     await SendAsync(new MqttPingRespPacket(), cancellationToken).ConfigureAwait(false);
                 }
-                else if (packet is MqttDisconnectPacket)
+                else if (packet is MqttDisconnectPacket disc)
```
chkr1011, line 564: Please use long names like "disconnectPaket".

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -561,12 +561,16 @@ async Task TryProcessReceivedPacketAsync(MqttBasePacket packet, CancellationToke
...
                 {
                     // Also dispatch disconnect to waiting threads to generate a proper exception.
                     _packetDispatcher.Dispatch(packet);
 
-                    await DisconnectAsync(null, cancellationToken).ConfigureAwait(false);
+                    await DisconnectAsync(new MqttClientDisconnectOptions() 
+                    {
+                        // todo conversion
+                        ReasonCode = disc.ReasonCode
```
chkr1011, line 572: Is this only the conversion between the client  API and the internal protocol enum? If yes I usually cast the enum as int and then to the new enum (if positions ans numbers are matching). The reason is that the client may have less enum values because other code is handling this or it is dangerous if the user is able to set some "wrong" values which do not follow the internal logic etc.

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -364,38 +368,39 @@ async Task<MqttClientConnection> CreateClientConnectionAsync(MqttConnectPacket c
...
-                    _logger.Verbose("Created a new session for client '{0}'.", connectPacket.ClientId);
-                }
+                });
 
                 var connection = new MqttClientConnection(connectPacket, channelAdapter, session, _options, this, _retainedMessagesManager, onStart, onStop, _rootLogger);
+                MqttClientConnection existingConnection = null;
+                _connections.AddOrUpdate(connectPacket.ClientId, key => 
+                {
+                    return connection;
+                }, (key, tempexistingConnection) =>
```
chkr1011, line 394: Please rename to "tempExistingConnection".

chkr1011 (changes requested):

PR #1014, opened 2020-10-28
Fix evaluation of IgnoreCertificateRevocationErrors for setting CertificateRevocationCheckMode of SslClientAuthenticationOptions

Jay424:
When IgnoreCertificateRevocationErrors is set to true X509RevocationMode.NoCheck should be set. 
The if was just flipped.
Found this during updating MQTTNET in my project.


SeppPenner:
What was this about?

Jay424:
Was my first pull request and I f***** it up :-)
Can be deleted if possible.

SeppPenner:
Ah ok. It's already close, so nothing else needs to be done :)

PR #1019, opened 2020-10-29
Feature/single connection monitoring task

chkr1011:


JanEggers:
any benchmarks of the performance improvements?

chkr1011:
@JanEggers No but I will create some. But these improvements are just minor ones.

Could you please have a look at the code changes? My strategy was to clean some things up to:
1. Reduce "ObjectDisposedExceptions"
2. Fix race conditions in connection attempts.
3. Reduce the Task count by moving all the keep aplive checks to a single tasks/thread.
4. Fix disconnect packets for takeover / keep alive disconnect.

Source/MQTTnet/Adapter/IMqttChannelAdapter.cs
```diff
@@ -21,17 +21,15 @@ public interface IMqttChannelAdapter : IDisposable
...
         long BytesReceived { get; }
 
-        Action ReadingPacketStartedCallback { get; set; }
-
-        Action ReadingPacketCompletedCallback { get; set; }
+        bool IsReadingPacket { get; }
 
         Task ConnectAsync(TimeSpan timeout, CancellationToken cancellationToken);
 
         Task DisconnectAsync(TimeSpan timeout, CancellationToken cancellationToken);
```
JanEggers, line 28: can we also remove the timeout for all the apis and just use the cancellation token?

Source/MQTTnet/Implementations/PlatformAbstractionLayer.cs
```diff
@@ -1,20 +1,25 @@
...
-        public static Task CompletedTask
+        public static readonly byte[] EmptyByteArray = new byte[0];
+
+#if NET452
+        public static Task CompletedTask => Task.FromResult(0);
+#else
+        public static Task CompletedTask => Task.CompletedTask;
+#endif
+
+        public static void Sleep(TimeSpan timeout)
```
JanEggers, line 16: this makes no sense to me just use Thread.Sleep if you need to blocking wait

Source/MQTTnet/Internal/AsyncLock.cs
```diff
@@ -7,9 +7,11 @@ namespace MQTTnet.Internal
     // From Stephen Toub (https://blogs.msdn.microsoft.com/pfxteam/2012/02/12/building-async-coordination-primitives-part-6-asynclock/)
     public sealed class AsyncLock : IDisposable
     {
-        readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
+        readonly object _syncRoot = new object();
```
JanEggers, line 10: why lock a semaphore?

Source/MQTTnet/Internal/AsyncQueue.cs
```diff
@@ -24,22 +28,37 @@ public async Task<AsyncQueueDequeueResult<TItem>> TryDequeueAsync(CancellationTo
             {
                 try
                 {
-                    if (_semaphore == null)
+                    Task task;
+                    lock (_syncRoot)
```
JanEggers, line 32: again why lock the whole purpose of _semaphore is to avoid lock. if you want to block just use lock and remove _semaphore

Source/MQTTnet/Server/MqttServerKeepAliveMonitor.cs
```diff
@@ -0,0 +1,119 @@
...
+        void DoWork(CancellationToken cancellationToken)
+        {
+            try
+            {
+                _logger.Info("Starting keep alive monitor.");
+
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    TryMaintainConnections();
+                    PlatformAbstractionLayer.Sleep(_options.KeepAliveMonitorInterval);
```
JanEggers, line 44: i would use task.delay and pass in the cancellationtoken. why use blocking sleep

Source/MQTTnet/Adapter/IMqttChannelAdapter.cs
chkr1011: I was also thinking about this already. I would say yes but not yet since is this an official feature. But I have it in my notes.

Source/MQTTnet/Implementations/PlatformAbstractionLayer.cs
chkr1011: Thread.Sleep is not available for all build targets (when I am not wrong it is not available in UWP).

Source/MQTTnet/Internal/AsyncLock.cs
chkr1011: So make sure that no other thread will access the semaphore while it is being disposed. It will be set to _null_ when the async lock being released but in some cases (race conditions) I got ObjectDisposedExceptions. With the lock the semaphore is either available and ready to use or _null_ (whis is fine for the releaser).

Source/MQTTnet/Internal/AsyncQueue.cs
chkr1011: The initial idea of the semaphore was to have a lock which supports async/await.  This is not possible with a classic lock. The thread will be waiting and not returned to the thread pool.

Source/MQTTnet/Server/MqttServerKeepAliveMonitor.cs
chkr1011: My intention is that the thread is started as "long running" and thus is not relying in available threads in the thread pool when using async/await.

Source/MQTTnet/Implementations/PlatformAbstractionLayer.cs
```diff
@@ -7,8 +7,12 @@ public static class PlatformAbstractionLayer
     {
 #if NET452
         public static Task CompletedTask => Task.FromResult(0);
+
+        public static byte[] EmptyByteArray { get; } = new byte[0];
```
JanEggers, line 11: this exists in the box array.empty

chkr1011:
> any benchmarks of the performance improvements?

I made some quick tests. I started the program a couple of time and repeated the tests. The results are:

Performance_OLD_Async_Handling

Performance_NEW_NETCORE

PR #1020, opened 2020-10-30
Added support for Tls1.3

Dvergatal:


Source/MQTTnet/Client/Options/MqttClientTlsOptions.cs
```diff
@@ -26,7 +26,11 @@ public class MqttClientTlsOptions
         public List<SslApplicationProtocol> ApplicationProtocols { get; set; }
 #endif
 
+#if NETCOREAPP3_0 || NET5_0
+        public SslProtocols SslProtocol { get; set; } = SslProtocols.Tls13;
```
chkr1011, line 30: Is 1.3 backward compatible with 1.2? I have doubts if users can still connect to their servers after this update. Especially when they leave the initial values untouched (so 1.2).

Source/MQTTnet/Implementations/MqttTcpChannel.Uwp.cs
```diff
@@ -39,7 +39,11 @@ public MqttTcpChannel(StreamSocket socket, X509Certificate2 clientCertificate, I
 
             CreateStreams();
 
+#if NETCOREAPP3_0 || NET5_0
+            IsSecureConnection = socket.Information.ProtectionLevel >= SocketProtectionLevel.Tls13;
```
chkr1011, line 43: In my opinion we do not need this different check. Also 1.2 is "secure" because on older frameworks 1.3 is not supported so "1.2" becomes the "most secure" version available.

chkr1011 (changes requested):

Source/MQTTnet/Implementations/MqttTcpChannel.Uwp.cs
Dvergatal: 1.3 is backward compatible i have already checked it. The connection was being made on 1.3, but server was only using up to 1.2 and termshark showed that the established lin was 1.2.

Dvergatal: What do you mean by different check? This one? 
```cs
#if NETCOREAPP3_0 || NET5_0
```

chkr1011: Ah OK then it is fine. In my opinion it can stay at "=> ...Tls12" instead 1.3 because now every connection (which still uses 1.2) is treated as "unsafe".

Dvergatal: Hmm i'm sorry but your sentence sentence is contradicting itself to the conditional. How it can stay at "=> ...Tls12 if every 1.2 connection which uses 1.2 is treated as "not secure"? If it is "not secure" than i think it should rather be changed? Or am i missing something?

Dvergatal: Ahhh ok :] i have read it once again and checked the code. Ok this can be changed to the previous one :] but yes 1.2 is insecure for example by military :]

chkr1011: I can also change this after merge.

Source/MQTTnet/Client/Options/MqttClientOptionsBuilderTlsParameters.cs
```diff
@@ -19,7 +19,11 @@ public class MqttClientOptionsBuilderTlsParameters
 
         public Func<MqttClientCertificateValidationCallbackContext, bool> CertificateValidationHandler { get; set; }
 
+#if NETCOREAPP3_0 || NET5_0
```
chkr1011, line 22: Will NETCOREAPP3_0 work together with the nuget package? There is only 3.1 used. Shall we also move over to 3.1 instead?

    
    <file src="..\Source\MQTTnet\bin\Release\netcoreapp3.1\MQTTnet.*" target="lib\netcoreapp3.1\"/>

chkr1011 (changes requested):

Source/MQTTnet/Client/Options/MqttClientOptionsBuilderTlsParameters.cs
Dvergatal: I have used NETCOREAPP3_0 according to [preprocessor-if](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if) but i think we can move over to 3.1 instead.

chkr1011:
Thank you for your contribution. I will do the changes we discussed. 👍

Dvergatal:
No problem:] thank you too.

PR #1023, opened 2020-11-08
Fix: Compile error, cancellation token removed since the method does not support it

StefanOssendorf:
This fixes the current issue why it's not compiling.

Cancellation token removed because the SendAsync() method does not support it.


StefanOssendorf:
Well looks like more than only the C&P(?) error introduced by commit 06c9d27c872fa422eee1a8bf6d31301e966a2a6d.

Issue #1017, opened 2020-10-29
Asp.net core .Net 5 compatibility?


StefanOssendorf:
### Describe your question
Hello. I am using the Mqtt server with the AspNetCore integration and trying to upgrade to .Net 5 (RC2).
Sadly I get a compile error after upgrading to the latest Mqtt + .Net5 RC2 packages.

The error is:
CS7069	Reference to type 'ConnectionsRouteBuilder' claims it is defined in 'Microsoft.AspNetCore.Http.Connections', but it could not be found.

I found out that the type `ConnectionsRouterBuilder` is removed with .Net 5 hence the error.

The actual question: Is .Net 5 compatibility in work or is it planned? And if so will it be shipped before the .Net 5 release next month?

Thanks in advance

### Which project is your question related to?
- Server (AspNetCore integration)

Edit:

The method in question is 
```csharp
app.UseEndpoints(endpoints =>
{
    endpoints.MapMqtt("/mqtt");
});
```

JanEggers:
> Is .Net 5 compatibility in work or is it planned? 

nothing is planned but I will update MqttNet.AspnetCore as soon as AspnetCore is Released and most likely ifdef my way to compatibility. 

> And if so will it be shipped before the .Net 5 release next month?

no it will certainly not be shipped before .Net 5 release. maybe in december when im on vakation and have some spare time. If you need to early adopt .Net 5 feel free to create a pr.

JanEggers:
this seems to be the the issue:

https://docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0#signalr-usesignalr-and-useconnections-methods-removed

Jossec101:
Hi @StefanOssendorf  @JanEggers 

I made it work on NET 5.0 by just changing these lines in Startup.cs:

Old startup.cs:
`
                      app.UseEndpoints(endpoints =>
                {
                    endpoints.MapMqtt("/mqtt");
                });
`

.NET5 Startup.cs:
`
   app.UseEndpoints(endpoints =>
            {
                endpoints.MapConnectionHandler<MqttConnectionHandler>("/mqtt");
            });
`
I made a quick test and it is working without any issue.

Cheers

JanEggers:
MapMqtt should call exactly that I  guess ms just changed some types so the extension is not recogniced

314159265meow:
[...]

PR #1027, opened 2020-11-10
.Net5.0 TFMs added

StefanOssendorf:
This adds the .Net5.0 TFM to all project and related nuget packages are updated.

**BUT** sadly I could not run any tests since I always got a socket related exception and I don't know why.
Furthermore I could not build the `MQTTnet.TestApp.AspNetCore2` project with the following error: 
`TS2307	Build:Cannot find module 'mqtt' or its corresponding type declarations.	MQTTnet.TestApp.AspNetCore2`

### DIFF
--- a/Source/MQTTnet.AspnetCore/Client/Tcp/TcpConnection.cs
+++ b/Source/MQTTnet.AspnetCore/Client/Tcp/TcpConnection.cs
@@ -40,7 +40,7 @@ namespace MQTTnet.AspNetCore.Client.Tcp
             _sender = new SocketSender(_socket, PipeScheduler.ThreadPool);
             _receiver = new SocketReceiver(_socket, PipeScheduler.ThreadPool);
         }
-#if NETCOREAPP3_1
+#if NETCOREAPP3_1 || NET5_0
         public override ValueTask DisposeAsync()
 #else
         public Task DisposeAsync()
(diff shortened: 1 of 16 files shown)


JanEggers:
I will have alook at it now that 5.0 is out im ager to update too :)

StefanOssendorf:
Great :-)
And just to reference the issue #1017

JanEggers:
Hi @StefanOssendorf I rebased your changes and modified some ifdefs

the current state wass building for me but some tests are failing. @chkr1011 should the tests be green or are there some expected red dots?

chkr1011:
@JanEggers On my machine all tests are green. I am still trying to fix them in our CI but it seems that the server has some limitations which end up in broken tests (CPU, RAM etc.). But they should also run all on your machine as well.

Ich tests are failing?

JanEggers:
the same as in the ci run are failing on my machine.

 Failed Connect_To_Invalid_Server_Wrong_IP [2 s]
  Error Message:
   Test method threw exception System.OperationCanceledException, but exception MQTTnet.Exceptions.MqttCommunicationTimedOutException was expected. Exception message: System.OperationCanceledException: The operation was canceled.


 Try_Connect_Invalid_Host [3 s]
  Error Message:
   Test method MQTTnet.Tests.CrossPlatformSocket_Tests.Try_Connect_Invalid_Host threw exception: 
System.Net.Sockets.SocketException: The I/O operation has been aborted because of either a thread exit or an application request.

chkr1011:
These are OK from my point of view. I will merge this soon and see how to patch the tests so they will work as usual. Still wondering why ALL tests are green on my machine!?

PR #1048, opened 2020-12-21
Only set _cleanDisconnectInitiated if no disconnect is pending

tobiasfrick:
Hi,
I found a problem in the disconnect logic of the MqttClient.

If you call `DisconnectAsync()` while the client is in 'disconnect pending' state, `_cleanDisconnectInitiated` is set to true (see [MqttClient.cs:150](https://github.com/chkr1011/MQTTnet/blob/b95d6b08bc4d8c5a7cbd164ee712ef831982de16/Source/MQTTnet/Client/MqttClient.cs#L150)).

This caused an unlucky situation in our code when the broker restarted during `MqttClient.ConnectAsync(...)`. [MqttClient:492](https://github.com/chkr1011/MQTTnet/blob/b95d6b08bc4d8c5a7cbd164ee712ef831982de16/Source/MQTTnet/Client/MqttClient.cs#L492) throws an exception because of this restart. The exception is handled in [MqttClient:514](https://github.com/chkr1011/MQTTnet/blob/b95d6b08bc4d8c5a7cbd164ee712ef831982de16/Source/MQTTnet/Client/MqttClient.cs#L514) but ignored, because `_cleanDisconnectInitiated` is still true. Now `DisconnectInternalAsync(...)` is never called, which means the client stays in 'Connected' state, even though the connection to the broker was lost.

I fixed it by returning in `DisconnectAsync()` if a disconnect is already pending, since none of the subsequent code in `DisconnectAsync()` should be executed in this case.
Also I added two missing .ConfigureAwait(false).

Can you please have a look at it and merge the changes if there are no problems?

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -142,14 +142,17 @@ namespace MQTTnet.Client
 
             ThrowIfDisposed();
 
-            var disconnectIsPending = DisconnectIsPending();
+            if (DisconnectIsPending())
+            {
+                return;
+            }
 
             try
             {
                 _disconnectReason = MqttClientDisconnectReason.NormalDisconnection;
                 _cleanDisconnectInitiated = true;
 
-                if (!disconnectIsPending && _isConnected)
+                if (_isConnected)
                 {
                     var disconnectPacket = _adapter.PacketFormatterAdapter.DataConverter.CreateDisconnectPacket(options);
                     await SendAsync(disconnectPacket, cancellationToken).ConfigureAwait(false);
@@ -157,10 +160,7 @@ namespace MQTTnet.Client
             }
             finally
             {
-                if (!disconnectIsPending)
-                {
-                    await DisconnectInternalAsync(null, null, null).ConfigureAwait(false);
-                }
+                await DisconnectInternalAsync(null, null, null).ConfigureAwait(false);
             }
         }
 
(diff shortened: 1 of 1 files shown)


chkr1011:
@tobiasfrick From my point of view this looks OK to me. But do we have a chance to reproduce this in a Unit Test so that will never be broken again? I know that hitting the exact time slot where all properties are correct is difficult but maybe we can work with some test delays?

tobiasfrick:
It's very hard to reproduce this. I was able to create a flaky test case on my machine, but it only works when debugging in Visual Studio. When running it without debugging it usually does not fail. Also it only worked when creating artificial CPU load with some background threads. I think it's very difficult to get the timing right.

I pushed it to [MqttClientTests:791](https://github.com/tobiasfrick/MQTTnet/blob/dc70b0c370c2f65bdfd11d696f19bb1933c2fe04/Tests/MQTTnet.Core.Tests/MqttClient_Tests.cs#L791) if you want to have a look at it, but probably it isn't a good idea to merge this test into master since it causes 100% CPU usage for some minutes.
In the test I created a server that restarts after a random delay when the client connects. The client tries to keep it's connection by reconnecting in a loop. Prior to reconnecting it calls `DisconnectAsync(...)` to trigger the problem with `_cleanDisconnectInitiated`.

In the master branch the test fails most of the time at line 919, because the client state is still connected even though it disconnected from the server. With the fix from the pull requests it seems to work.

If you want I can add this to the pull request, or maybe you have some ideas to create a better test case.

PR #1071, opened 2021-01-26
Allow to respond with a reason code in PUBACK/PUBREC

### LINKED ISSUE #989: Allow to respond with non-success reach PUBACK/REC
MQTT 5 allows user properties and different reason codes on PUBACK/REC but it is impossible to control response in MQTTnet Client. I suppose it is a similar situation with Server but I didn't look closely at it.

- Client
- Server

Allow to set custom response in IMqttApplicationMessageReceivedHandler  API.

### Describe alternatives you've considered
N/A

### Additional context
N/A

muneebmajeed:
To fix #989 
1. Added a new property "ReasonCode" (int?) in MqttApplicationMessageReceivedEventArgs
2. Changed the return type of HandleReceivedApplicationMessageAsync from bool to int? with default return value 0
3. If ProcessingFailed = true, HandleReceivedApplicationMessageAsync  will return return the ReasonCode from eventArgs
4. If ProcessingFailed = false or handler  == null, HandleReceivedApplicationMessageAsync  will return 0 (Success) for reason code
5. In ProcessReceivedPublishPackets, call SendAsync only if there is a valid reason code

This update doesn't change the current behavior (i.e., if the ProcessingFailed = true but the ReasonCode = null, it will not send a PUBACK/PUBREC ), but gives the user the option to respond with a reason code (i.e., if the ProcessingFailed = true but the ReasonCode contains a valid reason code, it will send the PUBACK/PUBREC with the reason code).

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -637,26 +637,26 @@ namespace MQTTnet.Client
                     }
                     else if (publishPacket.QualityOfServiceLevel == MqttQualityOfServiceLevel.AtLeastOnce)
                     {
-                        if (await HandleReceivedApplicationMessageAsync(publishPacket).ConfigureAwait(false))
+                        var reasonCode = await HandleReceivedApplicationMessageAsync(publishPacket).ConfigureAwait(false);
+                        if (reasonCode != null && Enum.IsDefined(typeof(MqttPubAckReasonCode), reasonCode.Value))
                         {
                             await SendAsync(new MqttPubAckPacket
                             {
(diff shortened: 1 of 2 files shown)


chkr1011 (approved):

chkr1011:
@muneebm Small update: I changed the PR a bit so that the reason code is type safe now.

muneebmajeed:
> @muneebm Small update: I changed the PR a bit so that the reason code is type safe now.

I was just trying to accommodate both PUBREC/PUBACK reason codes with an int. Your solution is simpler but better. Thanks!

PR #1076, opened 2021-01-29
The topic filter comparer now checks for e.g. "foo/" matching "foo/#".

### LINKED ISSUE #1067: Wrong edge case in MqttTopicFilterComparer
Hello and thank you for this great library.

### Describe the bug
I think there is a wrong edge case in the `MqttTopicFilterComparer`. The topic `a/b/` should match the filter `a/b/#`. However, according to the comparer, it does not.

- Server

### To Reproduce
- Using MQTTnet 3.0.13.
- Call `MqttTopicFilterComparer.IsMatch("a/b/", "a/b/#");`.

### Expected behavior
The function should return `true`, however, it returns `false`.

### Additional context
The topics `a/b` and `a/b/c` are correctly matched to the filter `a/b/#`, it is just the corner case with the empty level `a/b/`. I know that it is not good practice to use empty levels, but according to the MQTT specification, they are valid.

dagophil:
The topic filter comparer now checks for e.g. `foo/` matching `foo/#`.

Fixes #1067.

### DIFF
--- a/Source/MQTTnet/Server/MqttTopicFilterComparer.cs
+++ b/Source/MQTTnet/Server/MqttTopicFilterComparer.cs
@@ -31,6 +31,13 @@ namespace MQTTnet.Server
                         {
                             return true;
                         }
+                        // Check for e.g. foo/ matching foo/#
+                        if (sPos == sLen - 2
+                                && filter[sPos] == LevelSeparator
+                                && filter[sPos + 1] == MultiLevelWildcard)
+                        {
+                            return true;
+                        }
                     }
 
                     sPos++;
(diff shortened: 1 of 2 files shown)


SeppPenner:
Thank you, this look good :)

dagophil:
Np, thank you for the quick review and merge :)

chkr1011:
@SeppPenner Please also update the nuget description and add a line describing this change including the mention of the creator.

SeppPenner:
> @SeppPenner Please also update the nuget description and add a line describing this change including the mention of the creator.

Sorry, I always forget that 😅

SeppPenner:
Is done with https://github.com/chkr1011/MQTTnet/commit/63ca6086d5e95de4930dc9211c597fee77e1542c.

Issue #1082, opened 2021-02-04
Regression: No exception on ConnectAsync(null, password)  [MQTT-3.1.2-22]


milbrandt:
### Describe the bug
According to the standard [3.1.2.9 Password Flag](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349236) MQTTnet 3.0.8 throw an MqttProtocolViolationException when a client tried to connect the server with `null` as username but a non-null password.

Since 3.0.9 this exception is no longer the result of a `MQTTnet.Client.MqttClient.ConnectAsync`.

### Which project is your bug related to?
- Client
- Server
- Generic

### To Reproduce
To reproduce the regression I introduced unit tests. 
1. Using this version of MQTTnet '3.0.8' -  https://github.com/milbrandt/MQTTnet/commit/57652c76cfb4aeb1af457eea65c93d816504831a - all unit tests are passing
2. Using current master of MQTTnet '3.0.14' (already in '3.0.9' )- 
 https://github.com/milbrandt/MQTTnet/commit/e342a51d5026e8a0ffa6c5783d6de0696f2b244f - the unit test 'Use_Username_Null_Password_Empty' is failing

### Expected behavior

As in '3.0.8' an 'MqttCommunicationException' with inner 'MqttProtocolViolationException' would be expected. The exception message was `If the User Name Flag is set to 0, the Password Flag MUST be set to 0 [MQTT-3.1.2-22].` and directing to the MQTT standard violation.

### Screenshots
~~~
 Use_Username_Null_Password_Empty
   Source: Server_Tests.cs line 54
   Duration: 292 ms

  Message: 
    Assert.ThrowsException failed. Threw exception TaskCanceledException, but exception MqttCommunicationException was expected. 
    Exception Message: A task was canceled.
    Stack Trace:    at MQTTnet.PacketDispatcher.MqttPacketAwaiter`1.WaitOneAsync(TimeSpan timeout) in D:\_Repos\MQTTnet\Source\MQTTnet\PacketDispatcher\MqttPacketAwaiter.cs:line 32
       at MQTTnet.Client.MqttClient.SendAndReceiveAsync[TResponsePacket](MqttBasePacket requestPacket, CancellationToken cancellationToken) in D:\_Repos\MQTTnet\Source\MQTTnet\Client\MqttClient.cs:line 415
[...]

PR #1083, opened 2021-02-04
Test for Regression: No exception on ConnectAsync(null, password) [MQTT-3.1.2-22]

milbrandt:
In #1082 I described that in 3.0.8 a MqttProtocolViolationException was thrown on `ConnectAsync(null, password)` which seems to match the MQTT standard.
Since 3.0.9 this exception can no longer be catched, the proposed test is canceled by the testing framework.

Here are unit test for all four combinations null/non-null of (username, password) .

- initial commit based on 3.0.8: tests are all passing
- second commit (based on current master, but also already on 3.0.9): test `Use_Username_Null_Password_Empty` is failing


chkr1011:
Thanks for adding the Unit Tests. I will check why this is broken after 3.0.8.

chkr1011 (approved):

PR #1094, opened 2021-02-16
Fix execute rpc async

SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/1093.

### DIFF
--- a/Source/MQTTnet.Extensions.Rpc/RpcAwareApplicationMessageReceivedHandler.cs
+++ b/Source/MQTTnet.Extensions.Rpc/RpcAwareApplicationMessageReceivedHandler.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading.Tasks;
 using MQTTnet.Client.Receiving;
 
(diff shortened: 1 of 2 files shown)


Source/MQTTnet.Extensions.Rpc/RpcAwareApplicationMessageReceivedHandler.cs
```diff
@@ -20,11 +20,7 @@ public sealed class RpcAwareApplicationMessageReceivedHandler : IMqttApplication
 
         public Task HandleApplicationMessageReceivedAsync(MqttApplicationMessageReceivedEventArgs eventArgs)
         {
-            if (OriginalHandler != null)
-            {
-                return OriginalHandler.HandleApplicationMessageReceivedAsync(eventArgs);
-            }
-
+            OriginalHandler?.HandleApplicationMessageReceivedAsync(eventArgs);
```
borigas, line 23: I think something should be awaiting this task, not just discarding. As it sits, this method can return before `OriginalHandler?.HandleApplicationMessageReceivedAsync(eventArgs)` completes, which I don't think is intended. Exceptions from it are also discarded. Here's a code analysis rule that talk more about it. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs4014

```
public async Task HandleApplicationMessageReceivedAsync(MqttApplicationMessageReceivedEventArgs eventArgs)
{
    if (OriginalHandler != null)
    {
        await OriginalHandler.HandleApplicationMessageReceivedAsync(eventArgs).ConfigureAwait(false);
    }
    await _handleReceivedApplicationMessageAsync(eventArgs).ConfigureAwait(false);
}
```

SeppPenner:
Closed as this is already done in the master.

PR #1097, opened 2021-02-18
Update nuspecs

yyjdelete:
Manually sync with the file generated by `msbuild /t:pack` or `dotnet pack`, and remove some unnessary dependencies

### DIFF
--- a/Build/MQTTnet.AspNetCore.nuspec
+++ b/Build/MQTTnet.AspNetCore.nuspec
@@ -28,6 +28,14 @@
         <dependency id="MQTTnet" version="$nugetVersion" />
       </group>
     </dependencies>
+    <frameworkReferences>
+      <group targetFramework=".NETCoreApp3.1">
+        <frameworkReference name="Microsoft.AspNetCore.App" />
+      </group>
+      <group targetFramework="net5.0">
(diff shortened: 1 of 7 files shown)


Build/MQTTnet.AspNetCore.nuspec
```diff
@@ -28,6 +28,14 @@
         <dependency id="MQTTnet" version="$nugetVersion" />
       </group>
     </dependencies>
+    <frameworkReferences>
+      <group targetFramework=".NETCoreApp3.1">
```
chkr1011, line 32: @JanEggers Could you please verify that this is correct?

chkr1011 (changes requested):

Build/MQTTnet.AspNetCore.nuspec
JanEggers: I dont know whats the benefit of depending on a frameworkreference. it worked without them. my preference would be to remove the nuspec altogether and just use the csproj to build the packages. is there any info in the nuspec that cannot be provided by csproj?

chkr1011: For now I want to keep the nuspec due to the build script and release notes etc. etc. But for the future we can surely think about the generated packages from VS. So you say we can remove these references?

Issue #1110, opened 2021-03-04
Get SessionItems with GetSessionStatusAsync()


zaksnet:
### Describe the feature request
Hey guys!
I am currently in the process of moving from `SignalR` to `MQTTNet` because i had some issues with `Mono`. In `SignalR` i was using a custom `ConnectionManager` class to store additional information about the sessions like logged in users (since logging in happens after MQTT has initialized connection). My use case is:
- When a user tries to login i have to search the open connections and see if that user is logged in somewhere else.
My initial though was that i can save the `ConnectedUser` inside the `SessionItems` dictionary and then iterate through `SessionItems` when i get the list of clients but it seems that `GetSessionStatusAsync()` does not return the `SessionItems` for each connection.
Is there any other way to get the `SessionItems` and then iterate through them to see if the user is logged in somewhere else or should i use a custom dictionary to store such data? Otherwise, what is the use of SessionItems? Unfortunately I could not find any useful info on that when googling.

### Which project is your feature request related to?

- Server

### Describe the solution you'd like
Be able to get `SessionItems` for each connection when using `GetSessionStatusAsync()`

### Describe alternatives you've considered
Using an external `Dictionary<string, string> _connectedUsers` that stores the `ClientId` as the key and the username as a value and then iterate through that list to figure out if the user is connected. The problem with this approach is that it does not scale. If we need to use a cluster to manage the server then i would also need to somehow share the `_connectedUsers` dictionary among the servers.

Thank you!

zaksnet:
It seems that `SessionItems` is just named `Items` when you retrieve the `IMqttSessionStatus`. I'm not sure if that is a design decision but it is kind of confusing so maybe you should consider renaming it (i will make a pull request on this and you can either accept it or reject it) . In any case, my issue is fixed!

PR #1111, opened 2021-03-04
Rename Items to SessionItems.

zaksnet:
See #1110 for more info.

### DIFF
--- a/Source/MQTTnet/Server/MqttClientSession.cs
+++ b/Source/MQTTnet/Server/MqttClientSession.cs
@@ -86,7 +86,7 @@ namespace MQTTnet.Server
             status.ClientId = ClientId;
             status.CreatedTimestamp = _createdTimestamp;
             status.PendingApplicationMessagesCount = ApplicationMessagesQueue.Count;
-            status.Items = Items;
+            status.SessionItems = Items;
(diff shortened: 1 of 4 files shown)


chkr1011:
I do not agree with this change. The entire library has a naming scheme which tries to avoid duplicates in naming where possible. So with you change we will get "MqttSessionStatus.SessionItems". It is clear that these items are the session items because the belong to the "Session status" class. If we follow your change we also have to rename other properties like "CreatedTimestamp" to "SessionCreatedTimestamp" or "PendingApplicationMessagesCount" will become "PendingSessionApplicationMessagesCount" which has no benefit in my opinion.

zaksnet:
> I do not agree with this change. The entire library has a naming scheme which tries to avoid duplicates in naming where possible. So with you change we will get "MqttSessionStatus.SessionItems". It is clear that these items are the session items because the belong to the "Session status" class. If we follow your change we also have to rename other properties like "CreatedTimestamp" to "SessionCreatedTimestamp" or "PendingApplicationMessagesCount" will become "PendingSessionApplicationMessagesCount" which has no benefit in my opinion.

Alright, you'v got a good point there.

PR #1121, opened 2021-03-11
Added workflow to  build debian/ubuntu packages

asymetrixs:
This workflow generates MQTTnet.Server as debian package which can be easily installed on Ubuntu.
It puts a systemd service file in place to manage MQTTnet.Server as a systemd service.
It witelists the MQTTnet.Server executable to open ports on local IP interfaces.
Copyright of course can be adjusted.

A problem with Version and File Version exists, right now the workflow requests these values from the user executing the workflow. This should be improved. If it is the same for both, it could be taken from Tags if available.

### DIFF
--- a/.github/workflows/deb-packaging.yml
+++ b/.github/workflows/deb-packaging.yml
@@ -1,0 +1,67 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Deb Packaging
+
+# Controls when the action will run. 
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      FileVersion:
+        description: 'File Version'
+        required: true
+        default: ''
+      Version:
+        description: 'Version'
+        required: true
+        default: ''
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build MQTTnetServer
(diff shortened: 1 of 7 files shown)


asymetrixs:
The version is now extracted from the MQTTNet.Server.csproj.
During compilation, Version is not set via /p:Version, it should be taken automatically from the csproj-file.

chkr1011:
Good work. So I can download DEB packages now and add them to the release artifacts right? 
Please let me know as soon as you are ready because I sometimes see commits and don't know if this PR is done.

asymetrixs:
@chkr1011 Hi, yes that is the idea. I tested it, it gives you an artefact (zip) that contains the .deb - I am not exactly sure regarding the 'Version', as described, taking it from the .csproj now. This PR is done.

chkr1011:
The problem with the version is that it is set in the build script only. So maybe you don't have it available in the workflow. Is there a way to specify it manually? Or are we able to start the workflow with parameters from the build script? Otherwise I have to set the version in the CSProj file. But this would be OK for me.

asymetrixs:
I would recommend putting the Version into the respective field in the project file. Then also during compilation the correct version is written into the assembly. Otherwise, the script could take it from somewhere else in the sources or the workflow would ask for the version number and then you'd had to enter it every time you generate a .DEB package. I do not like that solution, whoever starts the workflow can put whatever he likes (more or less). I'd prefer taking it from the .csproj or another file in the sources, or maybe a TAG. Then it needs to be specified what is being used when no TAG is present.

chkr1011:
I will merge this for now. But we should review the version stuff. Maybe we can add a file in the repo containing the final version for nuget etc. as well.

chkr1011 (approved):

PR #1134, opened 2021-04-01
Disallow to call MqttClient.ConnectAsync while Disconnect is pending.

### LINKED ISSUE #996: Possbile concurrent issue between `ManagedMqttClient.ReconnectIfRequiredAsync()` and `MqttClient.DisconnectInternalAsync()`?
### Describe the bug
Failed to make an reproduce, but I get an MqttClient with `{ IsDisposed=false, _isConnected=true, _backgroundCancellationTokenSource=null,_isDisconnectPending=0, _publishPacketReceiverQueue=null, _keepAlivePacketsSenderTask=canceled, _packetReceiverTask =RanToCompletion, _publishPacketReceiverTask=canceled}` with ManagedMqttClient, and reconnect not work in this state.

Is it possbile concurrent issue between ManagedMqttClient.ReconnectIfRequiredAsync() and `MqttClient.DisconnectInternalAsync()`, if `ReconnectIfRequiredAsync()` is called while `DisconnectInternalAsync()` is still running(IsConnected is already false), and get an MqttClient in conflict state?

- Client : 3.0.12
- ManagedClient : 3.0.12

### To Reproduce
Failed to make an reproduce.

### Expected behavior
Auto reconnected should be handle by ManagedMqttClient after network issue.

### LINKED ISSUE #1010: ManagedClient 3.0.12 will never reconnect,then thread will grow
Line 18048: 2020-10-24 12:48:43.581 +08:00 [INF] mqttNetLog:LogId-,Message-TX (283 bytes) >>> Publish: [Topic= test/gate/up/data] [Payload.Length=259] [QoSLevel=AtMostOnce] [Dup=False] [Retain=False] [PacketIdentifier=],Source-MqttChannelAdapter,Exception:
	Line 18080: 2020-10-24 12:49:14.078 +08:00 [INF] mqttNetLog:LogId-,Message-TX (2 bytes) >>> PingReq,Source-MqttChannelAdapter,Exception:
	Line 18081: 2020-10-24 12:49:14.101 +08:00 [INF] mqttNetLog:LogId-,Message-RX (2 bytes) <<< PingResp,Source-MqttChannelAdapter,Exception:
	Line 18119: 2020-10-24 12:50:07.717 +08:00 [INF] mqttNetLog:LogId-client,Message-Disconnecting [Timeout=00:01:00],Source-MqttClient,Exception:
	Line 18120: 2020-10-24 12:50:07.720 +08:00 [INF] mqttNetLog:LogId-client,Message-Disconnected from adapter.,Source-MqttClient,Exception:
	Line 18157: 2020-10-24 12:50:51.581 +08:00 [INF] mqttNetLog:LogId-client,Message-Trying to connect with server '58.56.66.163:3989' (Timeout=00:01:00).,Source-MqttClient,Exception:
	Line 18159: 2020-10-24 12:50:51.613 +08:00 [INF] mqttNetLog:LogId-client,Message-Connection with server established.,Source-MqttClient,Exception:
	Line 18160: 2020-10-24 12:50:51.613 +08:00 [INF] mqttNetLog:LogId-,Message-TX (76 bytes) >>> Connect: [ClientId=310c37f3-d880-4950-9f59-3b27b832ee9e] [Username=sensorGate1] [Password=****] [KeepAlivePeriod=30] [CleanSession=True],Source-MqttChannelAdapter,Exception:
	Line 18195: 2020-10-24 12:51:35.187 +08:00 [INF] mqttNetLog:LogId-client,Message-Disconnecting [Timeout=00:01:00],Source-MqttClient,Exception:
	Line 18196: 2020-10-24 12:51:35.187 +08:00 [INF] mqttNetLog:LogId-client,Message-Disconnected from adapter.,Source-MqttClient,Exception:

yyjdelete:
Fix #996
Fix #1010

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -39,10 +39,15 @@ namespace MQTTnet.Client
         IMqttChannelAdapter _adapter;
         bool _cleanDisconnectInitiated;
         volatile int _connectState;
-        const int CONNECT_STATED_DISCONNECTED = 0;
-        const int CONNECT_STATED_DISCONNECTING = 1;
-        const int CONNECT_STATED_CONNECTED = 2;
-        const int CONNECT_STATED_CONNECTING = 3;
+
+        enum ConnectState
+        {
+            Disconnected = 0,
+            Disconnecting,
+            Connected,
+            Connecting
+        }
+
         MqttClientDisconnectReason _disconnectReason;
 
         DateTime _lastPacketSentTimestamp;
@@ -61,7 +66,7 @@ namespace MQTTnet.Client
 
         public IMqttApplicationMessageReceivedHandler ApplicationMessageReceivedHandler { get; set; }
 
-        public bool IsConnected => _connectState == CONNECT_STATED_CONNECTED;
+        public bool IsConnected => (ConnectState)_connectState == ConnectState.Connected;
 
         public IMqttClientOptions Options { get; private set; }
 
@@ -74,7 +79,7 @@ namespace MQTTnet.Client
 
             ThrowIfDisposed();
 
-            if (Interlocked.CompareExchange(ref _connectState, CONNECT_STATED_CONNECTING, CONNECT_STATED_DISCONNECTED) != CONNECT_STATED_DISCONNECTED)
+            if (CompareExchangeConnectState(ConnectState.Connecting, ConnectState.Disconnected) != ConnectState.Disconnected)
                 throw new InvalidOperationException("Not allowed to connect while connect/disconnect is pending.");
 
             MqttClientAuthenticateResult authenticateResult = null;
@@ -113,7 +118,7 @@ namespace MQTTnet.Client
                     _keepAlivePacketsSenderTask = Task.Run(() => TrySendKeepAliveMessagesAsync(backgroundCancellationToken), backgroundCancellationToken);
                 }
 
-                Interlocked.CompareExchange(ref _connectState, CONNECT_STATED_CONNECTED, CONNECT_STATED_CONNECTING);
+                CompareExchangeConnectState(ConnectState.Connected, ConnectState.Connecting);
 
                 _logger.Info("Connected.");
 
@@ -365,7 +370,7 @@ namespace MQTTnet.Client
             {
                 Cleanup();
                 _cleanDisconnectInitiated = false;
-                Interlocked.CompareExchange(ref _connectState, CONNECT_STATED_DISCONNECTED, CONNECT_STATED_DISCONNECTING);
+                CompareExchangeConnectState(ConnectState.Disconnected, ConnectState.Disconnecting);
 
                 _logger.Info("Disconnected.");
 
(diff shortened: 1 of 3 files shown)


yyjdelete:
Call MqttClient.ConnectAsync while Disconnect is pending may lead MqttClient to an broken state(disconnect free the resources alloc with the new connection instead of the old one), and block the usage of `ManagedMqttClient`(AutoReconnect may broken the client, and user has no way to check that and restore from the broken state)

An quick testcase can be found at https://github.com/chkr1011/MQTTnet/issues/996#issuecomment-699650850

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -38,8 +38,11 @@ public class MqttClient : Disposable, IMqttClient
 
         IMqttChannelAdapter _adapter;
         bool _cleanDisconnectInitiated;
-        long _isDisconnectPending;
-        bool _isConnected;
+        volatile int _connectState;
+        const int CONNECT_STATED_DISCONNECTED = 0;
```
chkr1011, line 42: Please convert this to an enum with PascalCase style because this common in C#. You can also add a private method as a replacement of the _Interlocked.CompareExchange_ calls which will also accept these enums only. Internally the new method will cast the num as integers to that we use the enum only in business code and have the technical conversion and thread safe access stuff separated.

chkr1011 (changes requested):

Source/MQTTnet/Server/MqttClientSessionsManager.cs
```diff
@@ -425,10 +426,13 @@ MqttClientConnection CreateClientConnection(MqttConnectPacket connectPacket, Mqt
                     _connections[connectPacket.ClientId] = connection;
                 }
 
-                existingConnection?.StopAsync(MqttClientDisconnectReason.SessionTakenOver).GetAwaiter().GetResult();
-
-                return connection;
+                if (existingConnection != null)
+                {
+                    await existingConnection.StopAsync(MqttClientDisconnectReason.SessionTakenOver).ConfigureAwait(false);
```
yyjdelete, line 431: Changes in this file is not required, but `Session_Tests.Manage_Session_MaxParallel()` may failed, because too many threads are blocked by `GetResult()` here when change https://github.com/chkr1011/MQTTnet/blob/7eb6d2d1e3d0480b89cbdafe0dc4380ea5dd055c/Tests/MQTTnet.Core.Tests/Session_Tests.cs#L101 from 10 to 50.

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -38,8 +38,16 @@ public class MqttClient : Disposable, IMqttClient
 
         IMqttChannelAdapter _adapter;
         bool _cleanDisconnectInitiated;
-        long _isDisconnectPending;
-        bool _isConnected;
+        volatile int _connectState;
+
+        enum ConnectState
```
chkr1011, line 43: Please move this to a dedicated c# file. In this project we do not have any private classes or enums etc. embedded in other classes. And consider renaming this to "MqttClientConnectionState" which follows the naming pattern in this project.

chkr1011 (changes requested):

Source/MQTTnet/Client/MqttClient.cs
yyjdelete: I can do this tomorrow, but the name is really easy to be confused with `MQTTnet.Server.MqttClientConnectionStatus`

chkr1011: You are right. Maybe we need to use the same name but due to a different namespace it is clear that the scope is a different one?

yyjdelete: Moved to `MQTTnet.Client.MqttClientConnectionStatus` and updated related field/method names.

chkr1011:

PR #1135, opened 2021-04-02
Add more checks for NET5_0

yyjdelete:


yyjdelete:
Maybe it's better to use `NETCOREAPP3_1_OR_GREATER` in the future(when support for net6.0 is added), but it require dotnet-sdk `>=5.0.200`(and maybe vs16.9?) to work.


See 
https://github.com/dotnet/designs/blob/main/accepted/2020/or-greater-defines/or-greater-defines.md#goals
https://github.com/dotnet/sdk/commit/c1f44b141df60245ed25a1148f3516cc971140a3
for details

And for `NETSTANDARD2_0_OR_GREATER`, 5.0.300
https://github.com/dotnet/sdk/pull/15660/commits/bed6f101704969d7893a5281af089b8601491d83

chkr1011:
Thank you for your contribution. Good work. Maybe we can switch to the pragma new strategy when 6.0 is final? The CI is updated only once in a while (when using the free tier). So it may have an old SDK.

Issue #1103, opened 2021-02-27
MqttServerOptionsBuilder.WithDisconnectedInterceptor NOT work


egops:
### Describe the bug
I started a server with disconnected interceptor

the DisconnectedInterceptor not called when i powered off my mqtt client

### Which project is your bug related to?

- Server

### To Reproduce
1. Using this version of MQTTnet '3.1.4'.
2. Run this code:
```csharp
options .WithConnectionValidator(c =>
 {
    // works well
    logger.LogInformation($"{c.ClientId} is connected");
});
options.WithDisconnectedInterceptor(c =>
{
    // but this not
    logger.LogWarning($"{c.ClientId} is disconnected: [{c.DisconnectType}] ");
});``

### Expected behavior
The interceptor should be called when client disconnected or keepalive timeout;

**Comment by egops** [comment]
I find IMqttServerOptions.ClientDisconnectedInterceptor not been used in source, it should be assign to MqttServerEventDispatcher.IMqttServerClientDisconnectedHandler.

**Comment by chkr1011** [comment]
The method is used in method _CleanUpClient_. Please try again with the latest verison.

**Comment by egops** [comment]
still not work.

**Comment by jtotht** [comment]
Neither for me. I created a [minimal repro](https://github.com/chkr1011/MQTTnet/files/6201185/MQTTnetDisconnectedRepro.zip) to demonstrate the issue: start the server, then run the client. Notice that `Validator fired` appears on stdout of the server, but `Interceptor fired` doesn’t.

**Comment by Octoate** [comment]
Ah, okay. I just thought that this is a problem of my implementation :-). I currently have the same problem, too.

**Comment by egops** [comment]
I found a solution temporary
Don't use the nuget package, clone the source to project and modify the code like this:

MqttClientSubscriptionsManager.cs   in constructor
```csharp
        public MqttClientSubscriptionsManager(MqttClientSession clientSession, MqttServerEventDispatcher eventDispatcher, IMqttServerOptions serverOptions)
        {
            _clientSession = clientSession ?? throw new ArgumentNullException(nameof(clientSession));
            _options = serverOptions ?? throw new ArgumentNullException(nameof(serverOptions));
            _eventDispatcher = eventDispatcher ?? throw new ArgumentNullException(nameof(eventDispatcher));

            // ADD this line 
            _eventDispatcher.ClientDisconnectedHandler = _options.ClientDisconnectedInterceptor;
        }
```
 
have fun!

[...]

PR #1141, opened 2021-04-10
fix issue #1103: MqttServerOptionsBuilder.WithDisconnectedInterceptor…

egops:
I'm not sure if this is a best place to place this code, but it works.


chkr1011:
This is not the right place. The problem is that this handler ist also declared at the server class and should be set there instead. I will remove the handlers from the options because this is the wrong place.

chkr1011:
Please check my change in master.

PR #1209, opened 2021-08-02
Added option for how many subscriptions to send in a single call to t…

scottbrogden-iheartmedia:
Added option for how many subscriptions to send in a single call to the server.  aws iotcore limits this to 8

### DIFF
--- a/Source/MQTTnet.Extensions.ManagedClient/IManagedMqttClientOptions.cs
+++ b/Source/MQTTnet.Extensions.ManagedClient/IManagedMqttClientOptions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using MQTTnet.Client.Options;
 using MQTTnet.Server;
 
@@ -17,5 +17,7 @@ namespace MQTTnet.Extensions.ManagedClient
         int MaxPendingMessages { get; }
 
         MqttPendingMessagesOverflowStrategy PendingMessagesOverflowStrategy { get; }
+
+        int? MaxSubcribeUnsubscribeMessagesAtOnce { get; }
     }
 }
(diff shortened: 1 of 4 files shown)


Source/MQTTnet.Extensions.ManagedClient/IManagedMqttClientOptions.cs
```diff
@@ -17,5 +17,7 @@ public interface IManagedMqttClientOptions
         int MaxPendingMessages { get; }
 
         MqttPendingMessagesOverflowStrategy PendingMessagesOverflowStrategy { get; }
+
+        int? MaxSubcribeUnsubscribeMessagesAtOnce { get; }
```
chkr1011, line 21: Please start with a value of int.Max. This avoids null checks and boxing.

Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
```diff
@@ -527,7 +558,19 @@ async Task PublishReconnectSubscriptionsAsync()
                 if (_reconnectSubscriptions.Any())
                 {
                     var subscriptions = _reconnectSubscriptions.Select(i => new MqttTopicFilter { Topic = i.Key, QualityOfServiceLevel = i.Value });
-                    await InternalClient.SubscribeAsync(subscriptions.ToArray()).ConfigureAwait(false);
+                    List<MqttTopicFilter> subs = new List<MqttTopicFilter>();
+                    foreach (var sub in subscriptions)
+                    {
+                        subs.Add(sub);
+                        //aws only allows 8 in a single message
```
chkr1011, line 565: Please move this comment to a real summary comment at property level. So all users can read that without accessing the source code.

chkr1011 (changes requested):

chkr1011:
I will merge this feature and do the changes on my own so that we have this feature available. @SeppPenner Do you want to add the notes to the wiki (only 8 topic filters are supported in AWS per subscription)? I will add this in code summary.

SeppPenner:
I have just added a small text to https://github.com/chkr1011/MQTTnet/wiki/Tips-and-Tricks#limits-for-multi-subscriptions-per-single-call. @chkr1011

Issue #1217, opened 2021-08-20
Only 2 of 10 public broker tests work when running MQTTnet.TestApp.NetCore


hakakou:
I was trying to debug why I kept getting the exception MQTTnet.Exceptions.MqttCommunicationTimedOutException when testing a connection to test.mosquitto.org.

It seems to me, even the MQTTnet.TestApp.NetCore does not run correctly. Only 2 out of 10 public brokers work.  Tested in both Windows Server 2016 and Windows 10

### To Reproduce
1. Using this version of MQTTnet: 3.0.16
2. Run MQTTnet.TestApp.NetCore console app from commit d1c883e (Jun 29 2021)
3. Click "5"

### Screenshots
![image](https://user-images.githubusercontent.com/10587642/130267943-2b78d880-944a-4f6a-b68f-34a0b0bb4cf1.png)

SeppPenner:
Hi, can you give me the list of brokers you tested? I will try to find the issue(s) if possible.

hakakou:
I made a pull request with some of my own fixes. On my computer now I have this result:
![image](https://user-images.githubusercontent.com/10587642/131319617-d02406f8-9664-4975-95bb-1cd2261120eb.png)

SeppPenner:
I guess, `broker.emqx.io` does not allow TLS1.3 (Can you test with 1.2 here?). For the WebSockets under `test.mosquitto.org`, I have no idea :D

For me, the pull request looks good. @chkr1011 Can you check https://github.com/chkr1011/MQTTnet/pull/1223 as well?

hakakou:
OK I just make another commit with some more lines, that actually test both cases - TLS12 and TLS13 for clarity. These are my result now.

![image](https://user-images.githubusercontent.com/10587642/131326705-c87773b1-8bbf-48c3-9301-6eabd0196813.png)

About broker.emqx.io, indeed they don't seem to support TLS13, however using TLS12 gives a different error.

SeppPenner:
> "Cannot determine the frame size or a corrupted frame was received."

Mhm, interesting. I have no idea here. @chkr1011 or @JanEggers Can you maybe have a look?

chkr1011:
This is related to an TLS issue which we cannot fix because we use the implementation from .NET.

SeppPenner:
> 
> 
> This is related to an TLS issue which we cannot fix because we use the implementation from .NET.

[...]

PR #1223, opened 2021-08-30
Updated tests with public servers.

hakakou:
For the project MQTTnet.TestApp.NetCore, updated the file PublicBrokerTest.cs to add some public test servers and remove older, inactive servers. 

The MQTTnet.TestApp.NetCore project should be compiled using Net5.0. Using any previous version, most connections fail.

This is a contribution to issue #1217


chkr1011 (approved):

PR #1270, opened 2021-10-20
test ResultCode for invalid credentials (no userName but password)

milbrandt:
Test for the result code if user == null, password != null (see #1082)

Additional changed order of parameters in assert to match expected order in `AreEqual<T>(T expected, T actual)`


chkr1011:

PR #1277, opened 2021-10-28
Fixes #1276 timestamp check in server keep-alive monitor

### LINKED ISSUE #1276: Client publishing QoS 0 messages is forcefully disconnected
### Describe the bug
A client that repeatedly publishes QoS 0 messages within the keep-alive interval is disconnected by the server.

I have noticed this when running the benchmark test publishing 10000 messages. Sometimes the benchmark fails to complete because the client gets disconnected.

Will follow up with a pull request shortly.

- Server

### To Reproduce
1. Connect a client to the server.
2. Repeat sending Publish requests with QoS 0.
3. Shortly after the keep-alive period has expired, the server disconnects the client.

### Expected behavior
The server should not disconnect the client because the server continuously receives Publish packets.

The following test re-produces the issue.
```csharp
  [TestMethod]
  public async Task Publish_QoS_0_Over_Period_Exceeding_KeepAlive()
  {
      using (var testEnvironment = new TestEnvironment(TestContext))
      {
          const int KeepAlivePeriodSecs = 3;

          var options = new MqttClientOptionsBuilder().WithKeepAlivePeriod(TimeSpan.FromSeconds(KeepAlivePeriodSecs));
          var client = await testEnvironment.ConnectClient(options);
          var message = new MqttApplicationMessageBuilder().WithTopic("a").Build();

          try
          {
              // Publish messages over a time period exceeding the keep alive period.
              // This should not cause an exception because of, i.e. "Client disconnected".

              for (var count = 0; count < KeepAlivePeriodSecs * 3; ++count)
              {
                  // Send Publish requests well before the keep alive period expires
                  await client.PublishAsync(message);
                  await Task.Delay(1000);
              }
          }
          catch (Exception ex)
          {
              Assert.Fail(ex.Message);
          }
      }
  }
```
Output:

```batch
Publish_QoS_0_Over_Period_Exceeding_KeepAlive
   Source: MqttClient_Tests.cs line 608
   Duration: 5.1 sec

  Message: 
Assert.Fail failed. The client is not connected.

  Stack Trace: 
Client_Tests.Publish_QoS_0_Over_Period_Exceeding_KeepAlive() line 634
ThreadOperations.ExecuteWithAbortSafety(Action action)
```
Pull request: #1277

logicaloud:
A client that publishes QoS 0 messages will not receive acknowledgement packages from the server, leaving the _"last packet sent by the server timestamp"_ unchanged. The `MqttServerKeepAliveMonitor` was checking this _"last packet sent by the server timestamp"_ to determine whether the client is still alive, instead of checking against a _"last packet received from the client timestamp"_.

This pull request changes ambiguous naming in the `MqttClientConnectionStatistics` class and then uses the `LastPacketReceivedFromClientTimestamp` in the  `MqttServerKeepAliveMonitor` class to determine whether the keep-alive period has been exceeded.

The added test confirms that the client can keep publishing QoS 0 messages beyond the keep-alive period without being disconnected.

Side note: The naming in the `MqttClientStatus` is still ambiguous. It seems unused but could become a pitfall in the future.

Fixes issue # #1276

### DIFF
--- a/Source/MQTTnet/Server/Internal/MqttClientConnectionStatistics.cs
+++ b/Source/MQTTnet/Server/Internal/MqttClientConnectionStatistics.cs
@@ -32,12 +32,15 @@ namespace MQTTnet.Server.Internal
             _lastNonKeepAlivePacketReceivedTimestamp = _connectedTimestamp;
         }
 
-        // This class is tracking all values from the Client's perspective regarding the internal "received"/"sent" timestamps.
-        // "Sent" means, a package has been received from the client.
-        // "Received" means, a package has been sent by the server.
-        // The public property below makes it less ambiguous.
+        /// <summary>
+        /// Timestamp of the last package that has been received from the client ("sent" from the client's perspective)
+        /// </summary>
+        public DateTime LastPacketSentTimestamp => _lastPacketSentTimestamp;
 
-        public DateTime LastPacketReceivedFromClientTimestamp => _lastPacketSentTimestamp;
+        /// <summary>
+        /// Timestamp of the last package that has been sent by the client ("received" from the client's perspective)
+        /// </summary>
+        public DateTime LastPacketReceivedTimestamp => _lastPacketReceivedTimestamp;
 
 
         public void HandleReceivedPacket(MqttBasePacket packet)
--- a/Source/MQTTnet/Server/Internal/MqttClientConnectionStatistics.cs
+++ b/Source/MQTTnet/Server/Internal/MqttClientConnectionStatistics.cs
@@ -38,7 +38,7 @@ namespace MQTTnet.Server.Internal
         public DateTime LastPacketSentTimestamp => _lastPacketSentTimestamp;
 
         /// <summary>
-        /// Timestamp of the last package that has been sent by the client ("received" from the client's perspective)
+        /// Timestamp of the last package that has been sent to the client ("received" from the client's perspective)
         /// </summary>
         public DateTime LastPacketReceivedTimestamp => _lastPacketReceivedTimestamp;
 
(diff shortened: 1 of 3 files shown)


Source/MQTTnet/Server/Internal/MqttClientConnectionStatistics.cs
```diff
@@ -32,7 +32,13 @@ public MqttClientConnectionStatistics()
             _lastNonKeepAlivePacketReceivedTimestamp = _connectedTimestamp;
         }
 
-        public DateTime LastPacketReceivedTimestamp => _lastPacketReceivedTimestamp;
+        // This class is tracking all values from the Client's perspective regarding the internal "received"/"sent" timestamps.
+        // "Sent" means, a package has been received from the client.
+        // "Received" means, a package has been sent by the server.
+        // The public property below makes it less ambiguous.
+
+        public DateTime LastPacketReceivedFromClientTimestamp => _lastPacketSentTimestamp;
```
chkr1011, line 40: Please do not rename the property for now. This will cause breaking changes. The comment can be wrapped into a summary tag so that every user can read this explanation.

chkr1011 (changes requested):

Source/MQTTnet/Server/Internal/MqttClientConnectionStatistics.cs
logicaloud: True. I have modified the code to expose the additional required `LastPacketSentTimestamp` for use in the keep alive monitor and added "Summary" comments to both, the "Sent" and "Received" properties. Please have a look.

chkr1011:

PR #1282, opened 2021-11-03
Add message throughput test to NetCore Test App

logicaloud:
I'd like to look at improving performance for the somewhat typical use case where there are many publishers and topics but only few subscribers. Currently, the server slows down quite a bit as the number of publishers and topics increase.

This pull request merely adds a "Message Throughput Test" to the NetCore App as the basis for comparison. The test creates 1 subscriber client and 5000 publishers with 10 topics each (50000 topics total), generates subscriptions for no-wildcard topics, single-wildcard topics and multi-wildcard topics and measures the time it takes to publish and receive messages. The number of publishers and topics can be adjusted to suit but require a rebuild with the constants changed. This could be a bit more flexible in the future.

In the coming days I'd also like to propose some performance improvements if that is of interest? For the "5000 publishers / 50000 topics" case the achievable speed-up seems to be about 50x, and for "100 publishers / 1000 topics" it is still 3x or so (but then the server is already relatively fast). @chkr1011, I'm aware that there is other work going on, but maybe it would be helpful to add the proposal as a pull request now even if any potential merge occurs much later?

Here is some test output for the added test.

mqttnet-msg-throughput-test

### DIFF
--- a/Tests/MQTTnet.TestApp.NetCore/MessageThroughputTest.cs
+++ b/Tests/MQTTnet.TestApp.NetCore/MessageThroughputTest.cs
@@ -1,0 +1,341 @@
+using MQTTnet.Client;
+using MQTTnet.Client.Options;
+using MQTTnet.Server;
+using System.Collections.Generic;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Linq;
+
+namespace MQTTnet.TestApp.NetCore
+{
+    /// <summary>
+    /// Connect a number of publisher clients and one subscriber client, then publish messages and measure
+    /// the number of messages per second that can be exchanged between publishers and subscriber.
+    /// Measurements are performed for subscriptions containing no wildcard, a single wildcard or multiple wildcards.
+    /// </summary>
+    public class MessageThroughputTest
+    {
+        // Change these constants to suit
+        const int NumPublishers = 5000;
+        const int NumTopicsPerPublisher = 10;
+
+        // Note: Other code changes may be required when changing this constant:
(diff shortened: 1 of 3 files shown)


JanEggers:
we already have some benchmarks with benchmark.net can you adjust your code to run with benchmark.net so we have proper warmup and measuring with standard deviation?

logicaloud:
@JanEggers Benchmark.net was my first though too, but this message throughput test is more aimed at running "scenarios" where Benchmark does not quite fit, i.e. "use this many publishers, subscribers, topics, over this time period and see what kind of message throughput can be sustained". Currently it is not very flexible and uses a fixed set of publishers/subscribers/topics over a fixed time span to provide a quick qualitative indication of performance for comparison.

I agree that it would make sense to also measure the performance of some affected operations (i.e. single message exchange) in Benchmark.net as another point of reference. I'll have a look at what is not covered yet.

logicaloud:
@chkr1011 I have applied the optimization changes to the  _feature/Reason_Code_In_Subscription_Interceptor_ branch without much difficulty. The server there is faster to start with but the speedup (for the use case in this test) still appears to be about 15x to 30x.

What would be the best way going forward? Prepare a pull request now against  _feature/Reason_Code_In_Subscription_Interceptor_ to describe the optimization approach and review the code?

chkr1011:
@logicaloud Yes please create a pull request targeting the branch _feature/Reason_Code_In_Subscription_Interceptor_.

logicaloud:
I have included this test in #1283, so closing this one.

Issue #1208, opened 2021-08-02
Store pending messages on broker's side


elektromntr:
### Describe the feature request
I need to store messages pending in server's queues, that are not yet delivered to subscribers. And when server crashes, I would like to restore these queues to be sure that every message reaches each subscriber.

### Which project is your feature request related to?
- Server

### Describe the solution you'd like
I need to store messages pending in server's queues, that are not yet delivered to subscribers. And when server crashes, I would like to restore these queues to be sure that every message reaches each subscriber.
For example, when client (subscriber) is disconnected for some time, its queue on server side is populated with new pending messages. And when server crashes at this point, subscriber will lost those messages. Even when the server restarts and subscriber's reconnect will take place, messages are lost.

### Additional context
Place of storing should be decided on my own. It may be file, database or whatever.

@saba-sabrin I think we just store the messages in memory, so if the broker is restarted you will still loose messages. 

@chkr1011 should we extend IMqttServerStorage or provide a similar interface like IMqttSessionStorage   to store pending messages of a session?

_Originally posted by @JanEggers in https://github.com/chkr1011/MQTTnet/issues/498#issuecomment-448552957_

chkr1011:
This feature is still in progress but unfortunately it is hard to implement. Especially when performance should by high and feature set rich. The problem is when should the messages being stored? Every time a message is produced? Then the broker will get slow when a lot of traffic is happening. But when storing the messages on a clean exit of the server you will lose all messages on an app crash etc.

tcbarton:
I wanted to comment this is an important feature.  I understand the performance implications.  Other brokers offer session persistence on stop/start scenarios...I'm not sure how they work on crash protection.  Since there will be a performance impact for a crash protection setup, then it could be designed as an optional feature to turn on/off so it could be implemented by the user depending on their use case.  Right now I'm using mqtt for a platform that won't have extremely high throughput, but session persistence is very important.

Thanks for your work on this project.

logicaloud:
There are quite a few touch points in code where the "Persistent Session Manager" interface needs to be called but it all looks good so far. Because I have been working on the interface only (not the storage behind the interface), I think I'll open another issue or pull request "at the right time". I suspect that a good time for a pull request will be when the _Reason_Code_In_Subscription_Interceptor_ branch is merged into master.

mark-trumed:
@logicaloud This is a feature that I would very much like to utilize. What is the current status?

logicaloud:
[...]

PR #1283, opened 2021-11-05
Improved message processing performance for many topics

logicaloud:
### Motivation

In a typical use case for MQTT there may be many publishers (sensors or other devices in the field) and few subscribers (monitoring all or many topics). Each publisher may have one or more topic(s) to publish and therefore both, the number of publishers and the number of topics may be large.

Currently there are two bottlenecks in MQTTnet that slow down processing in this scenario:

1. For any message published, the server loops through all subscriber *and* publisher sessions to find subscriptions.

2. For any subscriber, the server loops through the number of subscribed topics until it finds a matching one.

I'd like to put a proposal up for discussion that speeds up processing of application messages (about 15x to 30x for many publishers and topics based on the additional test that is included for a quick assessment, same as #1282)

### Maintain subscribers in a separate container

Instead of placing all sessions into the one `_clientSession` container, subscribers are added into another `_subscriberSessions` container (if a client is a subscriber *and* a publisher then the client is present in both containers). The cost is some additional bookkeeping work upon subscription where each client session needs to maintain a list of subscribed topics. 

When the client subscribes to the first topic, then the session manager adds the client to the `_subscriberSessions` container, and when the client unsubscribes from the last topic then the session manager removes the client from the container. Now, when an application message arrives, only the list of subscribers need processing instead of looping through potentially thousands of publishers.

### Improve subscriber topic lookup

This one is more complicated. For each subscriber, it needs to be determined whether an application message matches any topic the subscriber has subscribed to. There may only be few subscribers but there may be many subscribed topics, including wildcard topics.

The proposed approach uses a topic hash and a hash mask calculated on the subscribed topic and the published topic (the application message topic) to find candidates for a match, with the existing match logic evaluating a reduced number of candidates. 

For each subscription, the topic hash and a hash mask is stored with the subscription, and for each application message received, the hash is calculated for the published topic before attempting to find matching subscriptions. The hash calculation itself is simple and does not have a large performance impact.

I'll first explain how topic hashes and hash masks are constructed and then how they are used.

### Topic hash 

Topic hashes are stored as 64-bit numbers. Each byte within the 64-bit number relates to one MQTT topic level. A checksum is calculated for each topic level by iterating over the characters within the topic level (cast to `byte`) and the result is stored into the corresponding byte of the 64-bit number. If a topic level contains a wildcard character, then `0x00` is stored instead of the checksum. 

If there are less than 8 levels then the rest of the 64-bit number is filled with `0xff`. If there are more than 8 levels then topics where the first 8 MQTT topic levels are identical will have the same hash value. 

This is the topic hash for the MQTT topic below: `0x655D4AF1FFFFFF`

mqtt-topic-hash-no-wildcard

This is the topic hash for an MQTT topic containing a wildcard: `0x655D00F1FFFFFF`

mqtt-topic-hash-single-wildcard

For topics that contain the multi level wildcard `#` at the end, the topic hash is filled with `0x00`: `0x65004A00000000`

mqtt-topic-hash-single-with-last-level-wildcard

### Topic hash mask

The hash mask simply contains `0xff` for non-wildcard topic levels and `0x00` for wildcard topic levels. Here are the topic hash masks for the examples above.

mqtt-topic-hash-mask-no-wildcard

mqtt-topic-hash-mask-single-wildcard

mqtt-topic-hash-mask-single-with-last-level-wildcard

### Topic hash and hash mask properties

The following properties of topic hashes and hash masks can be exploited to find potentially matching subscribed topics for a given published topic.

1. If a subscribed topic does not contain wildcards then the topic hash of the subscribed topic must be equal to the topic hash of the published topic, otherwise the subscribed topic cannot be a candidate for a match.

2. If a subscribed topic contains wildcards then the hash of the published topic masked with the subscribed topic's hash mask must be equal to the hash of the subscribed topic. I.e. a subscribed topic is a candidate for a match if: `(publishedTopicHash & subscribedTopicHashMask) == subscribedTopicHash`

3. If a subscribed topic contains wildcards then any potentially matching published topic must have a hash value that is greater than or equal to the hash value of the subscribed topic (because the subscribed topic contains zeroes in wildcard positions).

### Match finding

The subscription manager maintains two separate dictionaries to assist finding matches using topic hashes: a 
`_noWildcardSubscriptionsByTopicHash` dictionary containing all subscriptions that do not have wildcards, and a `_wildcardSubscriptionsByTopicHash` dictionary containing subscriptions with wildcards.

For subscriptions without wildcards, all potential candidates for a match are obtained by a single look-up (exploiting point 1 above).

For subscriptions with wildcards, the subscription manager loops through the wildcard subscriptions and selects candidates that satisfy condition `(publishedTopicHash & subscribedTopicMask) == subscribedTopicHash` (point 2). A loop could be avoided here by exploiting point 3 but the available C# containers (i.e. `SortedDictionary`) do not make that efficient.

(If wildcard subscriptions were stored into a balanced binary tree then a single binary search could find the first non-matching node that has `subscribedTopicHash > publishedTopicHash` (or "end"); all matching candidates from thereon can be found by traversing the tree in reverse until `(publishedTopicHash & subscribedTopicHashMask) == subscribedTopicHash` no longer holds true. I'm not sure how much faster such an approach would be in practice).

### Other considerations

Just a few more thoughts:

Characters in the topic string are cast to byte and any additional bytes in a multi-byte character are disregarded. My guess is that this does not impact performance in practice.

Future work may include adding the balanced binary tree as mentioned above. 

Instead of one-byte checksums per topic level, one-word checksums per topic level could be used. If most topics contained four levels or less then hash buckets would be shallower.

For very large numbers of topics, performing a parallel search may help further.

To also handle a larger number of subscribers, it may be beneficial to maintain a subscribers-by-subscription-topic dictionary.


logicaloud:
The commit list is getting a bit long. If any part of this is of interest then I can prepare a separate pull request with the relevant parts and squashed commits.

chkr1011:
I am already checking out the changes but they are quite a lot. And my primary interest is in handling collisions etc. The whole idea looks solid but I am wondering if collisions can happen if the topic gets longer and longer etc.

logicaloud:
Thank you for reviewing this. In the worst case scenario all topics would have the same hash value and then it falls back on looping through all topics to find a match, just like it is now. There is also a test that checks the hash bucket depth for the test case (50000 topics) and the maximum depth there is around 80; for most parts the depth is 1, I believe. So collisions are expected, but the number of potential candidates for match testing is reduced significantly.

chkr1011:
Ah OK got it. I will have a look at this PR in the next weeks. My primary focus is v4 preview and fixing some other issues. So please do not wonder if there is no progress on this change.

logicaloud:
Thank you for the heads-up. On my to-do list I still have the persisted-session handling (mentioned in #1208) and session/message expiry. 

The persisted session handling is fairly complete, following the event-based approach of interacting with implementers (i.e. database storage). There is a bit of overlap with this pull request so it is best merged or considered after this one.

If the session and/or message expiry is already in progress then I can wait, otherwise please let me know and I can contribute.

logicaloud:
I had to rewrite some history in the fork after merge which resulted in this untidy commit history. I'll regroup and recommit a separate pull request soon.

PR #1284, opened 2021-11-05
Fix NullReferenceException in MqttClientConnection.Dispose()

yyjdelete:


yyjdelete:
@chkr1011 
Is `_cancellationToken.Dispose();` in Dispose() really required? The CancellationTokenSource should always be disposed here.
https://github.com/chkr1011/MQTTnet/blob/0920c86d373cdce15efa27641c68bc468160c021/Source/MQTTnet/Server/Internal/MqttClientConnection.cs#L121

And since  Dispose() doesn't call `StopInternal()`, the CancellationToken can never be cancled after CancellationTokenSource is disposed, if Dispose() is called first by some reason?

chkr1011:
You are right. We can get rid of the disposable here at all. The _RunAsync_ method will always dispose the token properly.

chkr1011:
I will close this because @SeppPenner created a similar pull request including the full removal of the dispose pattern.

PR #1288, opened 2021-11-08
Removed IDisposable from MqttClientConnection class.

SeppPenner:
Fixes the issue from https://github.com/chkr1011/MQTTnet/pull/1284.


SeppPenner:
Fixes https://github.com/chkr1011/MQTTnet/issues/1292.

chkr1011:
@SeppPenner I think this is the same as #1284 right?

SeppPenner:
> 
> 
> @SeppPenner I think this is the same as #1284 right?

Yes, with the difference that I completely removed the IDisposable as you suggested. Just didn't want to add it directly to not break anything.

chkr1011:
Thank you for your contribution. Good work.

PR #1294, opened 2021-11-10
Fixes #1293, MQTT 3.1.1 session persistence when CleanSession=1

### LINKED ISSUE #1293: MQTT 3.1.1 client sessions persist when they should not
### Describe the bug
Currently the persistence of client sessions is purely controlled by server option `EnablePersistentSessions`. That means, sessions are only deleted on disconnect when this option is set to `false`.

A client that first connects with `CleanSession=1`, disconnects, and then re-connects later with `CleanSession=0` will find that a session is present. According to MQTT v3.1.1, this should not be the case:

_"If CleanSession is set to 1, the Client and Server MUST discard any previous Session and start a new one. This Session lasts as long as the Network Connection. State data associated with this Session MUST NOT be reused in any subsequent Session [MQTT-3.1.2-6]."_

- Server

### To Reproduce
The following test re-produces this error:

```
  [TestMethod]
  public async Task Clean_Session_Persistence()
  {
      using (var testEnvironment = new TestEnvironment(TestContext))
      {
          // Create server with persistent sessions enabled

          await testEnvironment.StartServer(o => o.WithPersistentSessions());

          const string ClientId = "Client1";

          // Create client with clean session and long session expiry interval

          var client1 = await testEnvironment.ConnectClient(o => o
              .WithProtocolVersion(Formatter.MqttProtocolVersion.V311)
              .WithTcpServer("127.0.0.1", testEnvironment.ServerPort)
              .WithSessionExpiryInterval(9999) // not relevant for v311 but testing impact
              .WithCleanSession(true) // start and end with clean session
              .WithClientId(ClientId)
              .Build()
          );

          // Disconnect; empty session should be removed from server

          await client1.DisconnectAsync();

          // Simulate some time delay between connections

          await Task.Delay(1000);

          // Reconnect the same client ID without clean session

          var client2 = testEnvironment.CreateClient();
          var options = testEnvironment.Factory.CreateClientOptionsBuilder()
              .WithProtocolVersion(Formatter.MqttProtocolVersion.V311)
              .WithTcpServer("127.0.0.1", testEnvironment.ServerPort)
              .WithSessionExpiryInterval(9999) // not relevant for v311 but testing impact
              .WithCleanSession(false) // see if there is a session
              .WithClientId(ClientId)
              .Build();


          var result = await client2.ConnectAsync(options).ConfigureAwait(false);

          await client2.DisconnectAsync();

          // Session should NOT be present for MQTT v311 and initial CleanSession == true

          Assert.IsTrue(!result.IsSessionPresent, "Session present");
      }
  }
```

Output:

```
Message: 
Assert.IsTrue failed. Session present

```

logicaloud:
Changes in this pull request add an `IsPersistent` flag to the `MqttClientSession` object. The flag is set when the session is created and does not change for the lifetime of the session. 

A session should be persistent if a client connects with `CleanSession == false` (for MQTT 3.1.1) or with `SessionExpiryInterval != 0` (for MQTT 5).

When a client disconnects then the `IsPersistent` flag is evaluated as well as the already existing `EnablePersistentSessions` server option. The session is deleted if either persistent sessions are not enabled or the if the session is not marked as persistent.

The change resulted in two other test updates; these test assumed that a session should persist purely based on the `EnablePersistentSessions` option. The updated tests now also use `CleanSession == false` to persist the session.

Fixes #1293

As far as I can see,  `SessionExpiryInterval` (for MQTT 5) is not yet evaluated - is this something best left for version 4?

### DIFF
--- a/Source/MQTTnet/Server/Internal/MqttClientSession.cs
+++ b/Source/MQTTnet/Server/Internal/MqttClientSession.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using MQTTnet.Server.Status;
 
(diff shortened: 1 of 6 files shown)


chkr1011:
The session expiry is planned for v4.

chkr1011:
Thank you for your contribution. Good work. 👍

PR #1295, opened 2021-11-12
MqttRpcClient now adds ResponseTopic to MqttApplicationMessage when using MQTT v5

MD-V:
Currently when using MqttRpcClient the ResponseTopic property is not set even when using protocol version V500.


chkr1011:

PR #1300, opened 2021-11-17
Fix MQTT5 'IsPersisted' session flag handling

logicaloud:
Related to #1293. This fixes a bug in setting the 'IsPersisted' session flag for connecting MQTT5 clients when the `SessionExpiryInterval` is undefined or when `SessionExpiryInterval` changes to zero or undefined for an existing session (then the previously persistent session should no longer be considered persistent). Two additional tests check these scenarios. Other than that, there is no additional handling for `SessionExpiryInterval` but I think this one still belongs into V3.


chkr1011:

PR #1302, opened 2021-11-19
Improve exception message when handling connection errors

patagonaa:
Right now, when the MQTT connection times out, the exception message is pretty confusing, as the message from the InnerException (`OperationCanceledException`) is used:
```
MQTTnet.Exceptions.MqttCommunicationTimedOutException: The operation was canceled.
 ---> System.OperationCanceledException: The operation was canceled.
   at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ThrowException(SocketError error, CancellationToken cancellationToken)
[...]
```
This PR changes the message from `The operation was canceled.` to `The operation has timed out.`, which seems more appropriate for this kind of exception.

### DIFF
--- a/Source/MQTTnet/Exceptions/MqttCommunicationTimedOutException.cs
+++ b/Source/MQTTnet/Exceptions/MqttCommunicationTimedOutException.cs
@@ -8,7 +8,7 @@ namespace MQTTnet.Exceptions
         {
         }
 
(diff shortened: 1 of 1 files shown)


chkr1011:

PR #1322, opened 2022-01-03
Fixed latent bug in MqttApplicationMessageBuilder.WithPayload stream overload

gfoidl:
`Stream.Read` can read less than requested bytes, so we need read in a loop until either the requested byte-count is read or 0 is returned.

It's hard to construct a proper test for this, but that behavior is documented.


chkr1011:

jogibear9988:
wouldn't it be better to only start a loop when the first read returns less bytes?

gfoidl:
@jogibear9988 that's basically done with the `do-while` loop.

jogibear9988:
I only thought it would maybe more effective. But I've not analyzed the IL...
So. OK

PR #1323, opened 2022-01-04
MqttNetSourceLogger with IsEnabled property and use it on hot-paths

gfoidl:
## Description

With that change the call to `_logger` is only done, when not the null-logger is used.
So valuetype arguments don't get boxed, and the arguments-array doesn't need to be allocated.

The change is applied not to all calls to `_logger`, only on hot-paths.
So connection, disconnections, and exception handling is covered by that change, to keep the code concise.

### Setup

Server / broker runs in a virtual machine, so to have some network latency, but not too much.
Client publishes 100.000 messages.

<details>
  <summary>Client code</summary>

```c#
using MQTTnet;
using MQTTnet.Client;
using MQTTnet.Client.Options;

public sealed class Worker : BackgroundService
{
    private const int Iterations = 100_000;

    private readonly IHostApplicationLifetime _applicationLifetime;

    public Worker(IHostApplicationLifetime hostApplicationLifetime)
    {
        _applicationLifetime = hostApplicationLifetime;
    }

    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {
        byte[] payload = new byte[100];
        Random.Shared.NextBytes(payload);

        IMqttClientOptions mqttClientOptions = new MqttClientOptionsBuilder()
            .WithClientId("client1")
            .WithTcpServer("10.0.0.20")
            .Build();

        using IMqttClient mqttClient = new MqttFactory().CreateMqttClient();

        await mqttClient.ConnectAsync(mqttClientOptions, stoppingToken);

        MqttApplicationMessage publishMessage = new MqttApplicationMessageBuilder()
            .WithTopic("test")
            .WithPayload(payload)
            .Build();

        for (int i = 0; i < Iterations; ++i)
        {
            await mqttClient.PublishAsync(publishMessage, stoppingToken);
        }

        await mqttClient.DisconnectAsync(stoppingToken);

        _applicationLifetime.StopApplication();
    }
}
```
</details>

### Baseline

Profile is based on 5e5818d7df78de6a0a1342c9dca692346f70ad97

baseline

### PR

perf_01

One can see that the ~100.000 allocations for the `object[]` (= args for the logger), and the ~100.000 boxing of `int` is gone.

Side note: the time shown also decreased (as expected), but I wouldn't count on this as it's a allocation profile.

### DIFF
--- a/Source/MQTTnet/Adapter/MqttChannelAdapter.cs
+++ b/Source/MQTTnet/Adapter/MqttChannelAdapter.cs
@@ -132,7 +132,10 @@ namespace MQTTnet.Adapter
 
                     Interlocked.Add(ref _bytesReceived, packetData.Count);
 
-                    _logger.Verbose("TX ({0} bytes) >>> {1}", packetData.Count, packet);
+                    if (_logger.IsEnabled)
+                    {
+                        _logger.Verbose("TX ({0} bytes) >>> {1}", packetData.Count, packet);
+                    }
                 }
                 catch (Exception exception)
                 {
@@ -180,7 +183,10 @@ namespace MQTTnet.Adapter
                     throw new MqttProtocolViolationException("Received malformed packet.");
                 }
 
-                _logger.Verbose("RX ({0} bytes) <<< {1}", receivedPacket.TotalLength, packet);
+                if (_logger.IsEnabled)
+                {
+                    _logger.Verbose("RX ({0} bytes) <<< {1}", receivedPacket.TotalLength, packet);
+                }
 
                 return packet;
             }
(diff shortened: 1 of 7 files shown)


chkr1011:
Writing _if_ statements at every single place where boxing should be avoided is tedious and error prone in my opinion. I pushed a change to master where all parameters are generic types to that they will not get boxes. The final array is only composed when the logger is enabled. Thanks to @Tymoniden for sharing the idea. Please let me know what you think about that strategy instead.

gfoidl:
https://github.com/dotnet/MQTTnet/commit/b362b2afc203c9c127b2e8adab03e6b96d25fbc1 looks good, and the allocation profile doesn't show up with the (valuetype) argument-boxes and the `object[]`.

A (theoretical) downside is that this may result in generic bloat. In pratice this won't be a problem, as the logging extension methods are used "internally" so it's under control what types are used or not.

Undoing the `if`-checks from this PR after _master_ is merged in, would result in an (almost) empty PR. So I'm going to close this PR here.

PR #1324, opened 2022-01-04
Elide repeated Action-delegate allocation by caching it

gfoidl:
Here C# compiler has too little knowledge about the usage of the delegate, we know better so cache it on our own to avoid the repeated allocation.

From the profiling session in https://github.com/chkr1011/MQTTnet/pull/1323 we see that there are ~100.000 allocations of the `Action`-delegate. They come from `cancellationToken.Register` and the C# compiler can't cache the method-group here.

With the change from this PR these allocations are gone. The following images shows the changes of https://github.com/chkr1011/MQTTnet/pull/1323 combined with this one:

perf_02

So for the client publishing use case only the allocations for tasks, the publish packet, and the publish result show up. 
To avoid these allocations I fear some bigger refactorings are needed, and thats a different story...

### DIFF
--- a/Source/MQTTnet/Implementations/CrossPlatformSocket.cs
+++ b/Source/MQTTnet/Implementations/CrossPlatformSocket.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.IO;
 using System.Net;
 using System.Net.Sockets;
@@ -11,12 +11,14 @@ namespace MQTTnet.Implementations
     public sealed class CrossPlatformSocket : IDisposable
     {
         readonly Socket _socket;
+        readonly Action _socketDisposeAction;
 
         NetworkStream _networkStream;
 
         public CrossPlatformSocket(AddressFamily addressFamily)
         {
             _socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
+            _socketDisposeAction = _socket.Dispose;
         }
 
         public CrossPlatformSocket()
@@ -24,12 +26,16 @@ namespace MQTTnet.Implementations
             // Having this constructor is important because avoiding the address family as parameter
             // will make use of dual mode in the .net framework.
             _socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
+
+            _socketDisposeAction = _socket.Dispose;
         }
 
         public CrossPlatformSocket(Socket socket)
         {
             _socket = socket ?? throw new ArgumentNullException(nameof(socket));
             _networkStream = new NetworkStream(socket, true);
+
+            _socketDisposeAction = _socket.Dispose;
         }
 
         public bool NoDelay
@@ -112,7 +118,7 @@ namespace MQTTnet.Implementations
                 _networkStream?.Dispose();
 
                 // Workaround for: https://github.com/dotnet/corefx/issues/24430
-                using (cancellationToken.Register(() => _socket.Dispose()))
+                using (cancellationToken.Register(_socketDisposeAction))
                 {
                     cancellationToken.ThrowIfCancellationRequested();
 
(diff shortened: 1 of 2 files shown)


Source/MQTTnet/Implementations/MqttTcpChannel.cs
```diff
@@ -150,7 +156,7 @@ public async Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancellat
                 }
 
                 // Workaround for: https://github.com/dotnet/corefx/issues/24430
```
gfoidl, line 158: Question: starting with .NET Core 3.0 this workaround isn't needed anymore, so the code could be conditionally compiled for a "legacy" target and for new targets.

W/o the workaround perf in general should be better, as the cost of creating, and disposing the cancellationtoken-registration can be avoided.

chkr1011: If it is fixed we should add a compiler switch (NETCOREAPP3_0_OR_GREATER).

Source/MQTTnet/Implementations/MqttTcpChannel.cs
```diff
@@ -150,7 +156,7 @@ public async Task<int> ReadAsync(byte[] buffer, int offset, int count, Cancellat
...
-                using (cancellationToken.Register(Dispose))
```
chkr1011, line 153: In the above code I agree that the lambda is not required but here we already have a method without wrapping it. Do we really need that optimization here as well?

chkr1011 (changes requested):

Source/MQTTnet/Implementations/MqttTcpChannel.cs
gfoidl: From my profiling it showed up in `WriteAsync` (L187), so it's just duplicated here.

But when the conditional compilation is added for .NET Core 3.0 and newer, this won't be hit. So for the legacy code we could
* keep the state of _master_-branch
* keep the cached delegate of PR's current state

I'm leaning to the latter, as the change isn't huge and so older targets can benefit from that change too.

gfoidl: Ah, now I see your concern. Yes we need it. Whether it's a delegate created from a method group or a lambda, the delegate won't be cached. Checked the IL for both.

chkr1011: OK then we should keep it.

chkr1011 (changes requested):

Source/MQTTnet/Implementations/CrossPlatformSocket.cs
```diff
@@ -112,7 +118,7 @@ public async Task ConnectAsync(string host, int port, CancellationToken cancella
                 _networkStream?.Dispose();
 
                 // Workaround for: https://github.com/dotnet/corefx/issues/24430
-                using (cancellationToken.Register(() => _socket.Dispose()))
```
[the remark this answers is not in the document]
gfoidl: `Socket.ConnectAsync(string, int, CancellationToken)` exists only starting from .NET 6.
So for the moment we have to stay at the current code.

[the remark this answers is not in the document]
chkr1011: OK.

chkr1011 (approved):

chkr1011:

yyjdelete:
@gfoidl 
Maybe it's better to use static Action instead of instance ones, and pass `_socket` or `this` as state for `CancellationToken.Register()`
```cs
static readonly Action<object> _socketDisposeAction = state => ((Socket)state).Dispose();

using (cancellationToken.Register(_socketDisposeAction, _socket))
```

gfoidl:
@yyjdelete that's another way to do it, but I don't know if it is actually "better".
Due the static field the instance gets smaller a few bytes, but these objects are intended to live longer, so it won't matter much.
Calling a static delegate from an instance method has more runtime overhead (due to shuffle thunks that the runtime needs to create), which won't matter here too (could have an effect on (very) hot code-pathes).

So for the use here it could be done either way. In this PR I just the most easy to write and read way.
But I'm open to change it to your suggestion, leaving the decision to up to others (or you open a PR and it can be discussed further over there).

Issue #1347, opened 2022-01-31
SSLProtocols default value is invalid under .net 4.6 and earlier


simonthum:
### Describe the bug
Just enabling TLS in MQTTnet will yield a Sslprotocols value of `SslProtocols.None`. This value is invalid under .net 4.6 and earlier, a fact enforced in .net 4.8 too.

Nota bene: In 4.7 and beyond, `SslProtocols.None` leaves it to the OS to decide, which is not a bad idea, so in any case the fix sould not change the default for all targets.

.net side enforcement:

https://referencesource.microsoft.com/#System/net/System/Net/SecureProtocols/_SslState.cs,165

Default 1:

https://github.com/dotnet/MQTTnet/blob/d5691346a9d2433e3530fda571f1c8defbe70c6f/Source/MQTTnet/Client/Options/MqttClientOptionsBuilderTlsParameters.cs#L22

Default 2:

https://github.com/dotnet/MQTTnet/blob/d5691346a9d2433e3530fda571f1c8defbe70c6f/Source/MQTTnet/Client/Options/MqttClientTlsOptions.cs#L29

### Which project is your bug related to?
- Client
- ManagedClient

### To Reproduce
Open a connection using TLS in a .net 4.6 vanilla (no special SSL settings, neither in .net nor MQTTnet) compiled app. You do not need an actual broker, it will fail with the ArguementException inidcated above.

### Expected behavior
Some default suite of SSL versions should be picked to at least attempt a connection.

### Code example

```csharp
clientOptionsBuilder.WithTls(); // Would work if we weren't still on .net 4.6 ;)
```

simonthum:
This was last changed in Oct. 2021: https://github.com/dotnet/MQTTnet/pull/1271

I favor leaving SSLprotocols to the OS, but older .net will fail, so you have to specify. Note this is an enum Flag, you can specify `SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12`, no need to single out a specific revision.

simonthum:
To be clear, this isn't so bad. It's quite straightforward to get from error message to the workaround, i.e. explicitly naming TLS versions. But I still think a default value should not cause needless failure.

chkr1011:
That should be a small change. I will consider adding this soon.

simonthum:
Thank you!

SeppPenner:
[...]

PR #1348, opened 2022-02-02
Adjusted default value for the supported Net frameworks versions where "None" is incompatible.

SeppPenner:
Fixes https://github.com/dotnet/MQTTnet/issues/1347.

However, I don't know whether NetStandard or UWP or something is concerned here, too.

### DIFF
--- a/Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
+++ b/Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
@@ -52,6 +52,12 @@ namespace MQTTnet.Extensions.WebSocket4Net
                 }
             }
 
+#if NET452 || NET46 || NET461 || NET462
+        var sslProtocols = _webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
+#else
+        var sslProtocols = _webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.None;
+#endif
+
             var sslProtocols = _webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.None;
             var subProtocol = _webSocketOptions.SubProtocols.FirstOrDefault() ?? string.Empty;
 
--- a/Source/MQTTnet/Client/Options/MqttClientOptionsBuilderTlsParameters.cs
+++ b/Source/MQTTnet/Client/Options/MqttClientOptionsBuilderTlsParameters.cs
@@ -19,7 +19,11 @@ namespace MQTTnet.Client.Options
 
         public Func<MqttClientCertificateValidationCallbackContext, bool> CertificateValidationHandler { get; set; }
 
+#if NET452 || NET46 || NET461 || NET462
+        public SslProtocols SslProtocol { get; set; } = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
+#else
         public SslProtocols SslProtocol { get; set; } = SslProtocols.None;
+#endif
 
 #if WINDOWS_UWP
         public IEnumerable<IEnumerable<byte>> Certificates { get; set; }
--- a/Source/MQTTnet/Client/Options/MqttClientTlsOptions.cs
+++ b/Source/MQTTnet/Client/Options/MqttClientTlsOptions.cs
@@ -26,7 +26,11 @@ namespace MQTTnet.Client.Options
         public List<SslApplicationProtocol> ApplicationProtocols { get; set; }
 #endif
 
+#if NET452 || NET46 || NET461 || NET462
+        public SslProtocols SslProtocol { get; set; } = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
+#else
         public SslProtocols SslProtocol { get; set; } = SslProtocols.None;
+#endif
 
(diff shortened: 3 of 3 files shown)


Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
```diff
@@ -52,6 +52,12 @@ public async Task ConnectAsync(CancellationToken cancellationToken)
                 }
             }
 
+#if NET452 || NET46 || NET461 || NET462
+        var sslProtocols = _webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
```
chkr1011, line 56: In my opinion we should only allow 1.2 by default. 1.1 is clearly obsolete and should be activated by users actively, knowing the risk.

chkr1011 (changes requested):

Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
simonthum: I second that. IETF declared 1.0 and 1.1 "historic" last year, it's time to move on.

Also, you can enable 1.3 even in net 4.5/4.6 using a numeric enum constant. A bit hacky, but better IMHO.

simonthum:
I think .net 4.7 is affected, please see the issue for detail.

Source/MQTTnet.Extensions.WebSocket4Net/WebSocket4NetMqttChannel.cs
SeppPenner: From the issue and here combined: 
@chkr1011 What do you think of 1.3 as numeric value?

I will add NET47 to the filter.

simonthum:
I gave this a bit more thought; please bear with me.

The current proposal will make OLDER .net versions MORE SECURE in that they sport a more modern set of allowed protocol(s), i.e. TLS 1.2  only at the moment. Newer .net will yield to schannel, which in case of Windows 11 will happily connect using TLS 1.0. I think this is quite confusing.

I see two non-confusing options:

1) Use SslProtocol.None as a moniker to use SslStream overload without explicit SslProcol argument. Being a "mere" library, this yields the burden of choice to the current .net+schannel mashup the developers of a given solution should already know how to work with or against. On an older .net, this means outdated defaults (only 1.0 and 1.1), and newer .net will fall back to 1.0. On the upside, no .net version compiler black magic.

2) Let MQTTnet make a deliberate, well-documented default of TLS 1.2+TLS1.3. This is 2022, and you still have the option to specify weaker protocol. 4.6 could do 1.2 only, or use a numeric enum definition (I tested this in another setting).

I think you should rather choose from these two options.

chkr1011:
In my opinion option 2 ist best because it is explicit and no implicit knowledge or lots of comments are required to explain why certain (empty) constructors etc. must be used.

SeppPenner:
>     2. Let MQTTnet make a deliberate, well-documented default of TLS 1.2+TLS1.3. This is 2022, and you still have the option to specify weaker protocol. 4.6 could do 1.2 only, or use a numeric enum definition (I tested this in another setting).

Er, okay.

Is this meant to look somehow like this or did I understand this completely wrong (I didn't have the time yet to check how the integer value should be set)?

```csharp
#if NET452 || NET46 || NET461 || NET462 || NET47
            var sslProtocols = _webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.Tls12 | (SslProtocols)13;
#else
            var sslProtocols = _webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.None;
#endif
```

simonthum:
@SeppPenner Basically, yes.

But the expression `_webSocketOptions?.TlsOptions?.SslProtocol ?? SslProtocols.None` MAY never hit because SslProtocol is not an optional enum (not `SslProtocol?`).

I would suggest to just change the ctor assignment in the tls options. That way None is still an option users can set, plus the default becomes visible in the debugger. AFAICT there is no need to change other code.

On .net 4.6 and below, you may want to use `SslProtocols.Tls12|(SslProtocols)0x00003000 /*Tls13*/` instead of the symbols. (4.7.0 should already have the Tls13 Symbol).

simonthum:
Oh, sorry I misread your  example!

Option 2) means `SslProtocols.None` is not a default any more. The new default for all targets would be Tls12|Tls13 (or whatever), and target discrimination ideally only serves compiler compatibility, i.e. no or only 'expected' semantic changes.

chkr1011:
@SeppPenner  Can I support here? I want to start with 4.0 beta releases which should include this change.

SeppPenner:
> (4.7.0 should already have the Tls13 Symbol).

Nope: https://docs.microsoft.com/en-us/dotnet/api/system.security.authentication.sslprotocols?view=netframework-4.7

As far as I have seen it, only Net6.0, Net5.0, NetCore3.0 and 3.1 and NetFramework4.8 support the name. Please check my changed pull request :)

SeppPenner:
> @SeppPenner Can I support here? I want to start with 4.0 beta releases which should include this change.

It should be alright now (I hope) 😄

chkr1011 (approved):

simonthum:
Thanks!

One nit: I suggest to special-case the older versions requiring the numeric enum. A new .net 7 is much more likely than .net 4.9 ;)

Issue #1362, opened 2022-03-08
Project build errors


friesendrywall:
Using Visual Studio Community 2019 the project build errors out, perhaps due to not checking every possibility on my initial install, Its not clear.  Could there be another readme or addition with the prerequisites for successfully building this project?

In my case, I just want the standard windows build so I can test with this.

![Screenshot 2022-03-08 09 29 55](https://user-images.githubusercontent.com/5509710/157258224-3a54e628-0203-4b42-8e63-e6529901dc75.png)
.

SeppPenner:
Two problems I can find from reading the error messages:

1. Visual Studio 2019 doesn't support Net6.0 builds, you need to switch to VS2022.
2. If you don't have installed Net6.0, you need to install it, since the projects have dependencies to the framework.

The only other option is to remove all `net6.0` frameworks from the *csproj files (Located under `TargetFrameworks` in the project (In your local folder) and try again.

chkr1011:
Another dependency is the Windows SDK for UWP builds. The project requires the SDK 10.0.18362.0 to be installed. This can be installed via the Visual Studio Installer (Individual Components).

friesendrywall:
Thank you for that clarification.  For now I just ended up using the 3.1.2 standard nuget package.  But it would be nice to include this basic build requirement info somewhere in the repo.

SeppPenner:
> Another dependency is the Windows SDK for UWP builds. The project requires the SDK 10.0.18362.0 to be installed. This can be installed via the Visual Studio Installer (Individual Components).

Er, the readme states 10.0.10240+ as supported UWP version. This doesn't seem to fit...

I've added a small pull request, just let me know if the UWP version needs to be changed :)

SeppPenner:
I have added this to the wiki now: https://github.com/dotnet/MQTTnet/wiki/Build.

PR #1364, opened 2022-03-09
Add build doc to readme.

SeppPenner:
Fixes https://github.com/dotnet/MQTTnet/issues/1362.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,11 @@ MQTTnet is a high performance .NET library for MQTT based communication. It prov
 | Xamarin.iOS        | 10.14+      |
 | Blazor WebAssembly | 3.2.0+      |
 
+## Build requirements
+
+* Visual Studio 2022 (Net6.0 is targeted in the project and Visual Studio 2019 won't work with Net6.0)
+* UWP Support for Visual Studio (Windows SDK for UWP version 10.0.18362.0+)
+
 ## References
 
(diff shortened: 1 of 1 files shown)


chkr1011:
Please move this to the Wiki instead. In my opinion this is nothing very important which should be present on the first page which is shown to the users.

README.md
```diff
@@ -66,6 +66,11 @@ MQTTnet is a high performance .NET library for MQTT based communication. It prov
 | Xamarin.iOS        | 10.14+      |
 | Blazor WebAssembly | 3.2.0+      |
 
+## Build requirements
+
+* Visual Studio 2022 (Net6.0 is targeted in the project and Visual Studio 2019 won't work with Net6.0)
+* UWP Support for Visual Studio (Windows SDK for UWP version 10.0.18362.0+)
```
chkr1011, line 72: We need exactly 10.0.18362.0. This value is also configured in the .csproj file. There is also the minimum compatible version specified.

SeppPenner: Ok, thanks for the information. I have no idea since I only use NetFramework 4.8 (Where I didn't have the energy to solve Net6.0 compatibility issues) and Net6.0 for now 😄

SeppPenner:
It's documented under https://github.com/dotnet/MQTTnet/wiki/Build.

Issue #1286, opened 2021-11-08
Retain the the `IMqttRetainedMessagesManager` interface in version 4?


logicaloud:
### Describe the feature request
It looks like the upcoming version 4 (referring to branch `feature/Reason_Code_In_Subscription_Interceptor`) no longer exposes the `IMqttRetainedMessagesManager` interface as an extension point. Using the `IMqttServerStorage` interface instead is fairly limiting and may not be the best option when dealing with physical storage (see also comments in #1206). So far the `IMqttRetainedMessagesManager` could be used to inject one's own retained message manager, which means one is not limited by `IMqttServerStorage` .

### Which project is your feature request related to?
- Server

### Describe the solution you'd like
It would be good to keep the `IMqttRetainedMessagesManager` extension point available in version 4.

### Describe alternatives you've considered
Alternatively the `IMqttServerStorage` could be modified to contain the methods that were previously in the `IMqttRetainedMessagesManager` interface.

logicaloud:
I have noticed that the new philosophy is to use the event handler based approach on MqttServer; this works fine for me. Closing.

fogzot:
Does the new handler-based approach in `IMqttServerStorage` allows for completely custom retain logics? Like storing the retained messages on Redis and retrieving them on the fly, like we're doing today using the old interface?

logicaloud:
@fogzot I think you can do the same as with `IMqttRetainedMessagesManager`. Instead of

`.WithRetainedMessagesManager(...)`

you hook up events like so:

```
server.RetainedMessageChangedAsync += Server_RetainedMessageChangedAsync;
server.LoadingRetainedMessageAsync += Server_LoadingRetainedMessageAsync;
server.RetainedMessagesClearedAsync += Server_RetainedMessagesClearedAsync;
```

fogzot:
IMHO, this is a regression because it always keep a copy of all retained messages in memory while the interface could have been improved to allow for a completely different storage.

logicaloud:
The interface is more concise and convenient for extension developers, I agree, but whether retained message payloads are held in memory or for how long, etc., is dictated by the internal handling of retained messages, not by the event vs interface exposure. It would be fairly easy to re-introduce the interface and hookup events internally to the interface I think. I'll reopen and maybe @chkr1011 can comment.

chkr1011:
[...]

Issue #1206, opened 2021-07-27
Saving retained application messages


elektromntr:
### Describe your question
Hi, I've seen in documentation some implementation of Storage for "Saving retained application messages":
https://github.com/chkr1011/MQTTnet/wiki/Server#saving-retained-application-messages

When those methods _SaveRetainedMessagesAsync_ and _LoadRetainedMessagesAsync_ are called?
I have subscribers with persistent sessions and publish messages mostly with QoS > 0

### Which project is your question related to?
- Server

BrightSoul:
Disclaimer: not a Mqttnet dev.

From my experience with the product, the `LoadRetainedMessagesAsync` is [called once by the default MqttRetainedMessageManager](https://github.com/chkr1011/MQTTnet/blob/8f1d4e3c22f3226570eec681bccef3a88c4ebd16/Source/MQTTnet/Server/MqttRetainedMessagesManager.cs#L30) at server startup. In there, you'll have to retrieve ALL retained messages from the database and they will be kept in memory. That's right, if you have 100k retained messages, they will ALL sit in memory. And since each message can be quite large, you could end up wasting a great amount of RAM.

The `SaveRetainedMessagesAsync` is called each time a new persisted message is sent to the server. The method [gets passed the whole list of retained messages](https://github.com/chkr1011/MQTTnet/blob/8f1d4e3c22f3226570eec681bccef3a88c4ebd16/Source/MQTTnet/Server/MqttRetainedMessagesManager.cs#L99) (the aforementioned 100k messages), not just the one that was sent.

After struggling a bit with this, I ended up writing my own implementation of the `IMqttRetainedMessagesManager`. In there, you'll have a chance to load/save only the relevant retained messages.
- In the `GetSubscribedMessagesAsync` method, load and return only the persisted messages that match the provided topic filters;
- In the `HandleMessageAsync` method, persist the provided message in the database;
- Just return `Task.CompletedTask` in all other methods.

elektromntr:
Hi @BrightSoul 
Big big thanks for your reply. It's very helpful.
I have 2 more questions, related to your response:

1. When you say _persisted message_, you mean every message that has QoS not equal to 0?
2. When persisted message is removed from that aforementioned 100k list? When is acknowledged as delivered?

BrightSoul:
Hey @elektromntr 

> When you say persisted message, you mean every message that has QoS not equal to 0?

[...]

PR #1392, opened 2022-04-20
Add List of TopicFilters when getting Retained messages

YarekTyshchenko:
This is so that storage can fetch only the messages that it needs
based on the current subscription, and not all messages.

If the broker is backed by a database, keeping all messages in memory
doesn't make sense. It would be better to call up individual sets of
messages on demand.

See #1286 and https://github.com/dotnet/MQTTnet/issues/1206#issuecomment-887417891

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,17 +30,11 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v2
 
-      - name: Setup Signing Certificate
-        run: |
-          $secret = '${{ secrets.SNC_BASE64 }}'
-          $decoded = [System.Convert]::FromBase64CharArray($secret, 0, $secret.Length)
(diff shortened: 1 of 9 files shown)


chkr1011:
Please do not create pull requests as drafts. There is no chance for us to known when they are finished.
Also do not change any parts which are related to CI configuration or nuget packages.

YarekTyshchenko:
@chkr1011 The CI Changes weren't meant to go into this PR. I had trouble building the project on Linux, so had to use Github Actions. I can't think of any other way to package it locally for testing. I'll split and force push in a minute.

I've set the PR as Draft because I realised my changes won't work, since the interface to the retained message manager is not exposed. Do you have any advice on how to make the change I'm looking for?

chkr1011:
You can just rebuild the solution. It should generate the nuget packages but with version 0. This should be enough for testing. 

The topic you are facing is quite hard. I will think about it and let you know.

Issue #1186, opened 2021-06-29
WithNoLocal(true) fails to suppress receipt of message


PeterWone:
### Describe the bug
`WithNoLocal(true)` fails to suppress receipt of message a message published by the same client.

### Which project is your bug related to?

- ManagedClient

### To Reproduce
1. Using this version of MQTTnet 3.0.15 
2. Run this code against current Mosquitto, either a local instance unauthenticated or running on another computer configured using the default password file remembering to uncomment the WithCredentials line.
```
using System;
using System.Threading;
using MQTTnet;
using MQTTnet.Client.Options;
using MQTTnet.Client.Receiving;
using MQTTnet.Extensions.ManagedClient;
using MQTTnet.Formatter;

namespace MockApp
{
  class Program
  {
    static string ClientId = $"MockApp2_{DateTime.Now.Ticks}";
    static void Main(string[] args)
    {
      // Setup and start a managed MQTT client.
      var options = new ManagedMqttClientOptionsBuilder()
      .WithAutoReconnectDelay(TimeSpan.FromSeconds(5))
      .WithClientOptions(new MqttClientOptionsBuilder()
          .WithProtocolVersion(MqttProtocolVersion.V500)
          .WithClientId(ClientId)
          .WithTcpServer("localhost")
          // .WithCredentials("roger", "password")
          .Build())
      .Build();

      var mqttClient = new MqttFactory().CreateManagedMqttClient();
      mqttClient.ApplicationMessageReceivedHandler = new MqttApplicationMessageReceivedHandlerDelegate(mre =>
      {
        var payload = mre.ApplicationMessage.ConvertPayloadToString();
        Console.WriteLine($"RECV {payload}");
      });
      mqttClient.StartAsync(options).Wait();

      var arbitraryTopicName = "wibble";
      mqttClient.SubscribeAsync(new MqttTopicFilterBuilder()
      .WithTopic(arbitraryTopicName)
      .WithNoLocal(true)
      .Build());

      Thread.Sleep(300);
     
      mqttClient.PublishAsync(arbitraryTopicName, "This should not be seen by ApplicationMessageReceivedHandler");

      Console.WriteLine(ClientId);
      Console.ReadKey();
    }
  }
}

```
3. Local message is received.

### Expected behavior
`.WithNoLocal(true)` should suppress receipt of the message.

### Screenshots
[...]

PR #1405, opened 2022-05-04
re: #1186 - ManagedClient preserving TopicFilters across reconnects

gpd-jkohler:
modified ManagedClient to retain TopicFilters in subscription collections rather than just QOS.
- subscription/reconnectSubscription collections changed to hold MqttTopicFilter
- retained MqttTopicFilters used in subscriptions rather than constructing new MqttTopicFilters at subscription time
- unit test added

### DIFF
--- a/Source/MQTTnet/Packets/MqttTopicFilter.cs
+++ b/Source/MQTTnet/Packets/MqttTopicFilter.cs
@@ -45,7 +45,15 @@ namespace MQTTnet.Packets
         ///     separator).
         /// </summary>
         public string Topic { get; set; }
-
+        /// <summary>
+        /// Compares TopicFilters on Topic strings.  Overlapping Topics are not considered.
+        /// </summary>
+        public override bool Equals( object other )
+        {
+            if ( !( other is MqttTopicFilter otherFilter ) )
+                return false;
+            return this.Topic == otherFilter.Topic;
+        }
         public override string ToString()
         {
             return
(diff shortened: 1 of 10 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/MQTTnet?pullRequest=1405) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/MQTTnet?pullRequest=1405) before we can accept your contribution.<br/><br/>:x: gpd-jkohler [sign now](https://cla.dotnetfoundation.org/dotnet/MQTTnet?pullRequest=1405)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/MQTTnet?pullRequest=1405) it.</sub>

Source/MQTTnet/Packets/MqttTopicFilter.cs
```diff
@@ -45,7 +45,15 @@ public sealed class MqttTopicFilter
         ///     separator).
         /// </summary>
         public string Topic { get; set; }
-
+        /// <summary>
+        /// Compares TopicFilters on Topic strings.  Overlapping Topics are not considered.
+        /// </summary>
+        public override bool Equals( object other )
```
chkr1011, line 51: Instead of comparing the topic here only I recommend to only compare the topic where it is used in the other code. So that this Equals override is no longer needed. 

From MSDN (https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-6.0):

> You should not override [Equals](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-6.0) on a mutable reference type. This is because overriding [Equals](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-6.0) requires that you also override the [GetHashCode](https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-6.0) method, as discussed in the previous section. This means that the hash code of an instance of a mutable reference type can change during its lifetime, which can cause the object to be lost in a hash table.

It is also dangerous in my opinion to just compare some of the properties and not all because the caller might not know this.

chkr1011 (changes requested):

gpd-jkohler:
superseded by rev 4

PR #1447, opened 2022-06-28
Use IgnoreCertificateChainErrors option in TcpChannel

GodVenn:
The IgnoreCertificateChainErrors was not being used to actually ignore this type of SSL Error in the [MqttTcpChannel](https://github.com/dotnet/MQTTnet/blob/73d681bc1f978c4e6cf03266fcd1fb4a30a5d205/Source/MQTTnet/Implementations/MqttTcpChannel.cs).
The only place i could find that it was being used for .NETCOREAPP 3.1 / .NET 5 or greater was in the [MqttClientDefaultCertificateValidationHandler.cs](https://github.com/dotnet/MQTTnet/blob/73d681bc1f978c4e6cf03266fcd1fb4a30a5d205/Source/MQTTnet/Client/Options/MqttClientDefaultCertificateValidationHandler.cs), but this is not being referenced anywhere.

This PR adds a check for this option and accepts the corresponding SSL Error in the case it is enabled.

I hope I can contribute to this project, and please let me know if I have misunderstood something! :)

### DIFF
--- a/Source/MQTTnet/Implementations/MqttTcpChannel.cs
+++ b/Source/MQTTnet/Implementations/MqttTcpChannel.cs
@@ -3,6 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 #if !WINDOWS_UWP
+using MQTTnet.Channel;
+using MQTTnet.Client;
+using MQTTnet.Exceptions;
 using System;
 using System.IO;
 using System.Net.Security;
@@ -11,9 +14,6 @@ using System.Runtime.ExceptionServices;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
-using MQTTnet.Channel;
-using MQTTnet.Client;
-using MQTTnet.Exceptions;
 
 namespace MQTTnet.Implementations
 {
@@ -77,7 +77,7 @@ namespace MQTTnet.Implementations
                 {
                     socket.LingerState = _tcpOptions.LingerState;
                 }
-                
+
                 if (_tcpOptions.DualMode.HasValue)
                 {
                     // It is important to avoid setting the flag if no specific value is set by the user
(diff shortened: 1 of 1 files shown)


Source/MQTTnet/Implementations/MqttTcpChannel.cs
```diff
@@ -274,6 +274,10 @@ bool InternalUserCertificateValidationCallback(object sender, X509Certificate x5
                 return certificateValidationHandler(eventArgs);
             }
 
+            if(_tcpOptions?.TlsOptions?.IgnoreCertificateChainErrors ?? false)
```
chkr1011, line 277: Since the _SslPolicyErrors_ are flags I recommend to check for existence of "SslPolicyErrors.RemoteCertificateChainErrors" via "HasFlag()" and removing it when _IgnoreCertificateChainErrors _ is true. Then further properties can be added which will "remove" more flags from the validation.

chkr1011 (changes requested):

Source/MQTTnet/Implementations/MqttTcpChannel.cs
GodVenn: I used the AND operator along with the inversion of the bits to remove the flag. 
This will remove it if it exists and leave it unchanged if not.

chkr1011:

d5UtQvp8QQU9:
I wonder, why is `MqttClientDefaultCertificateValidationHandler.Handle()` not the default ValidationHandler? Same reasoning for `IgnoreCertificateChainErrors` would also be valid for `AllowUntrustedCertificates`.

Something like this would solve the problem in a way that I would expect from a class named _MqttClientDefaultCertificateValidationHandler_:
````
bool InternalUserCertificateValidationCallback(object sender, X509Certificate x509Certificate, X509Chain chain, SslPolicyErrors sslPolicyErrors)
{
    var certificateValidationHandler = _tcpOptions?.TlsOptions?.CertificateValidationHandler ?? MqttClientDefaultCertificateValidationHandler.Handle;

    var eventArgs = new MqttClientCertificateValidationEventArgs
    {
        Certificate = x509Certificate,
        Chain = chain,
        SslPolicyErrors = sslPolicyErrors,
        ClientOptions = _tcpOptions
    };

    return certificateValidationHandler(eventArgs);
}
````

PR #1466, opened 2022-07-21
Add PackageIdentifier to MqttApplicationMessageReceivedEventArgs

koepalex:
## Changes
* Provide `PackageIdentifier` as read only property in `MqttApplicationMessageReceivedEventArgs`
* Fixed small typo in XML comment

## Reason
* While working with different MQTT brokern, it happend that some messages were lost, we would like to implement an application-level logging for messages received
* We intend to use delayed acknowledge (`AutoAccknowledge=false`) and like to track in the logs which packages were not yet acknowledged

### DIFF
--- a/Source/MQTTnet/Client/Receiving/MqttApplicationMessageReceivedEventArgs.cs
+++ b/Source/MQTTnet/Client/Receiving/MqttApplicationMessageReceivedEventArgs.cs
@@ -25,6 +25,7 @@ namespace MQTTnet.Client
             ClientId = clientId;
             ApplicationMessage = applicationMessage ?? throw new ArgumentNullException(nameof(applicationMessage));
             PublishPacket = publishPacket;
+            PacketIdentifier = publishPacket.PacketIdentifier;
             _acknowledgeHandler = acknowledgeHandler;
         }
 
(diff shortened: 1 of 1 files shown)


chkr1011:
Thank you for your contribution. Good work 👍 I assume it will be released with version 4.1.0.

Source/MQTTnet/Client/Receiving/MqttApplicationMessageReceivedEventArgs.cs
```diff
@@ -25,6 +25,7 @@ public sealed class MqttApplicationMessageReceivedEventArgs : EventArgs
             ClientId = clientId;
             ApplicationMessage = applicationMessage ?? throw new ArgumentNullException(nameof(applicationMessage));
             PublishPacket = publishPacket;
+            PacketIdentifier = publishPacket.PacketIdentifier;
```
schlotter, line 28: @chkr1011 @koepalex If `publishPacket` is `null`, this will throw. Maybe a `null` check for `publishPacket` should be added, like for `applicationMessage` in line 26?

schlotter: ...or `PacketIdentifier` should not be set if `publishPacket` is `null`, leaving `PacketIdentifier` at its default value. This is the option that I would prefer, as existing code (for example unit testing code) does not need to be adapted to always supply a non-`null` `publishPacket`.

PR #1489, opened 2022-08-09
MqttApplicationMessageReceivedEventArgs: Prevent NullReferenceException.

schlotter:
Only set `PacketIdentifier` property of `MqttApplicationMessageReceivedEventArgs` if `publishPacket` is not `null`.

This prevents a `NullReferenceException` in cases `publishPacket` is `null`.

This should not happen in production code, but accepting a `publishPacket` argument which is `null` is useful in unit testing, where `publishPacket` is not important in many cases. In these unit test cases, it is convenient to initialize the `publishPacket` argument with `null`.

Based on: #1466


chkr1011:
I will not approve this pull request because in my opinion this is not the right decision. In my opinion I forgot the null check for the parameter. So that ideally an exception will be thrown because the publish packet is mandatory. Unit Tests from other projects are not relevant for design decisions etc. made in this project. I assume it should be easy to just create an instance of the publish packet because it requires no parameters at all so that this choice should also perfectly work in your mentioned Unit Tests.

PR #1504, opened 2022-08-29
Fix Exception "Collection was modified"

wdichler:
It leads to Exception "Collection was modified; enumeration operation may not execute." if during execution of InvokeAsync, RemoveHandler is called.
This is the case if we use RxMqttNet.

### DIFF
--- a/Source/MQTTnet/Internal/AsyncEvent.cs
+++ b/Source/MQTTnet/Internal/AsyncEvent.cs
@@ -37,7 +37,7 @@ namespace MQTTnet.Internal
 
         public async Task InvokeAsync(TEventArgs eventArgs)
         {
-            foreach (var handler in _handlers)
+            foreach (var handler in _handlers.ToArray())
             {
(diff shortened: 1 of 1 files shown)


chkr1011:
I assume we have to lock the access to the internal list because it might happen that the collection gets removed when the array is being constructed and all items are being copied over. This might be a very rare case but I am afraid this will happen. So in my opinion we should lock the access to the list, create a copy and then iterate over the copy. All other actions like removal or adding must also be locked.

wdichler:
Another solution would be an ImmutableArray. Which you would prefer?

chkr1011:
I am wondering how to get from the mutable list to an immutable one. This always requires locking in my opinion. I will create a branch to show exactly what my idea is.

chkr1011:
@wdichler Please checkout https://github.com/dotnet/MQTTnet/pull/1515. I found a solution which is lock less when reading the handlers and will not create a copy every time.

PR #1506, opened 2022-09-06
Fix: Send DISCONNECT packets to MQTT5 clients only

### LINKED ISSUE #1505: Server should not send DISCONNECT packages to MQTT 3.1.1 clients
### Describe the bug

When the server disconnects a client (i.e. via `DisconnectClientAsync`) with reason `SessionTakenOver` or `KeepAliveTimeout` then a DISCONNECT package is sent to clients. The MQTT 3.1.1 specification only mentions that DISCONNECT packages are sent from the client to the server (see section 3.14). The server therefore should not send DISCONNECT packages to MQTT 3.1.1 clients (this is a bug I believe). 

- Server

### To Reproduce

This test detects whether a DISCONNECT package is received for an MQTT 3.1.1 client; none should be received.
```
        [TestMethod]
        public async Task Disconnect_Client_with_Reason()
        {
            using (var testEnvironment = CreateTestEnvironment())
            {
                var disconnectPacketReceived = false;

                string testClientId = null;

                testEnvironment.Server.ClientConnectedAsync += e =>
                {
                    testClientId = e.ClientId;
                    return PlatformAbstractionLayer.CompletedTask;
                };

                var client = testEnvironment.CreateClient();

                client.InspectPackage += e =>
                {
                    if (e.Buffer.Length > 0)
                    {
                        if (e.Buffer[0] == (byte)Protocol.MqttControlPacketType.Disconnect << 4)
                        {
                            disconnectPacketReceived = true;
                        }
                    }
                    return PlatformAbstractionLayer.CompletedTask;
                };

                await client.ConnectAsync(
                    new MqttClientOptionsBuilder().WithTcpServer("localhost", testEnvironment.ServerPort)
                    .WithProtocolVersion(MQTTnet.Formatter.MqttProtocolVersion.V311)
                    .Build()
                    );

                await LongTestDelay();

                // Test client should be connected now

                Assert.IsTrue(testClientId != null);

                // Have the server disconnect the client with SessionTakenOver reason; no DISCONNECT should be sent to MQTT 3.1.1 clients

                await testEnvironment.Server.DisconnectClientAsync(testClientId, Protocol.MqttDisconnectReasonCode.SessionTakenOver);

                await LongTestDelay();

                // MQTT 3.1.1 client should not receive a disconnect packet

                Assert.IsFalse(disconnectPacketReceived, "Disconnect packet received when none is expected");
            }
        }
```

### Expected behavior

DISCONNECT packages should not be sent to MQTT 3.1.1 clients. Other disconnect reasons should be sent to MQTT 5 clients (this would be a new feature because it is not mandatory).

The fix would be simple, i.e. in `MqttClient.cs` in the `Server\Internal` folder the condition could be changed from

```
if (reason == MqttDisconnectReasonCode.SessionTakenOver || reason == MqttDisconnectReasonCode.KeepAliveTimeout)  
{
   ...
```
to

```
if (ChannelAdapter.PacketFormatterAdapter.ProtocolVersion == MqttProtocolVersion.V500 && reason != MqttDisconnectReasonCode.NormalDisconnection)
{
    ....
```

I'll add a pull request soon.

logicaloud:
The server should not send DISCONNECT packets to MQTT 3.1.1 clients.

Fixes #1505 and sends disconnect reasons other than "NormalDisconnect" to MQTT5 clients.

### DIFF
--- a/Source/MQTTnet.Tests/MQTTv5/Server_Tests.cs
+++ b/Source/MQTTnet.Tests/MQTTv5/Server_Tests.cs
@@ -175,6 +175,49 @@ namespace MQTTnet.Tests.MQTTv5
             }
         }
 
+        [TestMethod]
+        public async Task Disconnect_with_Reason()
+        {
+            using (var testEnvironment = CreateTestEnvironment())
+            {
+                var disconnectReason = MQTTnet.Client.MqttClientDisconnectReason.UnspecifiedError;
+
+                string testClientId = null;
+
+                await testEnvironment.StartServer();
+
+                testEnvironment.Server.ClientConnectedAsync += e =>
+                {
+                    testClientId = e.ClientId;
+                    return PlatformAbstractionLayer.CompletedTask;
+                };
+
+                var client = await testEnvironment.ConnectClient(new MQTTnet.Client.MqttClientOptionsBuilder().WithProtocolVersion(MQTTnet.Formatter.MqttProtocolVersion.V500));
+
+                client.DisconnectedAsync += e =>
+                {
+                    disconnectReason = e.Reason;
+                    return PlatformAbstractionLayer.CompletedTask;
+                };
+
+                await LongTestDelay();
+
+                // Test client should be connected now
+
+                Assert.IsTrue(testClientId != null);
+
+                // Have the server disconnect the client with AdministrativeAction reason
+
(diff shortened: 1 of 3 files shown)


chkr1011:
Thanks for your contribution. Good work 👍

PR #1515, opened 2022-09-13
Fix cross access to async events.

chkr1011:
This pull request adds support for adding and removing handlers of async events from different threads at the same time.

### DIFF
--- a/Source/MQTTnet/Internal/AsyncEvent.cs
+++ b/Source/MQTTnet/Internal/AsyncEvent.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using MQTTnet.Diagnostics;
 
@@ -13,7 +14,16 @@ namespace MQTTnet.Internal
     {
         readonly List<AsyncEventInvocator<TEventArgs>> _handlers = new List<AsyncEventInvocator<TEventArgs>>();
 
-        public bool HasHandlers => _handlers.Count > 0;
+        ICollection<AsyncEventInvocator<TEventArgs>> _handlersForInvoke;
+
+        public AsyncEvent()
+        {
+            _handlersForInvoke = _handlers;
+        }
+
+        // Track the existence of handlers in a separate field so that checking it all the time will not
+        // require locking the actual list (_handlers).
+        public bool HasHandlers { get; private set; }
 
         public void AddHandler(Func<TEventArgs, Task> handler)
         {
@@ -22,7 +32,13 @@ namespace MQTTnet.Internal
                 throw new ArgumentNullException(nameof(handler));
             }
 
-            _handlers.Add(new AsyncEventInvocator<TEventArgs>(null, handler));
+            lock (_handlers)
+            {
+                _handlers.Add(new AsyncEventInvocator<TEventArgs>(null, handler));
+
+                HasHandlers = true;
+                _handlersForInvoke = new List<AsyncEventInvocator<TEventArgs>>(_handlers.ToList());
+            }
         }
 
         public void AddHandler(Action<TEventArgs> handler)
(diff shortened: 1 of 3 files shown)


Source/MQTTnet/Internal/AsyncEvent.cs
```diff
@@ -13,7 +14,16 @@ public sealed class AsyncEvent<TEventArgs> where TEventArgs : EventArgs
     {
         readonly List<AsyncEventInvocator<TEventArgs>> _handlers = new List<AsyncEventInvocator<TEventArgs>>();
 
-        public bool HasHandlers => _handlers.Count > 0;
+        ICollection<AsyncEventInvocator<TEventArgs>> _handlersForInvoke;
```
gfoidl, line 17: Is this field necessary or can `_handlers` be used instead?
It looks quite redundant (and allocating) to me -- it's basically just `_handlers.ToList()`, so `_handlers` could be used directly?

Source/MQTTnet/Internal/AsyncEvent.cs
```diff
@@ -32,25 +48,46 @@ public void AddHandler(Action<TEventArgs> handler)
...
+            }
+
+            // Adding or removing handlers will produce a new list instance all the time.
+            // So locking here is not required since only the reference to an immutable list
+            // of handlers is used.
+            var handlers = _handlersForInvoke;
+            foreach (var handler in handlers)
             {
                 await handler.InvokeAsync(eventArgs).ConfigureAwait(false);
             }
```
gfoidl, line 74: Not related to the PR, just as idea:
Would something like be possible and show some perf-benefits?
```c#
[ThreadStatic]
private static List<Task> t_tasksCache;
// ...
var handlers = _handlersForInvoke;
List<Task> tasks = t_taskCache ?? new List<Task>(capacity: handlers .Count);
foreach (var handler in handlers)
{
    tasks.Add(handler.InvokeAsync(eventArgs));
}
await Task.WhenAll(tasks).ConfigureAwait(false);
tasks.Clear();
t_taskCache = tasks;
```
So issue all invokes at once, then wait for them at once instead of individually waiting.
To avoid the allocation of the list to hold the tasks, a thread-static used here -- but there may be better pooling / caching for this.

Source/MQTTnet/Internal/AsyncEvent.cs
chkr1011: My idea is that the second list is build when changing the handlers inside a lock. To avoid the locking in the read access I decided to create an immutable (not really immutable but it never gets changes after initial creation) copy which is used for read access. Since this involves only a reference swap there is no need for a lock.

Source/MQTTnet/Internal/AsyncEvent.cs
chkr1011: I don't like the idea because events are usually called in serial. There is also a way to pass data across handlers like "IsHandled" properties etc. When calling all at once in separate tasks the business logic may end up in race conditions. If this behavior is wanted users may just start a new Task in the event handler.

gfoidl: Makes sense.

Source/MQTTnet/Internal/AsyncEvent.cs
gfoidl: 👍🏻 

Is adding / removing done quite often?
Question is because of the allocation of the "immutable" list, and if it makes sense to avoid that repeated allocation in some way.
If it's not done often, the code as is, is fine and clean to read.

chkr1011: I don't know how often this might be done. This change comes from another pull request where Rx is used together with this library which will add and remove handlers from different threads. My goal was to keep reading the handlers as fast as possible because I assume that changing the handlers is not done very often in most use cases.

PR #1520, opened 2022-09-16
Improve async lock

chkr1011:
This pull request adds a new implementation of the Async Lock which has no longer a dead lock issue.


logicaloud:
Should the `_isDisposed` check in `WaitAsync` move inside the `_syncRoot` lock a bit further down?

chkr1011:
> Should the `_isDisposed` check in `WaitAsync` move inside the `_syncRoot` lock a bit further down?

Why do you think so? The field is atomic to there is no need to lock first in my opinion.

logicaloud:
I was thinking about multithreaded access where the bool is modified in one thread, then sits in a register and another thread reads a stale value. Maybe the bool could be declared as `volatile` or moved into the lock (https://learn.microsoft.com/en-us/archive/msdn-magazine/2012/december/csharp-the-csharp-memory-model-in-theory-and-practice#thread-communication-patterns). I'm not sure if it would make much of a difference in practice.

chkr1011:
Makes sense. I will move it for sure.

PR #1583, opened 2022-11-24
Set string.Empty as default value for SenderClientId

xljiulang:
As me konw, when the property `SenderClientId ` of `InjectedMqttApplicationMessage` type is null, `MqttServer. InjectApplicationMessage (InjectedMqttApplicationMessage)` will throws an `ArgumentNullException`. Unfortunately,  `MqttServerExtensions.cs` provide extension methods without the SenderClientId parameter, it allway throw `ArgumentNullException`.

```c#
public static Task InjectApplicationMessage(
    this MqttServer server,
    string topic,
    string payload = null,
    MqttQualityOfServiceLevel qualityOfServiceLevel = MqttQualityOfServiceLevel.AtMostOnce,
    bool retain = false)
```
So we might as well give the `SenderClientId`  property to set the default value of `string.Empty`


chkr1011:

Issue #1577, opened 2022-11-21
Use ReadonlyMemory replace byte Array


xljiulang:
Hello, can we change the `byte[]` type of the `Payload` to `ReadonlyMemory<byte>` so that it can be used together with `ArrayPool<byte>` to reduce GC pressure?

eg.
```c#
var message = new GoogleMessage();
var size = message.CalculateSize();
Debug.Assert(960 == size);

var buffer = ArrayPool<byte>.Shared.Rent(size);
Debug.Assert(1024 == buffer.Length);

var output = new CodedOutputStream(buffer);
message.WriteTo(output);
var payload = buffer.AsMemory(0, size);
// mqtt.SendAsync(payloadPacket)

ArrayPool<byte>.Shared.Return(buffer);
```

chkr1011:
I assume that you can already make use of the Array Pool. It returns a simple byte array which can be already used as the payload.

xljiulang:
@chkr1011 
However, the length of the rent `byte[]` is not equal to the length of the valid data(1024 != 960), which cannot be directly used as the payload. Instead, i need to copy the first 960 bytes of the rent `byte[]` as the payload, which is no different from new byte[960].

gfoidl:
> there is no disruption to the user

Strictly speaking this in an API-breaking change (different type).

`ROM<byte>` could be exposed as additional member, e.g. as `PayloadMemory` and the current `Payload` just forwards to `PayloadMemory` internally. In .NET the SocketsAwaitable uses as similar approach.

chkr1011:
We cannot change the type to Memory etc. because we still need to support older frameworks like 4.5.2, UWP etc.

I have a proposal on how to solve this issue. In my opinion the only problem is that there is currently no chance to pass the bounds of the payload so that a larger array from the pool can be reused etc.

So, I will add two new properties. 

**PayloadOffset** is the offset of the payload which will be initialized with **0** which is correct for all cases and causes no breaking changes.

**PayloadLength** is the length which should be used when accessing the payload array. This will be initialized with **int.MaxValue** which indicates (as a magic number) that the value from the payload array should be used by default. This approach is a little bit ugly but will keep the API stable without breaking changes. If the length is set, the value is used. Probably a bounds check can be made so that a higher value than the actual array size of the payload indicates that the full array should be used.

These two new properties will be properly filled for received messages but without any benefit because they will be **0** and the actual length of the payload all the time.

[...]

PR #1585, opened 2022-11-25
Add PayloadSegmentable feature

xljiulang:
### What's new
1. Add interface `IPayloadSegmentable`
2. Add `IPayloadSegmentable` extensions class `PayloadSegmentExtensions`
3. Let `MqttApplicationMessage` and `MqttPublishPacket` inherit from  `IPayloadSegmentable`
4. Refactoring the `WithPayload`  methods of `MqttApplicationMessageBuilder`
5. Add `WithPayload(ReadOnlyMemory<byte>)`  and `WithPayload(ArraySegment<byte>)`
6. Refactoring  `Encode(MqttPacket mqttPacket)`  of  `MqttPacketFormatter`

### Api Design
```c#
public interface IPayloadSegmentable
{
    /// <summary>
    /// Get or set Payload array
    /// </summary>
    byte[] Payload { get; set; }

    /// <summary>
    /// Get or set the offset of Payload
    /// </summary>
    int PayloadOffset { get; set; }

    /// <summary>
    /// Get or set the effective number of bytes of Payload
    /// Leaving null means equal to the total length of Payload minus PayloadOffset
    /// </summary>
    int? PayloadCount { get; set; }
}
```
```c#
public static class PayloadSegmentExtensions
{
    /// <summary>
    /// Get the ArraySegment style of Payload
    /// </summary>
    /// <param name="segmentable"></param>
    /// <returns></returns>
    public static ArraySegment<byte> GetPayloadSegment(this IPayloadSegmentable segmentable)
    {
          ...
    }

    /// <summary>
    /// Set payloadSegment to Payload, PayloadOffset and PayloadCount
    /// </summary>
    /// <param name="segmentable"></param>
    /// <param name="payloadSegment"></param>
    public static void SetPayloadSegment(this IPayloadSegmentable segmentable, ArraySegment<byte> payloadSegment)
    {
          ...
    }

#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_1
    /// <summary>
    /// Set payloadSegment to Payload, PayloadOffset and PayloadCount
    /// </summary>
    /// <param name="segmentable"></param>
    /// <param name="payloadSegment"></param>
    public static void SetPayloadSegment(this IPayloadSegmentable segmentable, ReadOnlyMemory<byte> payloadSegment)
    {
         ...
    }
#endif

}
```


Source/MQTTnet/IPayloadSegmentable.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <summary>
+        /// Get or set the offset of Payload
+        /// </summary>
+        int PayloadOffset { get; set; }
+
+        /// <summary>
+        /// Get or set the effective number of bytes of Payload
+        /// Leaving null means equal to the total length of Payload minus PayloadOffset
+        /// </summary>
+        int? PayloadCount { get; set; }
```
gfoidl, line 19:
```suggestion
        int? PayloadLength { get; set; }
```
is more .NET-like. I associated _Count_ with enumerating, so "how many payloads are there?" not as "how **long** is the payload".

Source/MQTTnet/MqttApplicationMessage.cs
```diff
@@ -48,11 +49,21 @@ public sealed class MqttApplicationMessage
...
+        /// <summary>
+        /// Get or set the offset of Payload
+        /// </summary>
+        public int PayloadOffset { get; set; }
+
+        /// <summary>
+        /// Get or set the effective number of bytes of Payload
+        /// Leaving null means equal to the total length of Payload minus PayloadOffset
+        /// </summary>
+        public int? PayloadCount { get; set; }
```
gfoidl, line 65: When new members are added to this type, then there could also be
```c#
public ReadOnlySpan<byte> Span => Payload.AsSpan(PayloadOffset, PayloadLength);
```
as convenience property to get the span of the payload.

Or is a ROM more suitable? Or both?
Then it's too much and just leave it out?

Source/MQTTnet/MqttApplicationMessageBuilder.cs
```diff
@@ -112,76 +119,89 @@ public MqttApplicationMessageBuilder WithMessageExpiryInterval(uint messageExpir
...
             }
 
-            if (payload.Length == 0)
-            {
-                _payload = null;
-            }
-            else
+            var payloadBuffer = new byte[length];
+            var totalRead = 0;
+            do
```
gfoidl, line 171: FYI: for .NET 7 onwards `ReadExactly` can be used. But ATM there's no such target https://github.com/dotnet/MQTTnet/blob/7f4a7ef668c15ba968a9950cca4898bf47a776c1/Source/MQTTnet/MQTTnet.csproj#L13

Source/MQTTnet/PayloadSegmentExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="segmentable"></param>
+        /// <returns></returns>
+        public static ArraySegment<byte> GetPayloadSegment(this IPayloadSegmentable segmentable)
+        {
+            if (segmentable.Payload == null)
+            {
+                return emptySegment;
+            }
+
+            var payloadCount = segmentable.PayloadCount == null
```
gfoidl, line 23:
```suggestion
            var payloadLength = segmentable.PayloadLength == null
```
and all other places as well.

Source/MQTTnet/Server/Internal/MqttRetainedMessagesManager.cs
```diff
@@ -135,5 +137,14 @@ public async Task ClearMessages()
                 await _eventContainer.RetainedMessagesClearedEvent.InvokeAsync(EventArgs.Empty).ConfigureAwait(false);
             }
         }
+
+        private static bool SequenceEqual(ArraySegment<byte> source, ArraySegment<byte> target)
+        {
+#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_1
+            return source.Count == target.Count && source.AsSpan().SequenceEqual(target);
```
gfoidl, line 144:
```suggestion
            return source.AsSpan().SequenceEqual(target);
```
`SequenceEqual` also does a length check before doing the actual compare, so no need to have that check twice.

gfoidl:
Please add a reference to https://github.com/dotnet/MQTTnet/issues/1577 in the opening comment, too.

Source/MQTTnet/IPayloadSegmentable.cs
xljiulang: Yes, I agree with you

Source/MQTTnet/MqttApplicationMessageBuilder.cs
xljiulang: Yes, but the original logic is retained here, just modified to be compatible with the new PayloadOffset and PayloadLength propties.

Source/MQTTnet/Formatter/MqttPublishPacketFactory.cs
```diff
@@ -17,6 +17,8 @@ public MqttPublishPacket Clone(MqttPublishPacket publishPacket)
             {
                 Topic = publishPacket.Topic,
                 Payload = publishPacket.Payload,
+                PayloadOffset = publishPacket.PayloadOffset,
+                PayloadLength = publishPacket.PayloadLength,
```
chkr1011, line 21: We have a problem when the value is not null here because several interceptors in the server allow for changing the payload. If the length is present in these cases existing code will fail because now the users have to update the length property accordingly.

xljiulang: So, what should we do here? Or we can design two fields, `byte[] Payload` and `ArraySegment<byte> PayloadSegment`. When set the value of one field, it affects the obtaining of the other

xljiulang:
I believe that simply adding an `public ArraySegment<byte> PayloadSegment`  property will solve the compatibility problem.
```c#
public sealed class MqttApplicationMessage
{
    private byte[] _payloadCache = null;
    private ArraySegment<byte> _payloadSegment = EmptyBuffer.ArraySegment;

    /// <summary>
    /// Gets or sets the payload.
    /// The payload is the data bytes sent via the MQTT protocol.
    /// </summary>
    public byte[] Payload
    {
        get
        {
            // just reference from _payloadSegment.Array
            if (_payloadSegment.Count == _payloadSegment.Array.Length)
            {
                return _payloadSegment.Array;
            }

            // copy from _payloadSegment
            if (_payloadCache == null)
            {
                _payloadCache = new byte[_payloadSegment.Count];
                Array.Copy(_payloadSegment.Array, _payloadSegment.Offset, _payloadCache, 0, _payloadCache.Length);
            }
            return _payloadCache;
        }
        set
        {
            _payloadCache = null;
            _payloadSegment = value == null || value.Length == 0
                ? EmptyBuffer.ArraySegment
                : new ArraySegment<byte>(value);
        }
    }

    /// <summary>
    /// Get or set ArraySegment style of Payload.
    /// </summary>
    public ArraySegment<byte> PayloadSegment
    {
        get
        {
            return _payloadSegment;
        }
        set
        {
            _payloadCache = null;
            _payloadSegment = value;
        }
    }
}
```

xljiulang:
@chkr1011 
The second modification of `PayloadSegment` has been completed, please review!

xljiulang:
I'd like to know what everyone thinks about this request. We need to be able to send `ArraySegment` payload that can be allocated infinitely close to zero.

chkr1011:
@xljiulang I am busy with finalizing version 4.1.4. I consider this as a new feature that it will be released with 4.2.x. I will review this feature after 4.1.4 is out.

chkr1011 (changes requested):

chkr1011:
@xljiulang I made the requested changes and pushed them. Please let me know if you agree with them. If so I will merge the branch.

xljiulang:
good work!

Issue #1584, opened 2022-11-24
Memory leak inside AsyncSignal / AsyncSignalWaiter


mario-zelger:
### Describe the bug
When calling the `Set()` Method of an `AsyncSignal` Object the `AsyncSignalWaiter` is not correctly disposed:
```csharp
public void Set()
{
    lock (_syncRoot)
    {
        _isSignaled = true;

        Cleanup();
        
        // If there is already a waiting task let it run.
        if (_waiter != null)
        {
            _waiter.Approve();
            _waiter = null; // Waiter is never disposed

            // Since we already got a waiter the signal must be reset right now!
            _isSignaled = false;
        }
    }
}
```

When I dispose the `AsyncSignalWaiter` after the approval inside `Set()` or inside the `Approve()` or `Cancel()` method of the `AsyncSignalWaiter` no memory leaks can be found (see screenshots).
```csharp
public void Set()
{
    lock (_syncRoot)
    {
        _isSignaled = true;

        Cleanup();
        
        // If there is already a waiting task let it run.
        if (_waiter != null)
        {
            _waiter.Approve();
            _waiter.Dispose(); // Dispose waiter
            _waiter = null;

            // Since we already got a waiter the signal must be reset right now!
            _isSignaled = false;
        }
    }
}
```
```csharp
public void Approve()
{
    _promise.TrySetResult(true);
    Dispose(); // Dispose itself
}

public void Dispose()
{
    _cancellationTokenRegistration.Dispose();

    _promise.TrySetException(new ObjectDisposedException(nameof(AsyncSignalWaiter)));
}

void Cancel()
{
    _promise.TrySetCanceled();
    Dispose();  // Dispose itself
}
```

### Which component is your bug related to?
- Client
- ManagedClient
- RpcClient (unsure)

### To Reproduce
1. Using version `4.1.3.436` of MQTTnet or the `master` branch.
2. Run a simple console app with either a `Client` or `ManagedClient` (see code below).
3. Run a memory profiler for some time (JetBrains dotMemory in my case)
4. Watch Gen 2 Heap growing and growing.

### Expected behavior
The `CancellationTokenRegistration` inside the `AsyncSignalWaiter` should always be disposed.

### Screenshots
Without disposing the `AsyncSignalWaiter`
![image](https://user-images.githubusercontent.com/15825982/203858029-3e0ae89c-4fb0-4631-9502-4ec8c7227183.png)

With disposing of the `AsyncSignalWaiter`
[...]

PR #1586, opened 2022-11-25
Dispose the AsyncSignalWaiter before setting it to null

mario-zelger:
Fixes a memory leak within `AsyncSignalWaiter` when setting the signal on an `AsyncSignal` instance. See #1584 for more details.


chkr1011:
Thanks for your contribution. Good work 👍

PR #1590, opened 2022-11-26
Avoid some full memory barriers

gfoidl:
`Interlocked.Read/Exchange` use full memory-barriers and have higher cost than release/acquire fences (`Volatile.Read/Write`). So change some uses of Interlocked to Volatile -- where possible and where the substituion is safe in terms of visibility.

It's hard to show the improvement in micro-benchmarks, one would need a real scenario to measure throughput, etc. I haven't done this.


Source/MQTTnet/Adapter/MqttChannelAdapter.cs
```diff
@@ -33,8 +33,7 @@ public sealed class MqttChannelAdapter : Disposable, IMqttChannelAdapter
 
         readonly AsyncLock _syncRoot = new AsyncLock();
 
-        long _bytesReceived;
-        long _bytesSent;
+        Statistics _statistics;     // mutable struct, don't make readonly!
```
gfoidl, line 36: Statistics moved to separate struct. Perf-wise there's no difference, only for hygiene.
And if there might be some better way to reset* in the future, it's only one place to update.

\* e.g. if the types were `int` instead of `long` one could use a "union" of the two ints overlayed with a long, then the reset would just be a volatile write to that long with 0.
Note: I don't propose to change to int, just as example.

Source/MQTTnet/Internal/AsyncQueue.cs
```diff
@@ -22,7 +22,11 @@ public sealed class AsyncQueue<TItem> : IDisposable
 
         public void Clear()
         {
+#if NETCOREAPP3_1_OR_GREATER
+            _queue.Clear();
```
gfoidl, line 26: It's cheaper and avoids another allocation.

Note: internally CQ.Clear just drops the segments and creates new ones atomically. So there's an allocation anyway, but we save the one for the CQ itself.

Source/MQTTnet/Internal/AsyncQueue.cs
```diff
@@ -32,6 +36,16 @@ public void Dispose()
                 _signal.Dispose();
 
                 _isDisposed = true;
+
+#if !NETSTANDARD1_3
+                if (typeof(IDisposable).IsAssignableFrom(typeof(TItem)))
```
gfoidl, line 41: For the use within MQTTnet not needed, as no `TItem` implements `IDisposable`, but this type is `public` so it should dispose of all items in the queue.

For NS1.3 there's no easy way to perform that check upfront, only within the loop, so I skipped it for that target. It wasn't done before, so there's no regression for NS1.3.

gfoidl (commented):
Some notes for review.

Source/MQTTnet/Internal/AsyncQueue.cs
damikun: This is nor related to pull but to point about NS.... Why is NS1.3 even still supported isnt better to drop that support in next major release at all ? I know there can be legacy but still let have a look on what is TLS and supported by ofiicial microsoft releases...

chkr1011: @damikun  This goal of this library is that it runs on most .NET versions. Especially because it covers a hardware topic. There are quite old machine out there which may run very old frameworks like 4.5.2 on XP machines. They cannot easily upgrade because they run on very special hardware. I also know people which use the library on a Pi 2 with an old version of UWP. I also wanted to support .NET 3.5 (which has support until 2029) but I requires too much changes.

PR #1593, opened 2022-11-28
1592 bug of mqttbufferreader

chkr1011:


Source/MQTTnet/Formatter/MqttBufferReader.cs
```diff
@@ -45,32 +44,36 @@ public byte ReadByte()
...
 
         public uint ReadFourByteInteger()
         {
             ValidateReceiveBuffer(4);
 
-            var byte0 = _buffer[_position++];
-            var byte1 = _buffer[_position++];
-            var byte2 = _buffer[_position++];
-            var byte3 = _buffer[_position++];
+            var byte0 = _buffer[_offset + Position];
```
gfoidl, line 54: FYI: for "modern targets" this could be done via `BinaryPrimitives.ReadUInt32BigEndian`-method.

Source/MQTTnet/Formatter/MqttBufferWriter.cs
```diff
@@ -63,17 +63,24 @@ public byte[] GetBuffer()
...
+            if (value <= 127)
             {
-                x /= 128;
-                result++;
-            } while (x > 0);
+                return 1;
+            }
 
-            return result;
+            if (value <= 16383)
```
gfoidl, line 73: Can this be expressed as `0x7F`, `0x3F_FF`, `0x1F_FF_FF`? To make it a bit more obvious where these constants come from? The old algorithm was clearer on that regard.
Or as comment just a link to the spec that dictates that encoding.

gfoidl: Also when I read https://github.com/dotnet/MQTTnet/blob/00481533c13560f04330bcfb4804f1ba90174293/Source/MQTTnet.AspnetCore/ReaderExtensions.cs#L107 correct and as I would assume only the most significant bit is relevant. 
With the constants used it's 1, 2, 3 bits that are "reserved for continuation", while only 1 is needed?
So the constants should be `0x7F`, `0x7F_FF`, `0x7F_FF_FF`? So bigger ranges can be encoded with less bytes used. Or I'm missing something...

Source/MQTTnet/Formatter/MqttBufferReader.cs
chkr1011: I adopted this but enforced the reverse order. With the method you mentioned it depends on the CPU.

Source/MQTTnet/Formatter/MqttBufferWriter.cs
chkr1011: I imported some documentation from the RFC to explain the values.

I do not really unterstand what you are referring to in the _ReaderExtensions_.

chkr1011: I will merge this PR so that I can release a hotfix version soon. We can discuss this in another branch.

Source/MQTTnet/Formatter/MqttBufferReader.cs
gfoidl: Ah, right that "cleans up" the order automatically. `MemoryMarshal.ReadXYZ` would be the correct one. 
The gain is one memory read instead of the individual bytes (which wil most likely be in the same cacheline). Effect will be only 1 bound-check instead of the 4.

Source/MQTTnet/Formatter/MqttBufferWriter.cs
gfoidl: > I do not really unterstand what you are referring to in the ReaderExtensions.

Sorry, I mixed the plain and the encoding threshoulds.

Issue #1592, opened 2022-11-26
Bug of MqttBufferReader


xljiulang:
### Verification
```c#
var buffer1 = new byte[] { 5, 6, 7, 8, 9 };
var buffer2 = new byte[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };

var segment1 = new ArraySegment<byte>(buffer1);
var segment2 = new ArraySegment<byte>(buffer2, 5, 5);
Debug.Assert(segment1.SequenceEqual(segment2));

var reader1 = new MqttBufferReader();
reader1.SetBuffer(segment1.Array, segment1.Offset, segment1.Count);

var reader2 = new MqttBufferReader();
reader2.SetBuffer(segment2.Array, segment2.Offset, segment2.Count);

// Assert failure here!!!
// reader1.BytesLeft == 5
// reader2.BytesLeft == 0
Debug.Assert(reader1.BytesLeft == reader2.BytesLeft);
```

### Describe the bug
I think it is probably a bug caused by the ambiguity of the `length` and `_length` in it, so what does `length` `_length` really mean?
Personally, I think the following understanding may be more accurate:

#### Parameter `length`
The number of elements in the range delimited by the `ArraySegment`.  It's value == arraySegment.Count

#### Field `_length`
The max position of input buffer,  It's value == arraySegment.Count + arraySegment.Offset

If we don't want to change the name of the field or parameter, here's how to fix the bug:
```c#
  public void SetBuffer(byte[] buffer, int position, int length)
  {
      _buffer = buffer;
      _initialPosition = position;
      _position = position;
      _length = length;   ==>   _length = position + length;
  }
```

But I prefer to add a new method，and then I don't have to think about whether the `length`  and `_length` is `ArraySegment.Array.Length` or `ArraySegment.Count`:

```
public void SetBuffer(ArraySegment<byte> buffer)
```

xljiulang:
If we rename `_length` to `_maxPosition`, the problem would be easier to understand.

chkr1011:
Thanks for reporting the issue. I attached a branch to this ticket and fixed the issue.
I also added some unit test according to your sample code.

Please have a look and share your thoughts. I will still add some more tests.

simonthum:
I'm looking into a quite strange issue with MQTTnet not receiving messages on Mosquitto 2.x but receiving fine with 1.6.x. All other clients I use work fine.

Could this be related?

chkr1011:
[...]

PR #1596, opened 2022-11-29
Avoid copying buffer from ReadOnlySequence as  ReceivedMqttPacket body

xljiulang:
### Related project
MQTTnet.AspNetCore

### How it affects
Read `ArraySegment<byte>` buffer from ReadOnlySequence as ReceivedMqttPacket body. 
It can greatly reduce unnecessary buffer-copying.

### Dependency
This request depend on #1592，after we make MqttBufferReader correctly support `ArraySegment<byte>`


xljiulang:
Also depend on #1593

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
```diff
@@ -81,23 +83,24 @@ static bool TryReadBodyLength(ref ReadOnlySequence<byte> input, out int headerLe
             var index = 1;
             headerLength = 0;
             bodyLength = 0;
-            
-            var temp = input.Slice(0, Math.Min(5, input.Length)).GetMemory().Span;
+
+            var valueSequence = input.Slice(0, Math.Min(5, input.Length));
+            var valueSpan = valueSequence.IsSingleSegment ? valueSequence.First.Span : valueSequence.ToArray();
```
gfoidl, line 88:
```suggestion
            var valueSpan = valueSequence.IsSingleSegment ? valueSequence.FirstSpan : valueSequence.ToArray();
```

gfoidl, line 88: `valueSequence.Length == 5` so rather short, and lets assume that `valueSequence.IsSingleSegment == true / false` is both likely (though == true will be more likely) so `ToArray` will be expensive.

```c#
Span<byte> valueSpan = stackalloc byte[8];    // 8 as next power of 2
valueSequence.CopyTo(valueSpan);
valueSpan = valueSpan.Slice(0, valueSequence.Length);
```

`FirstSpan` is also not really cheap, so I'd just straighten the code and copy always. It's only a few bytes, but the code is much shorter so it should outweight the cost of copying.

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
```diff
@@ -81,23 +83,24 @@ static bool TryReadBodyLength(ref ReadOnlySequence<byte> input, out int headerLe
...
 
-                encodedByte = temp[index];
+                encodedByte = valueSpan[index];
                 index++;
 
                 value += (byte)(encodedByte & 127) * multiplier;
                 if (multiplier > 128 * 128 * 128)
                 {
-                    throw new MqttProtocolViolationException($"Remaining length is invalid (Data={string.Join(",", temp.Slice(1, index).ToArray())}).");
+                    throw new MqttProtocolViolationException($"Remaining length is invalid (Data={string.Join(",", valueSpan.Slice(1, index).ToArray())}).");
```
gfoidl, line 103: This check should be moved down after L106 where `multiplier` is updated.
Then as this materializes a costly string it should use a throw-helper. Like
```c#
// L105
multiplier *= 128;

if (multiplier > 128 * 128 * 128)
{
    Throw(valueSpan, index);

    static void Throw(ReadOnlySpan<byte> valueSpan, int index)
        => throw new MqttProtocolViolationException($"Remaining length is invalid (Data={string.Join(",", valueSpan.Slice(1, index).ToArray())}).");
}
```
This keeps the loop-body short (from CPU-view), thus faster.

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
xljiulang: Benefit a lot！

xljiulang: We cannot use `FirstSpan` Api under some target frameworks, such as `netcoreapp` and `netstandard`

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
```diff
@@ -61,15 +62,16 @@ public static class ReaderExtensions
             return true;
         }
 
-        static ReadOnlyMemory<byte> GetMemory(this in ReadOnlySequence<byte> input)
+        static ArraySegment<byte> AsArraySegment(ref ReadOnlySequence<byte> input)
```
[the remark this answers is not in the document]
xljiulang: There cannot be interpreted as `ReadOnlyMemory` - > `ArraySegment`, but `ReadOnlyMemory.ToArray()` - > `MemoryMarshal. TryGetArray(ReadOnlyMemory, out ReadOnlyMemory)`

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
xljiulang: Great!!

Source/MQTTnet.AspnetCore/ReaderExtensions.cs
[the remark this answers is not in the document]
gfoidl: Yep, I just mean there should be a benchmark showing that it doesn't get worse. I'm uncertain in which way the needle will move, so just a measure of caution. I'm sure you have this on your table 😄.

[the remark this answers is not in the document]
xljiulang: Well, I will provide a Benchmark as soon as possible.

xljiulang:
The benchmark result shows a good reduction in memory allocation!
``` ini

BenchmarkDotNet=v0.13.2, OS=Windows 10 (10.0.19044.2251/21H2/November2021Update)
Intel Core i3-4150 CPU 3.50GHz (Haswell), 1 CPU, 4 logical and 2 physical cores
.NET SDK=7.0.100
  [Host]   : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT AVX2
  .NET 6.0 : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT AVX2

Job=.NET 6.0  Runtime=.NET 6.0  

```
| Method |     Mean |     Error |    StdDev | Ratio | Rank |    Gen0 | Allocated | Alloc Ratio |
|------- |---------:|----------:|----------:|------:|-----:|--------:|----------:|------------:|
| Before | 3.593 μs | 0.0124 μs | 0.0097 μs |  1.00 |    2 | 13.4201 |  20.57 KB |        1.00 |
|  After | 2.536 μs | 0.0151 μs | 0.0134 μs |  0.71 |    1 |  6.8932 |  10.54 KB |        0.51 |

Issue #1597, opened 2022-11-30
New Bug of MqttBufferReader


xljiulang:
### Bug1
There are two logical problems here:
1. An error will occur when `_offset` is not 0
2. An error will occur in Big-Endian platform
```c#
 #if NETCOREAPP3_0_OR_GREATER
            var value = System.Runtime.InteropServices.MemoryMarshal.Read<ushort>(_buffer.AsSpan().Slice(Position));
            value = System.Buffers.Binary.BinaryPrimitives.ReverseEndianness(value);
#else
            var msb = _buffer[_offset + Position];
            var lsb = _buffer[_offset + Position + 1];

            var value = (ushort)((msb << 8) | lsb);
#endif
```

fix code:
```c#
 #if NETCOREAPP3_0_OR_GREATER
            var value = MemoryMarshal.Read<ushort>(_buffer.AsSpan(_offset + Position));
            if (BitConverter.IsLittleEndian)
            {
                value = System.Buffers.Binary.BinaryPrimitives.ReverseEndianness(value);
            }
#else
            var msb = _buffer[_offset + Position];
            var lsb = _buffer[_offset + Position + 1];

            var value = (ushort)((msb << 8) | lsb);
#endif
```

### Bug2
_maxPosition  is incorrect when offset is not 0
```c#
        public void SetBuffer(byte[] buffer, int offset, int length)
        {
            _buffer = buffer ?? throw new ArgumentNullException(nameof(buffer));
            _offset = offset;

            Position = 0;
            _maxPosition = length;
        }
```

fix code:
```c#
        public void SetBuffer(byte[] buffer, int offset, int length)
        {
            _buffer = buffer ?? throw new ArgumentNullException(nameof(buffer));
            _offset = offset;

            Position = 0;
            _maxPosition = offset+ length;
        }
```

xljiulang:
There is also the problem that _offset is ignored on ReadByte and other ReadXX methods. I want to be able to code the full unit test, covering the case where _offset is not 0.

xljiulang:
This bug and https://github.com/dotnet/MQTTnet/issues/1592 has no direct relation.

PR #1598, opened 2022-11-30
fix MqttBufferReader #1597

xljiulang:
We design absolute `_postion` field and relative `Position` Property:
```c#
int _position;  // absolute 
int _maxPosition; // absolute 
public int Position => _position - _offset;  // relative 
```

### DIFF
--- a/Source/MQTTnet/Formatter/MqttBufferReader.cs
+++ b/Source/MQTTnet/Formatter/MqttBufferReader.cs
@@ -2,25 +2,27 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System;
-using System.Runtime.CompilerServices;
-using System.Text;
 using MQTTnet.Exceptions;
 using MQTTnet.Internal;
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Text;
 
 namespace MQTTnet.Formatter
 {
     public sealed class MqttBufferReader
     {
         byte[] _buffer = EmptyBuffer.Array;
-        int _maxPosition;
         int _offset;
+        int _position;
+        int _maxPosition;
 
-        public int BytesLeft => _maxPosition - Position;
+        public int BytesLeft => _maxPosition - _position;
 
         public bool EndOfStream => BytesLeft == 0;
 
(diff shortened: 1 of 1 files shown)


Source/MQTTnet/Formatter/MqttBufferReader.cs
```diff
@@ -34,52 +36,53 @@ public byte[] ReadBinaryData()
...
 
         public uint ReadFourByteInteger()
         {
             ValidateReceiveBuffer(4);
 
 #if NETCOREAPP3_0_OR_GREATER
-            var value = System.Runtime.InteropServices.MemoryMarshal.Read<uint>(_buffer.AsSpan().Slice(Position));
-            value = System.Buffers.Binary.BinaryPrimitives.ReverseEndianness(value);
+            var value = MemoryMarshal.Read<uint>(_buffer.AsSpan(_position));
+            if (BitConverter.IsLittleEndian)
```
chkr1011, line 57: I intentionally avoided added the endian check because the old code (below) also does not revert the values if required.
So, what will happen when the old code is running on a little-endian machine?

xljiulang: The old code is running on a little-endian machine correctly, but not on a big-endian machine.

xljiulang: The network byte order is the same as the big-endian machine byte order, so reverse will get the wrong value.

xljiulang: Left shift high bits is correct in all cases.
We can use [BinaryPrimitives.ReadUInt32BigEndian`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Buffers/Binary/ReaderBigEndian.cs#L112) method if the Api is supported,
```c#
public static uint ReadUInt32BigEndian(ReadOnlySpan<byte> source)
{
    return BitConverter.IsLittleEndian ?
        ReverseEndianness(MemoryMarshal.Read<uint>(source)) :
        MemoryMarshal.Read<uint>(source);
}
```

chkr1011: Then let's use the new API then. Then I will merge the PR.

PR #1599, opened 2022-11-30
Implement buffer copy using different policies

xljiulang:
### 
Currently, we still use `Array.Copy()`, the oldest Api for buffer copying. In other non-Net452 frameworks, we can use `System.Buffer.MemoryCopy()` or `Span.CopyTo()` for best copying performance.

```c#
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static void Copy(byte[] source, int sourceIndex, byte[] destination, int destinationIndex, int length)
{
#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_1
    source.AsSpan(sourceIndex, length).CopyTo(destination.AsSpan(destinationIndex, length));
#elif NET461_OR_GREATER || NETSTANDARD1_3_OR_GREATER
    unsafe
    {
        fixed (byte* pSoure = &source[sourceIndex])
        {
            fixed (byte* pDestination = &destination[destinationIndex])
            {
                System.Buffer.MemoryCopy(pSoure, pDestination, length, length);
            }
        }
    }
#else
    Array.Copy(source, sourceIndex, destination, destinationIndex, length);
#endif
}
```

### Benchmark
``` ini

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.819)
Intel Core i7-8565U CPU 1.80GHz (Whiskey Lake), 1 CPU, 8 logical and 4 physical cores
.NET SDK=7.0.100
  [Host]   : .NET 6.0.11 (6.0.1122.52304), X64 RyuJIT AVX2
  .NET 6.0 : .NET 6.0.11 (6.0.1122.52304), X64 RyuJIT AVX2

Job=.NET 6.0  Runtime=.NET 6.0  

```
|     Method | Length |      Mean |     Error |    StdDev |    Median | Ratio | RatioSD | Rank | Allocated | Alloc Ratio |
|----------- |------- |----------:|----------:|----------:|----------:|------:|--------:|-----:|----------:|------------:|
| **Array_Copy** |     **63** |  **5.988 ns** | **0.1452 ns** | **0.1987 ns** |  **6.077 ns** |  **1.00** |    **0.00** |    **1** |         **-** |          **NA** |
|  Memory_Copy |     63 |  5.917 ns | 0.1181 ns | 0.1656 ns |  5.916 ns |  0.99 |    0.04 |    1 |         - |          NA |
|            |        |           |           |           |           |       |         |      |           |             |
| **Array_Copy** |    **127** |  **7.579 ns** | **0.1771 ns** | **0.2809 ns** |  **7.671 ns** |  **1.00** |    **0.00** |    **2** |         **-** |          **NA** |
|  Memory_Copy |    127 |  5.588 ns | 0.1394 ns | 0.2441 ns |  5.648 ns |  0.74 |    0.03 |    1 |         - |          NA |
|            |        |           |           |           |           |       |         |      |           |             |
| **Array_Copy** |    **255** | **14.175 ns** | **0.3023 ns** | **0.4524 ns** | **14.257 ns** |  **1.00** |    **0.00** |    **2** |         **-** |          **NA** |
|  Memory_Copy |    255 | 11.452 ns | 0.2560 ns | 0.2628 ns | 11.431 ns |  0.80 |    0.04 |    1 |         - |          NA |
|            |        |           |           |           |           |       |         |      |           |             |
| **Array_Copy** |    **511** | **22.364 ns** | **0.4648 ns** | **0.5879 ns** | **22.661 ns** |  **1.00** |    **0.00** |    **2** |         **-** |          **NA** |
|  Memory_Copy |    511 | 19.992 ns | 0.3665 ns | 0.4363 ns | 20.175 ns |  0.90 |    0.03 |    1 |         - |          NA |
|            |        |           |           |           |           |       |         |      |           |             |
| **Array_Copy** |   **1023** | **44.274 ns** | **2.0469 ns** | **5.9709 ns** | **41.381 ns** |  **1.00** |    **0.00** |    **2** |         **-** |          **NA** |
|  Memory_Copy |   1023 | 39.433 ns | 0.7393 ns | 1.3141 ns | 39.203 ns |  0.85 |    0.12 |    1 |         - |          NA |
|            |        |           |           |           |           |       |         |      |           |             |
| **Array_Copy** |   **2047** | **76.521 ns** | **1.5469 ns** | **1.8415 ns** | **76.380 ns** |  **1.00** |    **0.00** |    **1** |         **-** |          **NA** |
|  Memory_Copy |   2047 | 75.734 ns | 1.4966 ns | 1.9979 ns | 75.636 ns |  0.99 |    0.04 |    1 |         - |          NA |
|            |        |           |           |           |           |       |         |      |           |             |
| **Array_Copy** |   **5095** | **69.378 ns** | **1.3988 ns** | **2.0937 ns** | **69.849 ns** |  **1.00** |    **0.00** |    **1** |         **-** |          **NA** |
|  Memory_Copy |   5095 | 68.523 ns | 1.4039 ns | 1.5022 ns | 68.705 ns |  0.99 |    0.04 |    1 |         - |          NA |


Source/MQTTnet/Internal/Memory.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.Runtime.CompilerServices;
+
+namespace MQTTnet.Internal
+{
+    public static class Memory
```
gfoidl, line 6: * `internal`? I see a lot of types in the "Internal" namespace are `public`, but for new types it should really be `internal`

* Naming is hard (I'm not better in this), but `Memory` is for me Memory/Span, so maybe `CopyExtension` and make the single method an extension method? Or leave as static method and name the type `CopyHelper`?

Source/MQTTnet/Internal/Memory.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Copy(byte[] source, int sourceIndex, byte[] destination, int destinationIndex, int length)
```
gfoidl, line 9:
```suggestion
        public static void Copy(ReadOnlySpan<byte> source, Span<byte> destination)
```
? Makes this API simpler and more future-proof.

For ease of use at callsites this could be another overload where the byte[]-based one just forwards to the span-based one.

Source/MQTTnet/Internal/Memory.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_1
+            source.AsSpan(sourceIndex, length).CopyTo(destination.AsSpan(destinationIndex, length));
+#elif NET461_OR_GREATER || NETSTANDARD1_3_OR_GREATER
```
gfoidl, line 13: Personally I'd leave these "old" target with `Array.Copy` and give the goodness only to newer targets.
But I'm fine with this change too.

Source/MQTTnet/Internal/Memory.cs
chkr1011: I prefer to keep most things public to make testing easier (i know that "internal visible to" exists). So public is fine in my opinion. The "Internal" namespace is an indicator that these components do not have a stable API.

chkr1011: Regarding the: I also would prefer "MqttMemoryHelper" etc. because "Memory" is already in use by the framework itself.

Source/MQTTnet/Internal/Memory.cs
chkr1011: Since .NET 4.8 is still supported and used a lot I would like to keep the optimization even for "old" targets.

Source/MQTTnet/Internal/Memory.cs
chkr1011: I will merge the PR and do the changes myself so that w new version can be released soon.

PR #1602, opened 2022-12-01
Add MqttBufferReader fixes and test

logicaloud:
I have found some more issues in the MqttBufferReader and added a test that builds a buffer with random sequences of various types and then reads the buffer with different offsets and segment lengths to see if the read result matches the expected result. Thank you to @xljiulang for proposed fixes in #1597 which are also incorporated.


chkr1011:
@logicaloud Since the _MqttBufferReader _ got some more changes in another PR I decided to close this PR but copy the new unit tests to master. I hope you are fine with that.

logicaloud:
No problem. I thought I had found some additional issues that were not in the first pull request by @xljiulang but missed the Position / _position change which fixes the same thing in a more elegant way. The test passes now and that's the main thing.

PR #1298, opened 2021-11-15
Improved message processing performance for many topics

logicaloud:
This is a continuation of #1283 with cleaned up commits and merged `feature/Reason_Code_In_Subscription_Interceptor` branch. As this request will likely remain unmerged for a while, I'll try to keep this pull request in sync with the branch for the coming weeks. The original description - updated with corrections - follows.

### Motivation

In a typical use case for MQTT there may be many publishers (sensors or other devices in the field) and few subscribers (monitoring all or many topics). Each publisher may have one or more topic(s) to publish and therefore both, the number of publishers and the number of topics may be large.

Currently there are two bottlenecks in MQTTnet that slow down processing in this scenario:

1. For any message published, the server loops through all subscriber *and* publisher sessions to find subscriptions.

2. For any subscriber, the server loops through the number of subscribed topics until it finds a matching one.

I'd like to put a proposal up for discussion that speeds up processing of application messages (about 15x to 30x for many publishers and topics based on the additional test that is included for a quick assessment, same as #1282)

### Maintain subscribers in a separate container

Instead of placing all sessions into the one `_clientSessions` container, subscribers are added into another `_subscriberSessions` container (if a client is a subscriber *and* a publisher then the client is present in both containers). The cost is some additional bookkeeping work upon subscription where each client session needs to maintain a list of subscribed topics. 

When the client subscribes to the first topic, then the session manager adds the client to the `_subscriberSessions` container, and when the client unsubscribes from the last topic then the session manager removes the client from the container. Now, when an application message arrives, only the list of subscribers need processing instead of looping through potentially thousands of publishers.

### Improve subscriber topic lookup

This one is more complicated. For each subscriber, it needs to be determined whether an application message matches any topic the subscriber has subscribed to. There may only be few subscribers but there may be many subscribed topics, including wildcard topics.

The proposed approach uses a topic hash and a hash mask calculated on the subscribed topic and the published topic (the application message topic) to find candidates for a match, with the existing match logic evaluating a reduced number of candidates. 

For each subscription, the topic hash and a hash mask is stored with the subscription, and for each application message received, the hash is calculated for the published topic before attempting to find matching subscriptions. The hash calculation itself is simple and does not have a large performance impact.

I'll first explain how topic hashes and hash masks are constructed and then how they are used.

### Topic hash 

[...]

PR #1615, opened 2022-12-07
#1612 fix wildcard topic subscription

### LINKED ISSUE #1612: [MqttServer] Subscription to similar base and subtopic with wildcards fails
We encountered a bug which was not present in earlier major versions of MQTTnet, but we met it since we had a major upgrade to 4.1.2.350. Same applies for the newest version 4.1.2.436

We start an MQTTnet server. Then a client. We add a subscription to these to topics:

house/+/room
house/+/room/+

Then we send messages with these two topics:

house/1/room
house/1/room/bed

**We would expect to receive both messages, but only the first one arrives**. When testing this with other servers (older MQTTnet server or Moquitto) both messages arrive. 

As you can see in the sample code below, we also made some slight variations (# instead of +), but no success. When we only use one subscription (#) everything is fine again. Replacing the mid-level wildcard with a concrete house id (1) also makes it work again.

It is not an ultra-critical bug, since there is a workaround. But we lost data in production since we did not realize this problem immediately. Might be that some others may stumble over this, too.
 
```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
using MQTTnet;
using MQTTnet.Client;
using MQTTnet.Server;

namespace MqttServerTest
{
	class Program
	{
		static void Main(string[] args)
		{
			Task.Run(StartMqttServer).Wait();

			var client = Task.Run(StartClient).Result;

			client.SubscribeAsync("house/+/room");
			client.SubscribeAsync("house/+/room/+");

			//You can try these two subscriptions instead, but this will also fail
			//client.SubscribeAsync("house/+/room");
			//client.SubscribeAsync("house/+/room/#");

			//This works
			//client.SubscribeAsync("house/1/room");
			//client.SubscribeAsync("house/1/room/+");

			//This works, too
			//client.SubscribeAsync("house/1/room/#");

			Publish(client, "house/1/room");
			Publish(client, "house/1/room/bed");

			//Result with first two examples: Only first message will be shown

			Console.ReadLine();
		}

		static async Task<MqttServer> StartMqttServer()
		{
			var mqttFactory = new MqttFactory();
			
			var mqttServerOptions = 
				mqttFactory
					.CreateServerOptionsBuilder()
					.WithDefaultEndpoint()
					.Build();
			var server = mqttFactory.CreateMqttServer(mqttServerOptions);
			await server.StartAsync();
			return server;
		}
		
		static async Task<IMqttClient> StartClient()
		{
			var mqttFactory = new MqttFactory();

			var mqttClientOptions =
				mqttFactory
					.CreateClientOptionsBuilder()
					.WithTcpServer("localhost")
					.WithClientId("test")
					.Build();

			var client = mqttFactory.CreateMqttClient();

			client.ApplicationMessageReceivedAsync += args =>
			{
				Console.WriteLine(args.ApplicationMessage.Topic);
				return Task.CompletedTask;
			};

			await client.ConnectAsync(mqttClientOptions, CancellationToken.None);
			
			return client;
		}
		
		private static Task<MqttClientPublishResult> Publish(IMqttClient client, string topic)
		{
			return client.PublishAsync(new MqttApplicationMessage
			{
				Topic = topic,
				Payload = new byte[] { 1, 2, 3, 4 }
			});
		}
	}
}
```

logicaloud:
Fixes #1612 

There was an error in the early elimination of wildcard topics. It worked under the assumption that the same topic hash will have the same topic hash mask, which is not always the case for wildcard topics. This is fixed by storing wildcard subscriptions by topic hash mask and by comparing against the correct hash mask at the elimination stage. Unit test added also.

### DIFF
--- a/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
+++ b/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
@@ -60,13 +60,16 @@ namespace MQTTnet.Server
 
                 foreach (var wcs in _wildcardSubscriptionsByTopicHash)
                 {
-                    var wildcardSubscriptions = wcs.Value;
                     var subscriptionHash = wcs.Key;
-                    var subscriptionHashMask = wildcardSubscriptions.HashMask;
-
-                    if ((topicHash & subscriptionHashMask) == subscriptionHash)
+                    var subscriptionsByHashMask = wcs.Value.SubscriptionsByHashMask;
+                    foreach (var shm in subscriptionsByHashMask)
                     {
-                        possibleSubscriptions.AddRange(wildcardSubscriptions.Subscriptions.ToList());
+                        var subscriptionHashMask = shm.Key;
+                        if ((topicHash & subscriptionHashMask) == subscriptionHash)
+                        {
+                            var subscriptions = shm.Value;
+                            possibleSubscriptions.AddRange(subscriptions.ToList());
+                        }
                     }
                 }
             }
@@ -265,8 +268,8 @@ namespace MQTTnet.Server
                             {
                                 if (_wildcardSubscriptionsByTopicHash.TryGetValue(topicHash, out var subscriptions))
                                 {
-                                    subscriptions.Subscriptions.Remove(existingSubscription);
-                                    if (subscriptions.Subscriptions.Count == 0)
+                                    subscriptions.RemoveSubscription(existingSubscription);
+                                    if (subscriptions.SubscriptionsByHashMask.Count == 0)
                                     {
                                         _wildcardSubscriptionsByTopicHash.Remove(topicHash);
                                     }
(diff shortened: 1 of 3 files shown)


xljiulang:
👍

Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
```diff
@@ -60,13 +60,16 @@ public CheckSubscriptionsResult CheckSubscriptions(string topic, ulong topicHash
...
-                    if ((topicHash & subscriptionHashMask) == subscriptionHash)
+                    var subscriptionsByHashMask = wcs.Value.SubscriptionsByHashMask;
+                    foreach (var shm in subscriptionsByHashMask)
                     {
-                        possibleSubscriptions.AddRange(wildcardSubscriptions.Subscriptions.ToList());
+                        var subscriptionHashMask = shm.Key;
+                        if ((topicHash & subscriptionHashMask) == subscriptionHash)
+                        {
+                            var subscriptions = shm.Value;
+                            possibleSubscriptions.AddRange(subscriptions.ToList());
```
chkr1011, line 71: Is the call of "ToList()" really required? If not, it is only a useless allocation.

logicaloud: Good point, that's redundant. I did not notice a change in performance with a cursory look; maybe it's optimized away. I'll do some benchmarking to compare against https://github.com/dotnet/MQTTnet/pull/1298#issuecomment-1036967766 to see if anything else has changed, also in respect of the new AsyncLock.

PR #1622, opened 2022-12-07
1612 mqttserver subscription to similar base and subtopic with wildcards fails

chkr1011:


logicaloud:
Regarding performance, wildcard subscription performance is slightly down, as one would expect; probably not noticeable unless there are thousands of wildcard subscriptions. The "MessageDeliveryBenchmark" measures non-wildcard subscriptions only; would be good to add some wildcard subscription benchmark at some point to have a reliable baseline. 

The change to `AsyncLock` holds up well, with performance on par with the semaphore, sometimes better. 

Since I have been looking at the "MessageDeliveryBenchmark" benchmark tests: The current version with NET6 is between 5 to 58 times faster than v3 /NET5 and reduces memory allocations by 90% or more. Compared to the first test after improvements in v4 / NET5, the current version with NET6 is between 1.6 and 2 times faster and reduces memory allocations by a further 10%-20%.

Issue #1627, opened 2022-12-15
Allow setting session items when calling InjectApplicationMessage


tboothman:
### Describe the feature request
Allow the user to specify the session items in the virtual session for the server when injecting application messages.

Background:
We inject MQTT messages using InjectApplicationMessage but since 4.0 the InterceptingPublishAsync hook is called for injected messages. We check the session for metadata about the client to see if they're allowed to publish, which is now just an empty dictionary so it gets rejected.

### Which project is your feature request related to?
- Server

### Describe the solution you'd like
The method InjectApplicationMessage  to take an argument for session items

### Describe alternatives you've considered
Set the session items on the server and just use that for every InjectApplicationMessage. There's a property called _sessionItems with an empty dictionary that's currently used as the session. Could add a method to change that property..

PR #1629, opened 2022-12-15
Custom session items when injecting application message

tboothman:
Optional parameter for session items to be used when publishing a message from the server

See #1627

### DIFF
--- a/Source/MQTTnet/Server/MqttServer.cs
+++ b/Source/MQTTnet/Server/MqttServer.cs
@@ -207,7 +207,15 @@ namespace MQTTnet.Server
             return _clientSessionsManager.GetSessionStatusAsync();
         }
 
-        public Task InjectApplicationMessage(InjectedMqttApplicationMessage injectedApplicationMessage, CancellationToken cancellationToken = default)
+        public Task InjectApplicationMessage(
+            InjectedMqttApplicationMessage injectedApplicationMessage,
+            CancellationToken cancellationToken = default) =>
+            InjectApplicationMessage(
+                injectedApplicationMessage,
+                _sessionItems,
+                cancellationToken);
+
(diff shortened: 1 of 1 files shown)


Source/MQTTnet/Server/MqttServer.cs
```diff
@@ -207,7 +207,15 @@ public Task<IList<MqttSessionStatus>> GetSessionsAsync()
...
-        public Task InjectApplicationMessage(InjectedMqttApplicationMessage injectedApplicationMessage, CancellationToken cancellationToken = default)
+        public Task InjectApplicationMessage(
+            InjectedMqttApplicationMessage injectedApplicationMessage,
+            CancellationToken cancellationToken = default) =>
+            InjectApplicationMessage(
+                injectedApplicationMessage,
+                _sessionItems,
+                cancellationToken);
+
+        public Task InjectApplicationMessage(InjectedMqttApplicationMessage injectedApplicationMessage, ConcurrentDictionary<object, object> sessionItems, CancellationToken cancellationToken = default)
```
chkr1011, line 218: I don't like the idea to pass in session items for every injected message. The server should behave like a single client. 

We may need to expose the session items as a public property for _MqttServer_. It is currently missing the feature to setup session items externally. This is usually done for clients when they connect but the server has no connection at all so we must prepare the session items somewhere else.

chkr1011 (changes requested):

Source/MQTTnet/Server/MqttServer.cs
YAJeff: As a part of #1635 I added the ability to get session items using the IMqttServerExtensibility interface. Maybe this is a way of exposing it without opening up a SessionItems dictionary to all? This also allows users to supply whatever session items they want by calling directly to the MqttClientSessionsManager.

chkr1011:
Closing this PR because a similar change is part of another PR.

PR #1634, opened 2022-12-23
Added AsyncReadWriteLock

YAJeff:
Added AsyncReadWriteLock and converted client subscriptions manager to using it.

In benchmark tests, this can provide logarithmic performance benefits. When applied to subscriptions, where messages might be occurring more frequently than subscribe/unsubscribe events, and there is no need for them to block each other.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/MQTTnet?pullRequest=1634) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/MQTTnet?pullRequest=1634) before we can accept your contribution.<br/><br/>:x: YAJeff [sign now](https://cla.dotnetfoundation.org/dotnet/MQTTnet?pullRequest=1634)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/MQTTnet?pullRequest=1634) it.</sub>

YAJeff:
Added extensibility interface to MqttServer for advanced clustering support scenarios.

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,214 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (_isDisposed)
+            {
+                throw new ObjectDisposedException(nameof(AsyncReadWriteLock));
+            }
+
+            lock (_syncRoot)
+            {
+                if (!_isLockedForWrite && !_writeWaiters.Any())
```
gfoidl, line 42:
```suggestion
                if (!_isLockedForWrite && _writeWaiters.Count == 0)
```

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                _isLockedForWrite = false;
+
+                ApproveNextWaiter();
+            }
+        }
+
+        private void ApproveNextWaiter()
+        {
+            while (_writeWaiters.Any())
```
gfoidl, line 111:
```suggestion
            while (_writeWaiters.Count > 0)
```

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,214 @@
...
+                var isApproved = waiter.Approve(_writeReleaser);
+                waiter.Dispose();
+
+                if (isApproved)
+                {
+                    _isLockedForWrite = true;
+                    return;
+                }
+            }
+            while (_readWaiters.Any())
```
gfoidl, line 123:
```suggestion
            while (_readWaiters.Count > 0)
```

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,214 @@
...
+                {
+                    _readLockCount++;
+                    return;
+                }
+            }
+        }
+
+        public void Dispose()
+        {
+            _isDisposed = true;
```
gfoidl, line 139: Should this be volatile write / read? (Except where in a `lock`)

gfoidl (commented):
Just looked at code.
I'm missing some context (in this lib) to review the actual change.

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,214 @@
...
+        private readonly Queue<ReadWriteLockWaiter> _writeWaiters = new Queue<ReadWriteLockWaiter>();
+        private readonly IDisposable _readReleaser;
+        private readonly IDisposable _writeReleaser;
+        private readonly Task<IDisposable> _readCompletedTask;
+        private readonly Task<IDisposable> _writeCompletedTask;
+        private int _readLockCount;
+        private bool _isLockedForWrite;
+        private bool _isDisposed;
+
+        public AsyncReadWriteLock()
```
chkr1011, line 23: Please add multiple unit tests and cover all features of this lock so that we can proof it is working.

chkr1011 (changes requested):

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
YAJeff: I'm not sure what you mean by this. Could you clarify?

gfoidl: Is there a possibility that `Dispose` is called from multiple threads? I don't know.
Is there a possibility that the value of `_isDisposed` is read from multiple threads? Yes.

At least the latter case may introduce a visibility issue on weaker memory model targets (like arm).
So I think here it should be 
```suggestion
            Volatile.Write(ref _isDisposed, true);
```
and in all reads that are not within a `lock` it should be `Volatile.Read(ref _isDisposed)` to be on the safe side.

On targets with strong memory models (like x86) these volatile read/writes evaporate and have no cost.

YAJeff: Thank you for pointing that out. I've been working in .NET for a very long time, and I had no idea these volatile keywords existed. You learn something new every day. Makes sense for ARM-based systems though. Thanks again.

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
YAJeff: Added multiple unit tests.

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,215 @@
...
+            _writeReleaser = new Releaser(this, false);
+            _readCompletedTask = Task.FromResult(_readReleaser);
+            _writeCompletedTask = Task.FromResult(_writeReleaser);
+        }
+
+        public Task<IDisposable> EnterReadAsync(CancellationToken cancellationToken = default)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (_isDisposed)
```
gfoidl, line 35:
```suggestion
            if (Volatile.Read(ref _isDisposed))
```

Source/MQTTnet/Internal/AsyncReadWriteLock.cs
```diff
@@ -0,0 +1,215 @@
...
+
+                return waiter.Task;
+            }
+        }
+
+        public Task<IDisposable> EnterWriteAsync(CancellationToken cancellationToken = default)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            if (_isDisposed)
```
gfoidl, line 59:
```suggestion
            if (Volatile.Read(ref _isDisposed))
```

gfoidl (commented):
Here volatile reads are needed.

YAJeff:
@chkr1011 - Do you have any more suggestions?

chkr1011:
Sorry I am not yet done with the reviews. I mark them as done as soon as I am.

YAJeff:
No worries. Wasn't sure if I missed something. I'm new to contributing. Having fun doing it. 👍

PR #1636, opened 2022-12-25
dotnet 7 packages

YAJeff:
Added dotnet 7 support/builds.


Source/MQTTnet/Client/Options/MqttClientOptionsBuilderTlsParameters.cs
```diff
@@ -15,7 +15,7 @@ public sealed class MqttClientOptionsBuilderTlsParameters
 
         public Func<MqttClientCertificateValidationEventArgs, bool> CertificateValidationHandler { get; set; }
 
-#if NET48 || NETCOREAPP3_1 || NET5 || NET6
+#if NET48 || NETCOREAPP3_1 || NET5 || NET6 || NET7_0
```
chkr1011, line 18: We should change this to _NETCOREAPP3_1_OR_GREATER_ so that we no longer need to update this for every new version.

chkr1011 (changes requested):

chkr1011:

Issue #1654, opened 2023-01-21
Broken build, IsAssignableFrom not found, when targetting UWP


YAJeff:
### Description
After cloing the repo, and attempting to rebuild all within Visual Studio, there is a failure on the MQTTnet project for UWP.

### Which component is your bug related to?
- MQTTnet library

### To Reproduce
1. Clone the latest repo onto a machine that has UWP SDKs installed
2. Attempt to Rebuild All from Visual Studio 2022

### Expected behavior
Build succeeds

### Additional context / logging
repos\MQTTnet\Source\MQTTnet\Internal\AsyncQueue.cs(41,41,41,57): error CS1061: 'Type' does not contain a definition for 'IsAssignableFrom' and no accessible extension method 'IsAssignableFrom' accepting a first argument of type 'Type' could be found (are you missing a using directive or an assembly reference?)

PR #1655, opened 2023-01-21
Fixed broken UWP build

YAJeff:
Added compiler declaritive to check for WINDOWS_UWP on IsAssignable from call.


YAJeff:
This is associated with #1654

Issue #1660, opened 2023-01-27
[Server] Retained messages store optimized for database


Int32Overflow:
### Describe the feature request
Currently, all retained messages are kept in the working memory, which can quickly become a problem with larger quantities. It would be a good idea to store these in a SQL database or in Redis, for example. 
However, the "interfaces" are currently very limited in order to implement such a concept.

### Which project is your feature request related to?
- Server

### Describe the solution you'd like
In the method "FilterRetainedApplicationMessages" the retained messages are no longer handled in a for loop. The topic is now passed to a method in the new interface "IMqttClientSubscriptionsManager", which can return several retained messages as a result.
This interface can of course be overwritten.

https://github.com/dotnet/MQTTnet/blob/2aa4219838a67e336ae4f5902d956c455bc48c11/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs#L401-L453

chkr1011:
I am wondering how to get the matching retained messages from the database. If the topic contains no wildcards ist might be working (or some usages of #). But how to get the proper set of messages from the database if the "+" wildcard is used?

Int32Overflow:
For example cache only topics in ram and load payload and other stuff on match from database.

Int32Overflow:
Perhaps it is also possible to split the topics into individual segments and then use them like a tree path. PostgreSQL can map trees for example...

logicaloud:
How the retained message is looked up should probably remain an extension because different storage providers will have different capabilities. For SQL, the `LIKE` condition would probably work, with wildcard topic characters becoming `%` in the query. I.e. for topic `house/1/room/+/temperature` the query would be something like `select * from retainedmessages where topic like 'house/1/room/%/temperature'`. Case sensitivity and special characters (underscore) would need to be taken care of.

Int32Overflow:
@chkr1011 Can you give me an update?

I was able to successfully save and search the Mqtt topics in a Postgresql table.
Postgresql supports a very fast way to search for topics that match a certain topic pattern with the help of index.

```sql
-- https://www.postgresql.org/docs/current/ltree.html

CREATE EXTENSION  ltree;

CREATE TABLE IF NOT EXISTS test (path ltree not null unique, message bytea);
CREATE INDEX path_gist_idx ON test USING GIST (path gist_ltree_ops(siglen=1000));
CREATE INDEX path_idx ON test USING BTREE (path);

INSERT INTO test VALUES ('Root1');
[...]

PR #1663, opened 2023-01-27
Retained message store extendable

Int32Overflow:
See the issue #1660

Now the interface "IMqttRetainedMessagesManager" can be used to build your own retained message store.


Int32Overflow:
@chkr1011 What do you think? Does it fit?

logicaloud:
Is this progressing? I just had an attempt to use events as an extension method for external storage but it is no good (#1763).

The current events work ok as long as one can accept that all retained messages are kept in memory, but now I'm looking for a solution where they can be loaded on demand, i.e. by using the proposed `IMqttRetainedMessageManager` interface.

Hoseinmj (approved):

PR #1685, opened 2023-03-02
ClientMessageQueueInterceptor continues with subscription processing.

RossHNPC:
@chkr1011 Here is a draft PR for the issue I found earlier with the ClientMessageQueueInterceptor with a unit test added to capture the use case of 3 subscribed Clients and a coarse filter in the handler to exclude one of them from receiving messages.

Unit tests I've ran locally... all pass when ran individually be few not so happy when running all together; consistently Disconnect_Clean and Disconnect_Clean_With_Custom_Reason irrespective of target framework, but when ran on their own will pass.


RossHNPC:
@dotnet-policy-service agree company="Natural Power Consultants Ltd."

PR #1694, opened 2023-03-21
Update AsyncLockBenchmark.cs

demo-478:


chkr1011:
@demo-478 Why did you made this change? What does it fix?

PR #1728, opened 2023-05-06
Expose proper connect result in clients Disconnected event

chkr1011:
This PR makes sure that the response from the server is now exposed in the disconnected event of the client in case the authentication has failed. Without this PR it is always _null_ in such cases.

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using MQTTnet.Adapter;
+using MQTTnet.Client.Internal;
 using MQTTnet.Diagnostics;
 using MQTTnet.Exceptions;
 using MQTTnet.Formatter;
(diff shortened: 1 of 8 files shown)


Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -427,18 +421,26 @@ Task AcknowledgeReceivedPublishPacket(MqttApplicationMessageReceivedEventArgs ev
...
-                var clientConnectResultFactory = new MqttClientConnectResultFactory();
-                result = clientConnectResultFactory.Create(connAckPacket, _adapter.PacketFormatterAdapter.ProtocolVersion);
+                if (receivedPacket is MqttConnAckPacket connAckPacket)
+                {
+                    var clientConnectResultFactory = new MqttClientConnectResultFactory();
+                    result = clientConnectResultFactory.Create(connAckPacket, _adapter.PacketFormatterAdapter.ProtocolVersion);
+                }
+                else
+                {
+                    throw new NotSupportedException("Extended authentication handler is not yet supported");
```
vishnureddy17, line 442: Just to clarify, extended auth is not supported in the case where the client receives an `AUTH` packet in between sending a `CONNACK` and receiving a `CONNECT` packet, right? But otherwise, MQTT enhanced auth is supported in the following cases:
- Sending authentication method and authentication data in the `CONNECT` packet and immediately receiving a `CONNACK` packet from the server without any `AUTH` packets in between.
- Initiating extended authentication while already connected and having an `AUTH` exchange of arbitrary length until the authentication succeeds (or a disconnection occurs).

@chkr1011 is my understanding right? Sorry if this is a basic question.

PR #1747, opened 2023-05-29
1738 userproperties in clientconnectedeventargs

chkr1011:
This PR exposes the user properties in the client connected event arguments.


wmchuang:
Excuse me, why not expose more? For example KeepAlivePeriod

Issue #1754, opened 2023-06-05
Websockets no longer working in Blazor WebAssembly


Nickztar:
### Describe the bug
After the latest release Websockets are no longer supported in a Browser platform like Blazor WASM. This is due to the new properties added with https://github.com/dotnet/MQTTnet/pull/1734. These properties are not available on a browser platform and setting them result in the current issue.

### Which component is your bug related to?
- Client

### To Reproduce
1. Using this version of MQTTnet 'v4.2.1.781'.
2. Run this code '[Basic setup](https://gist.github.com/Nickztar/d5e90a33f8abff6a8821088af44dfe41)'.
4. With these arguments 'WithWebsocketServer'.
5. See error in addition context.

### Expected behavior
Websockets should work as expected in browser. Just that browser will not have these settings available.

### Additional context / logging

```batch
MQTTnet.Exceptions.MqttCommunicationException: Operation is not supported on this platform.
 ---> System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Net.WebSockets.ClientWebSocketOptions.set_UseDefaultCredentials(Boolean value)
   at MQTTnet.Implementations.MqttWebSocketChannel.SetupClientWebSocket(ClientWebSocket clientWebSocket)
   at MQTTnet.Implementations.MqttWebSocketChannel.ConnectAsync(CancellationToken cancellationToken)
   at MQTTnet.Adapter.MqttChannelAdapter.ConnectAsync(CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at MQTTnet.Adapter.MqttChannelAdapter.WrapAndThrowException(Exception exception)
   at MQTTnet.Adapter.MqttChannelAdapter.ConnectAsync(CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.ConnectInternal(IMqttChannelAdapter channelAdapter, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.ConnectAsync(MqttClientOptions options, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.ConnectAsync(MqttClientOptions options, CancellationToken cancellationToken)
   at Project.Services.MqttService.Connect() in PATH\src\Project\Services\MqttService.cs:line 142
```
### Code example
[...]

PR #1734, opened 2023-05-10
UseDefaultCredentials options for WebSockets proxy configuration.

impworks:
This PR adds an option to pass `UseDefaultCredentials` option to the `WebProxy` used by WebSockets implementation.

PR #1755, opened 2023-06-05
Websockets no longer working in Blazor WebAssembly

Nickztar:
After the latest release Websockets are no longer supported in a Browser platform like Blazor WASM. This is due to the new properties added with https://github.com/dotnet/MQTTnet/pull/1734. These properties are not available on a browser platform and setting them result in the current issue.

This solves this issue by only applying them if they are explictly set. Not sure if this is the best way to solve this. Another option that could have worked is if you could conditionally compile for BROWSER platform but I could not get that work (not sure if it's actually supported?).

This solves my issue. #1754

### DIFF
--- a/Source/MQTTnet/Implementations/MqttWebSocketChannel.cs
+++ b/Source/MQTTnet/Implementations/MqttWebSocketChannel.cs
@@ -221,11 +221,20 @@ namespace MQTTnet.Implementations
 
 #if !NETSTANDARD1_3
 #if !WINDOWS_UWP
-            clientWebSocket.Options.UseDefaultCredentials = _options.UseDefaultCredentials;
+            if (_options.UseDefaultCredentials)
+            {
+                clientWebSocket.Options.UseDefaultCredentials = _options.UseDefaultCredentials;
+            }
 #endif
-            clientWebSocket.Options.KeepAliveInterval = _options.KeepAliveInterval;
+            if (_options.KeepAliveInterval != WebSocket.DefaultKeepAliveInterval)
+            {
+                clientWebSocket.Options.KeepAliveInterval = _options.KeepAliveInterval;
+            }
 #endif
-            clientWebSocket.Options.Credentials = _options.Credentials;
+            if (_options.Credentials != null)
+            {
+                clientWebSocket.Options.Credentials = _options.Credentials;
+            }
             
             var certificateValidationHandler = _options.TlsOptions?.CertificateValidationHandler;
             if (certificateValidationHandler != null)
--- a/Source/MQTTnet/Implementations/MqttWebSocketChannel.cs
+++ b/Source/MQTTnet/Implementations/MqttWebSocketChannel.cs
@@ -221,6 +221,8 @@ namespace MQTTnet.Implementations
 
 #if !NETSTANDARD1_3
 #if !WINDOWS_UWP
+            // Only set the value if it is actually true. This property is not supported on all platforms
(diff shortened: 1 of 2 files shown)


Nickztar:
@dotnet-policy-service agree

Source/MQTTnet/Implementations/MqttWebSocketChannel.cs
```diff
@@ -221,11 +221,20 @@ void SetupClientWebSocket(ClientWebSocket clientWebSocket)
 
 #if !NETSTANDARD1_3
 #if !WINDOWS_UWP
-            clientWebSocket.Options.UseDefaultCredentials = _options.UseDefaultCredentials;
+            if (_options.UseDefaultCredentials)
```
chkr1011, line 224: I also crated a branch with a fix then found your PR. Basically it is the same but I added the following comment above line 224.

            // Only set the value if it is actually true. This property is not supported on all platforms
            // and will throw a _PlatformNotSupported_ (i.e. WASM) exception when being used regardless of the actual value.

Please add such a comment so that other know why we have to do it this way.

Then I will approve and merge the PR.

chkr1011 (changes requested):

Nickztar:
Hi @chkr1011! Added the requested comment, maybe a similiar comment should be added to the other properies that also cannot be set on a browser platform?

chkr1011:
If you referring to the ones below I think it is enough.

Nickztar:
I was and you are probably correct! 👍🏼

jokke009:
Hi all,
First time posting in this project. Having gone many posts, I think I've missed something as this issue is closed. 
Has this been resolved? I found myself trying mqttnet 4.2.1.781 yesterday, on an existing blazer wasm project, first refactoring the code (handles --> asyncs etc.) but I could never get passed the 'platform no supported' error. Tried everything but I don't think I understand the workaround (if there is one?) I read something about UseDefaultCredentials, not sure how  that fits in. 

Would anybody be able to post a simple working example? (options and connection configuration).  Below is wat I tried.

For now I've reverted back to mqttnet 3.1.1, which still works without issue. 

Thanks in advance. 

`    public static async Task Connect()
    {
        var mqttFactory = new MqttFactory();
        using (var mqttClient = mqttFactory.CreateMqttClient())
        {
            var mqttClientOptions = new MqttClientOptionsBuilder().WithWebSocketServer("test.mosquitto.org:8081").Build();    
            mqttClientOptions.ChannelOptions = new MqttClientWebSocketOptions
                {
                    Uri = "test.mosquitto.org:8081",
                    UseDefaultCredentials = false
                };
            var response = await mqttClient.ConnectAsync(mqttClientOptions, CancellationToken.None);
            Console.WriteLine("The MQTT client is connected.");
        }
    }`

Nickztar:
Hi @jokke009! This has been closed since it's been merged and is working in the master branch. 

The reason why latest release still has this issue is that it hasn't been released yet. Not sure what the release strategy is for MQTTNet but I would guess they collect some amount of issues before actually making a release. 

I am in the same boat as you so we are currently running on version [v4.1.4.563](https://github.com/dotnet/MQTTnet/releases/tag/v4.1.4.563) which is working for me. Possible that 4.2 would work aswell, haven't checked.

chkr1011:
GitHub automatically closed the issue because there was a PR merged with a fix. But the nuget is not yet released. It will be released next weekend.

PR #1762, opened 2023-06-12
Fix #1761: MQTTRetainedMessageManager, return task with NULL result instead of NULL

### LINKED ISSUE #1761: MQTTRetainedMessageManager returns NULL instead of a task with NULL result
### Describe the bug

In `MQTTRetainedMessageManager`, method `GetMessage(string topic)`, NULL is returned if the specific topic is not found, instead of returning a task with NULL result.

- Server

### To Reproduce
Call `var result = GetMessage("undefined topic")` and await the result: This will cause an exception because the returned task is null.

### Expected behavior
A task with NULL result should be returned.

### Code example

Will submit a separate pull request.

logicaloud:
Fixes #1761

### DIFF
--- a/.github/workflows/ReleaseNotes.md
+++ b/.github/workflows/ReleaseNotes.md
@@ -1,1 +1,1 @@
-* 
+* [Server] Fixed _NullReferenceException_ in retained messages management (#1762, thanks to @logicaloud).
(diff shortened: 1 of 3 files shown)


logicaloud:
@gfoidl Thanks, that's better. I have updated the code.

Issue #1796, opened 2023-07-19
Add support for per request MqttRpcTopicPair specification to ExecuteAsync method


Temppus:
### Describe the feature request
I am using `MQTTnet.Extensions.Rpc` to do RPC calls via MQTT (great library btw 👍 ).
I would like to use one `IMqttRpcClient` to do requests to different request/response topics. Prove me wrong but this is not currently supported by library, because `ExecuteAsync` method overloads only accepts `methodName` as parameter. So when I want to create dynamic `MqttRpcTopicPair` for each `ExecuteAsync` call I have to encode/serialize this to method name and parse it in custom implementation of `IMqttRpcClientTopicGenerationStrategy`.

### Which project is your feature request related to?

- ManagedClient

### Describe the solution you'd like
Add some kind of key value property maybe `IDictionary<string, object>` to `TopicGenerationContext` so I do not have to serialize/parse this "per request" topic generation from method name.

### Describe alternatives you've considered
Add overload of `ExecuteAsync` with `MqttRpcTopicPair` as parameter ?

Temppus:
I am open to create PR for this if feature will be accepted and design choice made.

chkr1011:
Sounds like a good idea to me. To summarize it:
- Add a a new optional(!) parameter of type _IDictionary<string, object>_ called "parameters" to the _ExecuteAsnyc_ method
- Exposing the "Parameters" in the topic generation context so that the users can pass custom data which is required for topic generation.

Temppus:
Create PR https://github.com/dotnet/MQTTnet/pull/1798

Temppus:
Hey @chkr1011 can you find some time to review PR ?

PR #1798, opened 2023-07-24
Add possibility to specify parameter to RPC invocation

Temppus:
Fixes https://github.com/dotnet/MQTTnet/issues/1796

### DIFF
--- a/Source/MQTTnet.Extensions.Rpc/Options/TopicGeneration/TopicGenerationContext.cs
+++ b/Source/MQTTnet.Extensions.Rpc/Options/TopicGeneration/TopicGenerationContext.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using MQTTnet.Client;
 using MQTTnet.Protocol;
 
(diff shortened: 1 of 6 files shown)


Source/MQTTnet.Extensions.Rpc/Options/TopicGeneration/TopicGenerationContext.cs
```diff
@@ -3,23 +3,30 @@
...
     public sealed class TopicGenerationContext
     {
-        public TopicGenerationContext(IMqttClient mqttClient, MqttRpcClientOptions options, string methodName, MqttQualityOfServiceLevel qualityOfServiceLevel)
+        static readonly IDictionary<string, object> EmptyParameters = new Dictionary<string, object>();
+
+        public TopicGenerationContext(IMqttClient mqttClient, MqttRpcClientOptions options, string methodName,
+            IDictionary<string, object> parameters, MqttQualityOfServiceLevel qualityOfServiceLevel)
         {
             MethodName = methodName ?? throw new ArgumentNullException(nameof(methodName));
+            Parameters = parameters ?? EmptyParameters;
```
chkr1011, line 20: Please leave the parameter _null_ because otherwise the user can modify the instance of the "empty" dictionary because it is not read only (immutable).

chkr1011 (changes requested):

Source/MQTTnet.Extensions.Rpc/Options/TopicGeneration/TopicGenerationContext.cs
Temppus: Thanks @chkr1011 for review.
Would it be ok if I changed type for parameters from `IDictionary` to `IReadOnlyDictionary` ?

chkr1011: That would indeed circumvent the issue but I am afraid that maybe some users want to change these values and use it as a way to return data to the caller as well. So I would leave it null here. If you want to avoid checking for null all the time (what I can understand) you may pass an "Empty" immutable instance on your own.

Temppus: Made parameters nullable as suggested.

chkr1011 (approved):

PR #1803, opened 2023-07-29
Samples that show concurrent processing

ramonsmits:
Samples that show concurrent processing with:

- a concurrency limit
- without auto ACK

### DIFF
--- a/Samples/Client/Client_Subscribe_Samples.cs
+++ b/Samples/Client/Client_Subscribe_Samples.cs
@@ -5,8 +5,8 @@
 // ReSharper disable UnusedType.Global
 // ReSharper disable UnusedMember.Global
 // ReSharper disable InconsistentNaming
+// ReSharper disable UnusedMember.Local
 
-using MQTTnet;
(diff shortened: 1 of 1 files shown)


Samples/Client/Client_Subscribe_Samples.cs
```diff
@@ -177,4 +177,60 @@ public static async Task Subscribe_Topic()
             response.DumpToConsole();
         }
     }
-}
+
+    public static void ConcurrentProcessingWithLimit(CancellationToken shutdownToken, IMqttClient mqttClient)
```
chkr1011, line 181: Please apply snake case here for the method name because it gets printed to the console. All other methods are using snake case in order to improve reading in the console.

chkr1011, line 181: Please make sure that the sample can be executed in the console app. This requires that the method has no parameters. You may use hivemqtt or mosquitto public broker in this sample.

chkr1011 (changes requested):

Issue #1810, opened 2023-08-06
Option `WithAutoReconnectDelay` fails to reconnect when unauthorized.


jechtom:
Authorization failure breaks the managed client and it will never try to reconnect.

This happens if I use `ManagedMqttClient` with the feature `AutoReconnectDelay` ([tuned on by default](https://github.com/dotnet/MQTTnet/blob/9295626503231e50310aba104ed94e216ba4e95a/Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClientOptions.cs#L15)) together with client option `WithoutThrowOnNonSuccessfulConnectResponse` (expected to be [turned on by default in the future](https://github.com/dotnet/MQTTnet/blob/9295626503231e50310aba104ed94e216ba4e95a/Source/MQTTnet/Client/Options/MqttClientOptionsBuilder.cs#L84)).

### Which component is your bug related to?

- Client  (4.2.1.781, latest on 2023-08-06)
- ManagedClient (4.2.1.781, latest on 2023-08-06)

## To Reproduce

1. Have any MQTT server that requires authentication.
2. Use (currently latest) `MQTTnet.Extensions.ManagedClient 4.2.1.781`.
3. Run this code:
```csharp
            var mqttClientOptions = new ManagedMqttClientOptionsBuilder()
                //.WithAutoReconnectDelay(TimeSpan.FromSeconds(5)) <- no needed, turned on by default
                .WithClientOptions(b => b
                    .WithCredentials("aaaaaaaaaa invalid credentials")
                    .WithoutThrowOnNonSuccessfulConnectResponse() // will be enabled by default in the future versions
                    .WithConnectionUri("wss://address-of-you-mqtt-server")
                ).Build();

            var mqttClient = new MqttFactory().CreateManagedMqttClient();
            await mqttClient.StartAsync(mqttClientOptions).ConfigureAwait(false);
            await Task.Delay(Timeout.Infinite);
```
5. Run the code. 
6. Client will contact the server only once and will not retry.

### Expected behavior
The client should try to reconnect in regular intervals (5s by default).
Instead if will internally fail on error: 
```
Not allowed to connect while connect/disconnect is pending.
```

## Full code example with logging

Full code:
```csharp
using MQTTnet.Extensions.ManagedClient;
using MQTTnet;
using System.Diagnostics;

namespace MqttReconnectTest
{
    internal class Program
    {
        static async Task Main(string[] args)
        {
[...]

PR #1811, opened 2023-08-06
#1810 Close connection after failed connect.

jechtom:
If client option `WithoutThrowOnNonSuccessfulConnectResponse` is turned on (and it is expected to be [turned on by default in the future](https://github.com/dotnet/MQTTnet/blob/9295626503231e50310aba104ed94e216ba4e95a/Source/MQTTnet/Client/Options/MqttClientOptionsBuilder.cs#L84)), it will keep client in the _Connecting_ state after failed authentication attempt.

That can cause unexpected behavior, including #1810.

If option `WithoutThrowOnNonSuccessfulConnectResponse` is not turned on, it will work as expected, because the internal client [closes the connection in catch block](https://github.com/dotnet/MQTTnet/blob/9295626503231e50310aba104ed94e216ba4e95a/Source/MQTTnet/Client/MqttClient.cs#L179).

This fix will close the connection correctly if authentication failed.

Tested manually with test code #1810.

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -140,6 +140,7 @@ namespace MQTTnet.Client
                 if (connectResult.ResultCode != MqttClientConnectResultCode.Success)
                 {
                     _logger.Warning("Connecting failed: {0}", connectResult.ResultCode);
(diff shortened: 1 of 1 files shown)


jechtom:
Seems like bug in `MqttClient` in the method `ConnectAsync`. I tried to fix it in #1810.

PR #1848, opened 2023-09-23
Adds WithCaFile TlsOption to validate the connection with  X509ChainPolicy.CustomRootTrust 

ridomin:
# Summary

Most MQTT clients from Paho (python, Java, GO) and even mosquitto-clients, allow to specify a `CaFile` to connect to TLS endpoints protected with certificates issued by a private CA.

A good example is `test.mosquitto.org:8883` that requires https://test.mosquitto.org/ssl/mosquitto.org.crt to validate the server connection

# Details

- Adds a `CaFile` option to TlsOptions
- Adds a `WithCaFile(string caFile)` to TlsOptionsBuilder
- Adds a new CertificateValidationHandler to configure the X509 chain without revocation.
- Adds a new sample, including the test.mosquitto.org CaFile (this might require to be updated when it gets expired in 2030 :) )

CertificateValidationHandler cannot be set within `WithCaFile` to avoid overriding the handler.

Not use sure if the `#IF DEFS` can be optimized to other platforms, since the custom chain validator was added in .NET 5, but targeting .NET6 since .NET5 is already out of support.


Source/MQTTnet/Client/Options/MqttClientCaFileCertificateValidationHandler.cs
```diff
@@ -0,0 +1,55 @@
...
+                {
+                    string caFile = cvArgs.ClientOptions.TlsOptions.CaFile;
+                    if (!string.IsNullOrEmpty(caFile) && File.Exists(caFile))
+                    {
+                        X509Certificate2Collection caCerts = new X509Certificate2Collection();
+#if NET6_0_OR_GREATER
+                        caCerts.ImportFromPemFile(caFile);
+                        cvArgs.Chain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
+                        cvArgs.Chain.ChainPolicy.CustomTrustStore.AddRange(caCerts);
+                        cvArgs.Chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
```
ridomin, line 30: private CAs are very likely to do not offer revocation endpoints.

Source/MQTTnet/Client/Options/MqttClientTlsOptionsBuilder.cs
```diff
@@ -131,6 +137,17 @@ public MqttClientTlsOptionsBuilder WithCipherSuitesPolicy(EncryptionPolicy encry
             _tlsOptions.EncryptionPolicy = encryptionPolicy;
             return this;
         }
+
+        public MqttClientTlsOptionsBuilder WithCaFile(string caFile)
+        {
+            if (_tlsOptions.CertificateValidationHandler != null)
```
ridomin, line 143: WithCaFile needs to configure the CertificateValidationHandler, if it was manually assigned this throws an exception.

ridomin:
.. apologies for the 4 commits, ready to review now

PR #1851, opened 2023-09-29
Support custom CA chain validation

rido-min:
# Summary

Most MQTT clients from Paho (python, Java, GO) and even mosquitto-clients, allow to specify a `CaFile` to connect to TLS endpoints protected with certificates issued by a private CA.

A good example is `test.mosquitto.org:8883` that requires https://test.mosquitto.org/ssl/mosquitto.org.crt to validate the server connection

# Details

- Adds a `CertificationAuthoritiesFile` option to TlsOptions
- Adds a `WithCertificationAuthoritiesFile(string pemFile)` to TlsOptionsBuilder
- Configures SslStream with CustomTrust (.NET 7+ only)
- Adds a new sample, including the test.mosquitto.org (this might require to be updated when it gets expired in 2030 :) )

Overseeds https://github.com/dotnet/MQTTnet/pull/1848


rido-min:
How can I fix the build error? 

`D:\a\MQTTnet\MQTTnet\certificate.snk' -- Invalid public key`

Source/MQTTnet/Client/Options/MqttClientTlsOptions.cs
```diff
@@ -50,8 +50,13 @@ public sealed class MqttClientTlsOptions
 
 #if NET48 || NETCOREAPP3_1_OR_GREATER
         public SslProtocols SslProtocol { get; set; } = SslProtocols.Tls12 | SslProtocols.Tls13;
+
 #else
         public SslProtocols SslProtocol { get; set; } = SslProtocols.Tls12 | (SslProtocols)0x00003000 /*Tls13*/;
 #endif
+
+#if NET7_0_OR_GREATER
+        public string CertificationAuthoritiesFile { get; set; }
```
krwq, line 59: I'd recommend to make this API always visible but throw on the setter for <net7

Source/MQTTnet/Implementations/MqttTcpChannel.cs
```diff
@@ -115,15 +115,30 @@ public async Task ConnectAsync(CancellationToken cancellationToken)
...
+                            CertificateRevocationCheckMode = _tcpOptions.TlsOptions.IgnoreCertificateRevocationErrors ? 
+                                                                    X509RevocationMode.NoCheck : 
+                                                                    _tcpOptions.TlsOptions.RevocationMode,
                             TargetHost = targetHost,
                             CipherSuitesPolicy = _tcpOptions.TlsOptions.CipherSuitesPolicy,
                             EncryptionPolicy = _tcpOptions.TlsOptions.EncryptionPolicy,
                             AllowRenegotiation = _tcpOptions.TlsOptions.AllowRenegotiation
                         };
+#if NET7_0_OR_GREATER
+                        if (!string.IsNullOrEmpty(_tcpOptions.TlsOptions.CertificationAuthoritiesFile))
```
krwq, line 127: another option: `File.Exists`

krwq (approved):

rido-min:
@chkr1011 can you provide an ETA on when this PR will be reviewed?

Source/MQTTnet/Client/Options/MqttClientTlsOptionsBuilder.cs
```diff
@@ -132,5 +132,13 @@ public MqttClientTlsOptionsBuilder WithCipherSuitesPolicy(EncryptionPolicy encry
             return this;
         }
 #endif
+
+#if NET7_0_OR_GREATER
+        public MqttClientTlsOptionsBuilder WithCertificationAuthoritiesFile(string pemFile)
```
rido-min, line 137: another option will be `WithCaFile` is shorter -not as descriptive as the other methods in this class - but is aligned with other mqtt tooling.

logicaloud:
I think it would be good to support handling of certificate authorities directly within the MQTTnet library.

Some observations and questions:

1. Naming should probably be `...CertificateAuthorities...` instead of `...CertificationAuthorities...`?
2. Should the target conditional be `NET5_0_OR_GREATER`? Or are there features used that are not available since .NET 5? I'm aware that .NET 5 is out of support, but this would be in keeping with other conditionals used elsewhere supporting older .NET versions (and .NET 6 is still in support).
3. My preference would be to pass the certificate authorities as a list or collection of `X509Certificate2` to keep things in memory. Certificates may originate from a configuration file, database, of otherwise different format not readily available as a `.crt` file on disk. If the certificate authority certificates are indeed already available as a suitable file, then it is a small burden for the MQTTnet library user to load the file into a list or collection of X509Certificate2 certificates for handover to `WithCertificateAuthorities`. If the the certificates are not already in a suitable file, then passing a  filename could become a pain point, because the MQTTnet library user needs to convert the format into the `.crt` file format _and_ ensure that a there is a writable file system location available just for handover.

rido-min:
thanks @logicaloud  for you comments:

1. The name was suggested by @CiPop, maybe we could use `WithTrustChain` ?
2. IIRC The API we are using to use a custom trust store is available in .NET7+ only, but let me double check if this could also work in .NET5+
3. Good point, I'm updating the PR to accept a X509CertCollection

rido-min:
@logicaloud with this https://github.com/dotnet/MQTTnet/pull/1851/commits/0b5dff52ce526341cb02c0049631b3294625e192 I've addressed your comments:

1. Renamed to ` public MqttClientTlsOptionsBuilder WithTrustChain(X509Certificate2Collection chain)`
2. Unfortunately [SslClientAuthenticationOptions.CertificateChainPolicy](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslclientauthenticationoptions.certificatechainpolicy?view=net-7.0) requires .NET7
3. Agree, the user should provide the caChain as a cert collection without imposing any crt/pem format in a file

logicaloud:
Thank you for looking into this, @rido-min. Regarding naming, `WithTrustChain` seems more technical, `WithCertificateAuthorities` may need less explanation in the context of MQTT. Either way, it is not up to me to accept pull requests, so best to wait now until @chkr1011 has time to review and comment.

rido-min:
> WithTrustChain seems more technical,

Well, keep in mind this is on top of MqttTlsOptions, so I guess the TrustChain concept will be easier to get.

Source/MQTTnet/Client/Options/MqttClientTlsOptions.cs
chkr1011: Other properties in this project are usually not available because they are excluded by the compiler. I would like to stick to this approach and avoid changing the strategy here.

Source/MQTTnet/Client/Options/MqttClientTlsOptionsBuilder.cs
chkr1011: The other methods and properties in this project usually take the "long" version or adopt the same name when it is simply mapped to a property in the .NET framework.

rido-min: this comment is outdated, the last commit changed the signature to: `public MqttClientTlsOptionsBuilder WithTrustChain(X509Certificate2Collection chain)`

To align with .NET api this could be `.WithCustomTrustChain`

rido-min:
apologies for adding another commit https://github.com/dotnet/MQTTnet/pull/1851/commits/cc1ed2c128621246b8ce3e3f20501e10b03c3d0e

Now the chain validation uses `X509VerificationFlags.IgnoreEndRevocationUnknown` so users are not forced to completely disable revocation checks when using custom CAs.

@krwq do you mind taking another look?

chkr1011 (approved):

PR #1872, opened 2023-11-04
1846 server mqttserverstopasync doesnt indicate correct reason

chkr1011:
This PR adds new overloads for the server so that the user can send custom user properties or reason codes when stopping the server.
It also fixes the issue that the default used reason code was not correct when stopping the server.
A new property is also added to the server which allows to stop accepting new connections while the server is still running.

### DIFF
--- a/Source/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
+++ b/Source/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
@@ -4,7 +4,7 @@
 
 using System;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using MQTTnet.Adapter;
 using MQTTnet.Diagnostics;
 using MQTTnet.Implementations;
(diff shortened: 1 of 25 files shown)


Source/MQTTnet.AspnetCore/ServiceCollectionExtensions.cs
```diff
@@ -14,76 +14,102 @@ namespace MQTTnet.AspNetCore
...
+        }
+
+        public static void AddHostedMqttServer(this IServiceCollection services)
+        {
+            // The user may have these services already registered.
+            services.TryAddSingleton<IMqttNetLogger>(MqttNetNullLogger.Instance);
+            services.TryAddSingleton(new MqttFactory());
+
+            services.AddSingleton<MqttHostedServer>();
+            services.AddHostedService<MqttHostedServer>();
```
JanEggers, line 61: @chkr1011 this is bugged you will have 2 instances of MqttHostedServer. one when resolving the singleton and one IHostedService instance.  use AddHostedService(c => c.GetRequiredService<MqttHostedServer>())

chkr1011: Thanks @JanEggers for having a look at the change. Please see PR #1892 and let me know if it is correct now. I basically reverted the lines so that it is exactly the same as it was before.

PR #1877, opened 2023-11-11
V1.4.0.3

adrianstevens:


Source/MQTTnet/Implementations/CrossPlatformSocket.cs
```diff
@@ -23,7 +23,7 @@ public CrossPlatformSocket()
         {
             // Having this constructor is important because avoiding the address family as parameter
             // will make use of dual mode in the .net framework.
-            _socket = new Socket(SocketType.Stream, ProtocolType.Tcp);
+            _socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
```
chkr1011, line 26: Instead of making a change here you can set the address family to something different than "AddressFamily.Unspecified" in the client options. Then the code will use the other constructor.

Source/MQTTnet/Implementations/CrossPlatformSocket.cs
```diff
@@ -119,7 +119,7 @@ public async Task ConnectAsync(string host, int port, CancellationToken cancella
 #if NET452 || NET461
                     await Task.Factory.FromAsync(_socket.BeginConnect, _socket.EndConnect, host, port, null).ConfigureAwait(false);
 #else
-                    await _socket.ConnectAsync(host, port).ConfigureAwait(false);
+                    _socket.Connect(host, port);
```
chkr1011, line 122: What is the reason of not using the async variant here?

Source/MQTTnet/MQTTnet.csproj
```diff
@@ -1,41 +1,30 @@
...
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.3;netstandard2.0;netstandard2.1;netcoreapp3.1;net5.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">$(TargetFrameworks);net452;net461</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' AND '$(MSBuildRuntimeType)' != 'Core' AND '$(SolutionName)' != 'MQTTnet.noUWP' ">$(TargetFrameworks);uap10.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1</TargetFrameworks>
     <AssemblyName>MQTTnet</AssemblyName>
     <RootNamespace>MQTTnet</RootNamespace>
     <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
-    <Company />
+    <Company>Wilderness Labs, Inc</Company>
```
chkr1011, line 8: I assume that the PR is made against the wrong repo...

chkr1011 (changes requested):

PR #1882, opened 2023-11-16
Add ReasonCode to SendExtendedAuthenticationExchangeDataAsync

### LINKED ISSUE #1881: MqttClient.SendExtendedAuthenticationExchangeDataAsync() method doesn't set MqttAuthPacket.ReasonCode
### Describe the bug
MqttClient.SendExtendedAuthenticationExchangeDataAsync() method doesn't set MqttAuthPacket.ReasonCode. Because of this bug ReasonCode in AUTH packet is always 0 (Success)

- Client

### To Reproduce
1. Using this version of MQTTnet '4.3.1.873'.
2. Run this code
```csharp
await mqttClient.SendExtendedAuthenticationExchangeDataAsync(
    new MqttExtendedAuthenticationExchangeData() 
    {
        AuthenticationData = "AUTH-DATA", 
        ReasonCode = MQTTnet.Protocol.MqttAuthenticateReasonCode.ReAuthenticate
    });
```
3. On MQTT Broker side AUTH packet has ReasonCode 0 (Success) instead of 25 (ReAuthenticate)

### Expected behavior
On MQTT Broker side AUTH packet has ReasonCode 25 (ReAuthenticate)

rido-min:
Fixes #1881

### DIFF
--- a/.github/workflows/ReleaseNotes.md
+++ b/.github/workflows/ReleaseNotes.md
@@ -1,3 +1,4 @@
 * [Server] Fixed not working _UpdateRetainedMessageAsync_ public api (#1858, thanks to @kimdiego2098).
 * [Client] Added support for custom CA chain validation (#1851, thanks to @rido-min).
 * [Client] Fixed handling of unobserved tasks exceptions (#1871).
+* [Client] Fixed not specified ReasonCode when using _SendExtendedAuthenticationExchangeDataAsync_ (#1882, thanks to @rido-min).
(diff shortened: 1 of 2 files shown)


rido-min:
@chkr1011 is there a way to validate this fix from the MyGet feed?

chkr1011:
@rido-min The libraries cannot be signed because the branch is not created in this repo so that access to certain keys does not work and will thus the build will fail. I can create a separate branch or we merge this PR and create another one if further changes are required (I would create the branch then so that the Actions work properly).

PR #1891, opened 2023-12-06
1883 in a weak network environment the mqtt broker is unable to detect disconnections

chkr1011:
This PR changes the way a client is treated as "alive". The MQTT spec requires that only a _completely_ sent packet is considered when calculating the "inactive" time.


ohin:
I am wondering, how is the new DisconnectClientWhenReadingPayload flag working? I didn't find it referenced in anywhere in the code.

P-i-N:
@ohin I was wondering the same, but then I saw the usage of `DisconnectClientWhenReadingPayload` burried in collapsed part of the code in `MqttServerKeepAliveMonitor.cs`. You have to manually expand the lines in the UI to see it. Got me confused a bit too!

chkr1011:
The new flag is a leftover from the bugfix mentioned in this ticket. I will remove it soon because (as you already figured out) it is not used. It was an idea to have such a feature, but it turns out that enabling the feature will cause new problems.

From the MQTT spec the only correct way is to treat a fully transmitted packet as "traffic". So, the client must be able to send packets **within** the keep alive time. Clients have to specify a keep alive interval which also includes payload sizes and estimated network bandwidth etc.

PR #1894, opened 2023-12-08
1862 Add subscribe and unsubscribe results event for Managed Mqtt Client.

pdufrene:
Adding an event when ManagedMqttClient actually subscribe/unsubscribe to topics. This event has to be used in parallel with connected event to know the exact state.


pdufrene:
@dotnet-policy-service agree

Source/MQTTnet.Extensions.ManagedClient/ManagedMqttClient.cs
```diff
@@ -124,6 +125,12 @@ public ManagedMqttClient(IMqttClient mqttClient, IMqttNetLogger logger)
             remove => _synchronizingSubscriptionsFailedEvent.RemoveHandler(value);
         }
 
+        public event Func<SubscriptionsResultEventArgs, Task> SubscriptionsResultAsync
```
chkr1011, line 128: I recommend renaming this to something like "SubscriptionsChangedAsync" because it indicates WHEN the event is fired.

chkr1011 (changes requested):

PR #1954, opened 2024-03-19
Move "Client is not connected" error to it's own exception

marcpiulachs:
Create a custom exception to be able to catch the specific case when the mqtt client is not connected without having to compare the error message comparing strings.

In my application I am catching this type of exception and returning a custom localized error message so I have to catch the more generic MqttCommunicationException exception and compare the exception message for "The client is not connected." string, moving this error to it's own exception makes the solution much more elegant, hope it would be useful to others too.


marcpiulachs:
@dotnet-policy-service agree

PR #1974, opened 2024-04-22
rethrow MqttClientUnexpectedDisconnectReceivedException when Publish …

fazho:
rethrow MqttClientUnexpectedDisconnectReceivedException when QoS is 0 and Publish errors out because server disconnects the client.

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -38,6 +38,7 @@ namespace MQTTnet.Client
         // They contain the same values but the set is reduced in one case.
         int _disconnectReason;
         string _disconnectReasonString;
+        MqttDisconnectPacket _unexpectedDisconnectPacket;
         List<MqttUserProperty> _disconnectUserProperties;
 
         Task _keepAlivePacketsSenderTask;
@@ -122,6 +123,8 @@ namespace MQTTnet.Client
 
                 var adapter = _adapterFactory.CreateClientAdapter(options, new MqttPacketInspector(_events.InspectPacketEvent, _rootLogger), _rootLogger);
                 _adapter = adapter ?? throw new InvalidOperationException("The adapter factory did not provide an adapter.");
+
+                _unexpectedDisconnectPacket = null;
 
                 if (cancellationToken.CanBeCanceled)
                 {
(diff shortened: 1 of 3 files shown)


fazho:
@chkr1011 fyi

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -751,10 +755,22 @@ async Task<MqttClientPublishResult> PublishAtLeastOnce(MqttPublishPacket publish
...
+            try
+            {
+                // No packet identifier is used for QoS 0 [3.3.2.2 Packet Identifier]
+                await Send(publishPacket, cancellationToken).ConfigureAwait(false);
+
+                return MqttClientResultFactory.PublishResult.Create(null);
+            }
+            catch (Exception)
+            {
+                if (_unexpectedDisconnectPacket != null)
```
chkr1011, line 767: Does this only apply for QoS 0? I am wondering if this code must be also used for QoS 1 and 2?

fazho: I think it only applies to QoS 0 because for QoS 1, the client actually registers an awaitable and wait for the response packet, so it already throws an exception with the error code as of today. But for QoS 0, the client does not register any awaitable.  I guess this is by design because QoS 0 should not expect any response from server. But QoS 1 can always expect some response, whether its a PUBACK or DISCONNECT

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -751,10 +755,24 @@ async Task<MqttClientPublishResult> PublishAtLeastOnce(MqttPublishPacket publish
...
+                return MqttClientResultFactory.PublishResult.Create(null);
+            }
+            catch (Exception)
+            {
+                MqttDisconnectPacket localUnexpectedDisconnectPacket = _unexpectedDisconnectPacket;
 
-            return MqttClientResultFactory.PublishResult.Create(null);
+                if (localUnexpectedDisconnectPacket != null)
+                {
+                    throw new MqttClientUnexpectedDisconnectReceivedException(localUnexpectedDisconnectPacket);
```
chkr1011, line 771: I would pass the actual exception to the inner exception so that it is still accessible for the user.

chkr1011:
I will add the breaking change to the release notes. Breaking changes are happening from time to time and cannot be avoided. In this case we still get an exception. Only the the type is different now. I added a comment to the code.

chkr1011:
@fazho I made the change myself and also updated the release notes and added a comment. Please let me know if you agree with the changes (especially the comment). After that I will merge the PR.

fazho:
Hi @chkr1011 yes I still need it. Sorry for the delay. I was on vacation.

fazho:
@chkr1011 your changes look good to me. I just made another small change, setting the default inner exception as null in the constructor of MqttClientUnexpectedDisconnectReceivedException, so that if anyone explicitly creates an MqttClientUnexpectedDisconnectReceivedException object in their code (e.g., someone might mock the MQTTClient and create this exception in their unit tests), our changes in this PR will not break it.

I know this will be a breaking change either way, but just try to make the breaking part minimum.

fazho:
@chkr1011 feel free to revert my latest change if it does not make sense to you. I am ok either way.

PR #1987, opened 2024-05-02
fix: time out and cancel propagation in managed client

marve:
Eliminates race condition that may result in a "hang" on server shutdown. See added unit tests for example scenario.


marve:
@dotnet-policy-service agree company="Motorola Solutions"

PR #1988, opened 2024-05-02
fix(server): at-least-once messages received during subscription must…

marve:
… be delivered

### DIFF
--- a/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
+++ b/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
@@ -173,7 +173,6 @@ namespace MQTTnet.Server
                 throw new ArgumentNullException(nameof(subscribePacket));
             }
 
-            var retainedApplicationMessages = await _retainedMessagesManager.GetMessages().ConfigureAwait(false);
             var result = new SubscribeResult(subscribePacket.TopicFilters.Count);
 
             var addedSubscriptions = new List<string>();
--- a/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
+++ b/Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 using MQTTnet.Internal;
 using MQTTnet.Packets;
 using MQTTnet.Protocol;
+using static MQTTnet.Server.MqttClientSubscriptionsManager;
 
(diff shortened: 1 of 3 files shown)


Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
```diff
@@ -177,6 +178,13 @@ public async Task<SubscribeResult> Subscribe(MqttSubscribePacket subscribePacket
 
             var addedSubscriptions = new List<string>();
             var finalTopicFilters = new List<MqttTopicFilter>();
+            var atLeastOnceSubscriptionResults = new List<CreateSubscriptionResult>();
+
+            IList<MqttApplicationMessage> retainedApplicationMessages = null;
```
chkr1011, line 183: The code does not compile on my machine because the variable _retainedApplicationMessages_ is already defined above...

Source/MQTTnet/Server/Internal/MqttClientSubscriptionsManager.cs
```diff
@@ -172,11 +173,17 @@ public async Task<SubscribeResult> Subscribe(MqttSubscribePacket subscribePacket
...
 
-            var retainedApplicationMessages = await _retainedMessagesManager.GetMessages().ConfigureAwait(false);
             var result = new SubscribeResult(subscribePacket.TopicFilters.Count);
 
             var addedSubscriptions = new List<string>();
             var finalTopicFilters = new List<MqttTopicFilter>();
+            var atLeastOnceSubscriptionResults = new List<CreateSubscriptionResult>();
+
+            IList<MqttApplicationMessage> retainedApplicationMessages = null;
+            if (subscribePacket.TopicFilters.Any(f => f.QualityOfServiceLevel != MqttQualityOfServiceLevel.AtLeastOnce))
```
chkr1011, line 183: Why is this only relevant for at least once QoS? Isn't this also beneficial for QoS 0 and 2?
Or are you fixing a specific issue in your project here?

marve: Good question. The project wherein we discovered this bug uses solely `AtLeastOnce`. But I think you're right that it may fix potential issues with `AtMostOnce` and `ExactlyOnce` as well.

marve: @chkr1011 Is this PR sufficient as-is or are you requesting that I make changes?

PR #1993, opened 2024-05-11
fix build errors

SpringHgui:
fix build errors


SpringHgui:
@dotnet-policy-service agree

chkr1011:
Thank you for your work here but I was doing the same thing already :face_with_peeking_eye:

Issue #1943, opened 2024-03-04
Security alerts on MQTTnet.AspNetCore


milbrandt:
### Describe the feature request

Analyzing our application with Mend.io revelated 4 security alerts by transitive libraries brought with MQTTnet.AspNetCore.

|Library|Severity|Vulnerability|
|---|---|---|
|system.net.websockets.websocketprotocol.4.5.1|High|CVE-2019-0564|
|microsoft.aspnetcore.websockets.2.2.0|Medium|CVE-2019-0548|
|microsoft.aspnetcore.websockets.2.2.0|High|CVE-2019-0564|
|newtonsoft.json.11.0.2|High|CVE-2024-21907|

### Which project is your feature request related to?

MQTTnet.AspNetCore 4.3.3.952

### Describe the solution you'd like

Upgrading required libraries in the NuGet spec. My temprorary workaround is to add the following package references to my project file:

    
    <ItemGroup>
        <PackageReference Include="Microsoft.AspNetCore.WebSockets" Version="2.2.1" />
        <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
        <PackageReference Include="System.Text.Encodings.Web" Version="8.0.0" />
    </ItemGroup>

### Describe alternatives you've considered

Upgrading higher level NuGets to fix the issues, as microsoft.aspnetcore.http.connections.1.1.0 depends on multiple of these libraries.

microsoft.aspnetcore.http.connections.1.1.0 -> microsoft.aspnetcore.websockets.2.2.0 -> system.net.websockets.websocketprotocol.4.5.1
microsoft.aspnetcore.http.connections.1.1.0 ->newtonsoft.json.11.0.2

SeppPenner:
We should maybe check these dependencies within the project itself.

With the following settings in the projects

```csproj
<NoWarn>NU1803;NU1901;NU1902</NoWarn>
<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
<NuGetAuditMode>all</NuGetAuditMode>
<NuGetAudit>true</NuGetAudit>
```

this can be checked automatically within Visual studio.

**Explanation:**
* `<NoWarn>NU1803;NU1901;NU1902</NoWarn>`:
--> NU1803 suppresses errors if an internal NuGet feed is used without HTTPS (Guess, we don't need this here)
--> NU1901 and NU1902 suppress package vulnerabilites with low impact (Might be useful if no patches are there yet?)

[...]

PR #2016, opened 2024-06-04
Version5 auditmode

SeppPenner:
Fixes https://github.com/dotnet/MQTTnet/issues/1943 as requested in https://github.com/dotnet/MQTTnet/issues/1943#issuecomment-2132133261.

@chkr1011 I still have an issue I don't understand:

MQTTnet.Benchmarks throws:

```log
CS8892
Die Methode "TestingPlatformEntryPoint.Main(string[])" wird nicht als Einstiegspunkt verwendet,
weil ein synchroner Einstiegspunkt "Program.Main(string[])" gefunden wurde.
MQTTnet.Benchmarks (net8.0)	D:\Source\MQTTnet.Benchmarks\obj\Debug\net8.0\TestPlatformEntryPoint.cs
Die Methode "TestingPlatformEntryPoint.Main(string[])" wird nicht als Einstiegspunkt verwendet,
weil ein synchroner Einstiegspunkt "Program.Main(string[])" gefunden wurde.
```

Do you have an idea here?

Let me know if I need to change something here.

### DIFF
--- a/Samples/MQTTnet.Samples.csproj
+++ b/Samples/MQTTnet.Samples.csproj
@@ -9,7 +9,11 @@
         <DisableImplicitAspNetCoreAnalyzers>true</DisableImplicitAspNetCoreAnalyzers>
         <EnableNETAnalyzers>false</EnableNETAnalyzers>
         <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
-        <NoWarn>1591;NETSDK1138</NoWarn>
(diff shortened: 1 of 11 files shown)


chkr1011:
I also got this error while working on the version5 branch and it seems to me that Benchmark.NET generates its own main method since a certain version but we still have our own so that we can pick the right benchmark etc. We may need to check the best practices of Benchmark.NET and see how we should structure the benchmarks. We may also remove the benchmarks from the project completely because they are not used very often and can be created on demand without being part of the repository.

chkr1011 (approved):

PR #2018, opened 2024-06-10
Publish stored messages using Enumerator

nedimtokic:
When using the MqttManagedClient, the function LoadQueuedMessagesAsync retrieves all data from the user, and when we have a large amount of data, we can issue a high memory consumption.

This PR removes adding loaded data into queue and sends it just before starting publishing new data.


nedimtokic:
@dotnet-policy-service agree

nedimtokic:
@dotnet-policy-service agree

PR #2022, opened 2024-06-16
Topic template extension maintenance

simonthum:
Some docs and maintenance.

* Adds a README to the extension and its nuget
* Provide more examples
* Minor fixes and enhancements

Mainly, existing examples are modified to use the extension as is intended. I hope that's ok.

Also, parameter values are checked more thoroughly and extension methods are more complete now.

I considered extensions for MqttFactory as other extensions do, but that did not seem to add value.


simonthum:
Build failure seems unrelated.

chkr1011 (approved):

PR #2023, opened 2024-06-17
fix Managed_Client_Subscribe_Samples.Connect_Client sample:

JaggerJo:
- Result was never logged (`bool subscribed`)
- `subscribed` was never set, as it was not passed as a reference but as a value.


JaggerJo:
@dotnet-policy-service agree

JaggerJo:
@chkr1011 can you take a look?

chkr1011 (approved):

PR #2025, opened 2024-06-18
fix LoadingRetainedMessageAsync not exec

pigwing:
because IHostedService StartAsync before startup Configure 
see https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-8.0#ihostapplicationlifetime


pigwing:
@dotnet-policy-service agree

chkr1011 (approved):

PR #2026, opened 2024-06-18
Switch SubscribeAsync/UnsubscribeAsync to IEnumerable<string>

drelidan7:
The function currently requires an ICollection<string>, but doesn't actually depend on the ICollection<string> interface.  Relax the requirement to be IEnumerable<string> for better flexibility.


drelidan7:
@dotnet-policy-service agree

chkr1011 (approved):

PR #2046, opened 2024-07-14
Support for ReadOnlySequence as payload

mregen:
Support `ReadOnlySequence` as a payload in the MQTTnet client, for the new version5 release.

As a result, perfomance is affected by additional copies and memory is fragment by the additional random allocation of large buffers.

This is a minimal version without lifetime management in the receive code path to get started. 
Since the publish codepath can manage the lifetime of the buffers externally, all publishing applications can already benefit from the implementation.
To minimize the impact on existing code the PayloadSegment setter is kept to convert an ArraySegment to a ReadOnlySequence implicitly, allowing most of the tests to compile unchanged. 

Added also a few tests to show how to use a RecyclableMemoryStream.

see #1917 and #1918 with the issue explained.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
   sign:
     needs: build
     runs-on: windows-latest # Code signing must run on a Windows agent for Authenticode signing (dll/exe)
-
+    if: github.repository == 'dotnet/MQTTnet'
     steps:
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v4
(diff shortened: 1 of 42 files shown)


chkr1011 (approved):

mregen:
Hi @chkr1011, thanks for the code review! 
Once merged, with a signed preview package I can start testing version5 in our various projects and end 2 end tests. 
Cheers, Martin

.github/workflows/ci.yml
```diff
@@ -51,7 +51,7 @@ jobs:
   sign:
     needs: build
     runs-on: windows-latest # Code signing must run on a Windows agent for Authenticode signing (dll/exe)
-
+    if: github.repository == 'dotnet/MQTTnet'
     steps:
```
mregen, line 55: maybe also not run it from a PR, to avoid supply chain attacks that produce signed packages

PR #2102, opened 2024-11-12
Fix ignored exceptions in MqttHostedServer startup

beppemarazzi:
The exceptions raised during the startup of `MqttHostedServer` are completely ignored.
Nothing on the logs, the application continue to run, and the `MqttServer.StartedAsync` event is never raised leaving the application in an inconsistent state...
I found this issue because in my application I'm also using the Tcp endpoint (on the default port 1883), but if the port is already in use obviously the Tcp listener fails with an exception.

This PR fixes the issue shutting down the application in this case.
If the user has a service registration for `Microsoft.Extensions.Logging.ILogger` (quite common in real life scenario) an error is also logged.

### DIFF
--- a/Source/MQTTnet.AspnetCore/MqttHostedServer.cs
+++ b/Source/MQTTnet.AspnetCore/MqttHostedServer.cs
@@ -7,6 +7,8 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using MQTTnet.Diagnostics.Logger;
 using MQTTnet.Server;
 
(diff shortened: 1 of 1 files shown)


xljiulang:
I think the correct implementation should be: await the  MqttServer.StartAsync() behavior in BackgroundService so that the exception can be propagated. The specific implementation is similar to https://github.com/xljiulang/MQTTnet/blob/master/Source/MQTTnet.AspnetCore/Internal/AspNetCoreMqttHostedServer.cs#L39

beppemarazzi:
@xljiulang 
> I think the correct implementation should be: await the MqttServer.StartAsync() behavior in BackgroundService so that the exception can be propagated.

This was also my first thought.
And it was exactly the earlier (up to 2020) implementation (see https://github.com/dotnet/MQTTnet/blob/66048931a2bae650582e638b010cee0104a797f8/Source/MQTTnet.AspnetCore/MqttHostedServer.cs) but with https://github.com/dotnet/MQTTnet/commit/20eda969213f2ca1ffcafed8657974b97d3be325#diff-7489ffb431847a1474c7a8fa6dcd02fb1ed29ec54a204f6637789014ed277fb6R21-R27 the implementation was changed (i don't know the reason. Probably @JanEggers knows?...)
And moving forward with this one https://github.com/dotnet/MQTTnet/blob/e18a91a4b59bc312aa9acf9be575401d07b793e4/Source/MQTTnet.AspnetCore/MqttHostedServer.cs the implementation was changed again dispatching the `MqttServer.StartAsync() ` after the complete stratup of the Host is notified by `IHostApplicationLifetime` callback. See #2025 for details.

If i understand correctly this is a sort of chicken-and-egg problem: the `MqttServer` must start after the host because of #2025 BUT if we want to rely on the failure logic at host startup in case of exception during the `MqttServer` initialization we cannot wait...

xljiulang:
If you start MqttnetServer in the IHostedService.StartAsync method, there will generally be no problems. The worst case is that you unfortunately use MqttTcpServerAdapter, which increases the service startup time. If it is used as a windows service process, it may time out and fail.

xljiulang:
If you want to delay starting MqttnetServer, you should fix MqttConnectionHandleras follows:
```c#
    public override async Task OnConnectedAsync(ConnectionContext connection)
    {
        var clientHandler = ClientHandler;
        if (clientHandler == null)
        {
            connection.Abort();
            _logger.Publish(MqttNetLogLevel.Warning, nameof(MqttConnectionHandler), $"{nameof(MqttServer)} has not been started yet.", null, null);
            return;
        }
        ...
}
```

beppemarazzi:
@xljiulang 
Could you take a look at my last commit?
I just realizer it's IMHO the most obvious neat, clean KISS solution: relying on standard built-in `BackgroundService` logic!!!

beppemarazzi:
@JanEggers just rebased to current master and added two simple test
there is also complier constant  (to be removed before merge) just to illustrate the test failing with current behavior

beppemarazzi:
Just squashed and rebased onto current master (v5). 
@JanEggers any chance to merge this?

PR #2106, opened 2024-11-18
Set the default value of SslProtocol to SslProtocols.None.

xljiulang:
SslProtocols.None: 
Allows the operating system to choose the best protocol to use, and to block  protocols that are not secure. Unless your app has a specific reason not to,  you should use this field.

### DIFF
--- a/Source/MQTTnet.Server/Options/MqttServerTlsTcpEndpointOptions.cs
+++ b/Source/MQTTnet.Server/Options/MqttServerTlsTcpEndpointOptions.cs
@@ -22,7 +22,7 @@ namespace MQTTnet.Server
 
         public bool CheckCertificateRevocation { get; set; }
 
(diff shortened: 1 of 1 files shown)


Source/MQTTnet.Server/Options/MqttServerTlsTcpEndpointOptions.cs
```diff
@@ -22,7 +22,7 @@ public MqttServerTlsTcpEndpointOptions()
 
         public bool CheckCertificateRevocation { get; set; }
 
-        public SslProtocols SslProtocol { get; set; } = SslProtocols.Tls12;
```
chkr1011, line 25: Please add a summary comment including a reference to the MSDN so that users will understand the reason for "None" which looks like that SSL is not used at all.

chkr1011 (changes requested):

PR #2111, opened 2024-11-20
Added EndPoint RemoteEndPoint {get;} property to channel

xljiulang:
Each internal channel only retains the `EndPoint RemoteEndPoint {get;}` property, 
and the external event parameter type still retains the `string Endpoint {get;}` property.

```c#
public EndPoint RemoteEndPoint { get; }

[Obsolete("Use RemoteEndPoint instead.")]
public string Endpoint => RemoteEndPoint?.ToString();
```

### DIFF
--- a/Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
+++ b/Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
@@ -19,22 +19,18 @@ namespace MQTTnet.Server.Internal;
 public sealed class MqttClientSessionsManager : ISubscriptionChangedNotification, IDisposable
 {
     readonly Dictionary<string, MqttConnectedClient> _clients = new(4096);
-
     readonly AsyncLock _createConnectionSyncRoot = new();
-
     readonly MqttServerEventContainer _eventContainer;
     readonly MqttNetSourceLogger _logger;
     readonly MqttServerOptions _options;
-
     readonly MqttRetainedMessagesManager _retainedMessagesManager;
     readonly IMqttNetLogger _rootLogger;
-
     readonly ReaderWriterLockSlim _sessionsManagementLock = new();
 
     // The _sessions dictionary contains all session, the _subscriberSessions hash set contains subscriber sessions only.
     // See the MqttSubscription object for a detailed explanation.
     readonly MqttSessionsStorage _sessionsStorage = new();
-    readonly HashSet<MqttSession> _subscriberSessions = new();
+    readonly HashSet<MqttSession> _subscriberSessions = [];
 
     public MqttClientSessionsManager(MqttServerOptions options, MqttRetainedMessagesManager retainedMessagesManager, MqttServerEventContainer eventContainer, IMqttNetLogger logger)
     {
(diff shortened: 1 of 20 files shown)


Source/MQTTnet.Server/Internal/MqttConnectedClient.cs
```diff
@@ -227,26 +228,26 @@ async Task HandleIncomingPublishPacket(MqttPublishPacket publishPacket, Cancella
...
+                    Session.EnqueueControlPacket(new MqttPacketBusItem(pubAckPacket));
+                    break;
+                }
             case MqttQualityOfServiceLevel.ExactlyOnce:
-            {
-                var pubRecPacket = MqttPubRecPacketFactory.Create(publishPacket, dispatchApplicationMessageResult);
-                Session.EnqueueControlPacket(new MqttPacketBusItem(pubRecPacket));
-                break;
-            }
+                {
```
chkr1011, line 242: It seems your code style settings are different and thus the switch statements are reformatted without any reason.

Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
```diff
@@ -671,39 +671,39 @@ static bool ShouldPersistSession(MqttConnectedClient connectedClient)
...
+                }
 
             case MqttProtocolVersion.V311:
-            {
-                // MQTT 3.1.1 section 3.1.2.4: persist only if 'not CleanSession'
-                //
-                // If CleanSession is set to 1, the Client and Server MUST discard any previous Session and start a new one.
-                // This Session lasts as long as the Network Connection. State data associated with this Session MUST NOT be
-                // reused in any subsequent Session [MQTT-3.1.2-6].
+                {
```
chkr1011, line 693: It seems your code style settings are different and thus the switch statements are reformatted without any reason.

chkr1011 (changes requested):
Good feature. :+1: Just a few minor things to change in my opinion.

Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
xljiulang: > It seems your code style settings are different and thus the switch statements are reformatted without any reason.看来您的代码风格设置不同，因此 switch 语句毫无理由地被重新格式化。

Yes, this code formatting style issue bothers me and I don’t know how to deal with it.

chkr1011: Do you use Rider or ReSharper? The current style is configured for these tools. If you do not use them, it might be the reason.

xljiulang: No, I am only using the community preview version of vs2022, all with default settings.

chkr1011: OK then it makes sense that the code style is not applied. A solution may be to migrate to an "editorconfig" file. 
Rider is also free since a few weeks for noncommercial use. Maybe you want to give it a try :smile: 
I will apply the code style to this branch so that we can go ahead.

chkr1011 (approved):

Issue #2113, opened 2024-11-21
About the memory error problem of InjectApplicationMessage of MqttServer in client.WithCleanSession(false)


xljiulang:
### Describe the bug
When I was doing performance optimization extension for MqttServer's InjectApplicationMessage , I used `ArrayPool<byte>` memory and unexpectedly found that the test [QoS_Tests.Preserve_Message_Order_For_Queued_Messages](https://github.com/dotnet/MQTTnet/blob/master/Source/MQTTnet.Tests/Server/QoS_Tests.cs#L115)  failed because I cleared and recycled poolPayload after `await MqttServer.InjectApplicationMessage( poolPayload)` . I think there is no problem with my operation timing, and this behavior should be allowed.

If I use the byte[] poolPayload that I manually new, this test is successful. The reason is that the caller provides a payload in an independent space for InjectApplicationMessage, and this payload will only be used by MqttServer in the future until it is recycled by GC.

### Which component is your bug related to?

- Server

### To Reproduce
Modify the InjectApplicationMessage code of MqttServerExtensions and run the test Preserve_Message_Order_For_Queued_Messages.

```c#
public static async Task InjectApplicationMessage(
    this MqttServer server,
    string topic,
    string payload = null,
    MqttQualityOfServiceLevel qualityOfServiceLevel = MqttQualityOfServiceLevel.AtMostOnce,
    bool retain = false)
{
    ArgumentNullException.ThrowIfNull(server);
    ArgumentNullException.ThrowIfNull(topic);

    var payloadBuffer = Encoding.UTF8.GetBytes(payload ?? string.Empty);

    var message = new MqttApplicationMessageBuilder()
        .WithTopic(topic)
        .WithPayload(payloadBuffer)
        .WithQualityOfServiceLevel(qualityOfServiceLevel)
        .WithRetainFlag(retain)
        .Build();

    await server.InjectApplicationMessage(new InjectedMqttApplicationMessage(message));

    // clear the buffer
    payloadBuffer.AsSpan().Clear();
}
```
### Expected behavior
After await InjectApplicationMessage(), the memory reference to the payload must be released.

PR #2109, opened 2024-11-18
Reduce memory allocation of DecodePublishPacket.

xljiulang:
Now, memory allocation is reduced to 50%.
Job=.NET 8.0  Runtime=.NET 8.0  
 
| Method                        | PayloadSize | Mean      | Error     | StdDev    | Ratio | RatioSD | Gen0       | Gen1      | Allocated | Alloc Ratio |
|------------------------------ |------------ |----------:|----------:|----------:|------:|--------:|-----------:|----------:|----------:|------------:|
| **Send_1000_Messages_New** | **8192**        |  **8.727 ms** | **0.1393 ms** | **0.1368 ms** |  **1.00** |    **0.00** |  **1875.0000** |   **62.5000** |   **8.55 MB** |        **1.00** |
| Send_1000_Messages_Old    | 8192        | 18.182 ms | 0.1461 ms | 0.1220 ms |  2.08 |    0.03 |  3593.7500 |  250.0000 |  16.89 MB |        1.98 |
|                               |             |           |           |           |       |         |            |           |           |             |
| **Send_1000_Messages_New** | **65536**       | **40.791 ms** | **0.8041 ms** | **1.8475 ms** |  **1.00** |    **0.00** | **11076.9231** | **3307.6923** |  **63.22 MB** |        **1.00** |
| Send_1000_Messages_Old    | 65536       | 32.204 ms | 0.6323 ms | 0.7997 ms |  0.79 |    0.04 | 21200.0000 | 7066.6667 | 127.05 MB |        2.01 |

PR #2103, opened 2024-11-14
Refactoring MQTTnet.AspNetCore

xljiulang:
After several submissions, I have implemented the following refactorings. See also #2101

### IMqttBuilder
IMqttBuilder is used to provide common functions for creating MqttServer and MqttClient, such as logging.
```c#
public interface IMqttBuilder
{
    IServiceCollection Services { get; }
}
``` 

Extensions
```c# 
public static IMqttBuilder UseNullLogger(this IMqttBuilder builder);

public static IMqttBuilder UseLogger<TLogger>(this IMqttBuilder builder)where TLogger : class, IMqttNetLogger; 
```

### IMqttServerBuilder
```c#
public interface IMqttServerBuilder : IMqttBuilder
{
}
```
Extensions
```c# 
public static IMqttServerBuilder AddMqttServer(this IServiceCollection services);
    
public static IMqttServerBuilder ConfigureMqttServer(this IMqttServerBuilder builder, Action<MqttServerOptionsBuilder> configure);

public static IMqttServerBuilder ConfigureMqttServerStop(this IMqttServerBuilder builder, Action<MqttServerStopOptionsBuilder> configure);

public static IMqttServerBuilder AddMqttTcpServerAdapter(this IMqttServerBuilder builder); 
```

### IMqttClientBuilder
```c#
public interface IMqttClientBuilder: IMqttBuilder
{ 
}
```
Extensions
```c# 
public static IMqttClientBuilder AddMqttClient(this IServiceCollection services);

public static IMqttClientBuilder UseAspNetCoreMqttClientAdapterFactory(this IMqttClientBuilder builder);

public static IMqttClientBuilder UseMqttClientAdapterFactory<TMqttClientAdapterFactory>(this IMqttClientBuilder builder) where TMqttClientAdapterFactory : class, IMqttClientAdapterFactory;
```
 
### New implementation of IMqttClientAdapterFactory
The SocketConnection that the original MqttClientConnectionContextFactory depends on is a copy of the early code of the kestrel project. It is now deleted and replaced by the IConnectionFactory service of Asp.NetCore to create ConnectionContext instead of SocketConnection.

IConnectionFactory only provides an asynchronous ConnectAsync() method. In order to call this asynchronous method, we need to modify the CreateClientAdapter method of IMqttClientAdapterFactory to `ValueTask<IMqttChannelAdapter> CreateClientAdapterAsync()`.

[...]

PR #2115, opened 2024-11-22
Retain the copy message #2113

xljiulang:


xljiulang:
I found that this is a stubborn bug. After making the above fix, if I modify the logic of MqttChannelAdapter receiving data packets, the bug still exists. So once a packet is received, we have to copy its payload out instead of using the Pool memory?
```c#
async Task<ReceivedMqttPacket> ReceiveAsync(CancellationToken cancellationToken)
{
   if (cancellationToken.IsCancellationRequested)
   {
       return ReceivedMqttPacket.Empty;
   }

   var readFixedHeaderResult = await ReadFixedHeaderAsync(cancellationToken).ConfigureAwait(false);

   if (cancellationToken.IsCancellationRequested)
   {
       return ReceivedMqttPacket.Empty;
   }

   if (readFixedHeaderResult.IsConnectionClosed)
   {
       return ReceivedMqttPacket.Empty;
   }

   var fixedHeader = readFixedHeaderResult.FixedHeader;
   if (fixedHeader.RemainingLength == 0)
   {
       return new ReceivedMqttPacket(fixedHeader.Flags, ReadOnlySequence<byte>.Empty, 2);
   }

   var bodyLength = fixedHeader.RemainingLength;

   // Return and clear the previous body buffer
   _bodyOwner?.Dispose();

   // Re-rent a body buffer
   _bodyOwner = BufferOwner.Rent(bodyLength);
   var body = _bodyOwner.Buffer;

   var bodyOffset = 0;
   var chunkSize = Math.Min(ReadBufferSize, bodyLength);

   do
   {
       var bytesLeft = bodyLength - bodyOffset;
       if (chunkSize > bytesLeft)
       {
           chunkSize = bytesLeft;
       }

       var readBytes = await _channel.ReadAsync(body, bodyOffset, chunkSize, cancellationToken).ConfigureAwait(false);

       if (cancellationToken.IsCancellationRequested)
       {
           return ReceivedMqttPacket.Empty;
       }

       if (readBytes == 0)
       {
           return ReceivedMqttPacket.Empty;
       }

       bodyOffset += readBytes;
   } while (bodyOffset < bodyLength);

   PacketInspector?.FillReceiveBuffer(body.AsSpan(0, bodyLength));

   var bodySegment = body.AsMemory(0, bodyLength);
   var bodySequence = new ReadOnlySequence<byte>(bodySegment);
   return new ReceivedMqttPacket(fixedHeader.Flags, bodySequence, fixedHeader.TotalLength);
}
```

```c#
private sealed class BufferOwner : IDisposable
{
  private bool _disposed = false;

  public byte[] Buffer { get; private set; }

  public static BufferOwner Rent(int bufferSieze)
  {
      return new BufferOwner()
      {
          Buffer = ArrayPool<byte>.Shared.Rent(bufferSieze)
      };
  }

  public void Dispose()
  {
      if (!_disposed)
      {
          _disposed = true;
          ArrayPool<byte>.Shared.Return(Buffer);
      }
  }
}
```

Source/MQTTnet/Options/MqttClientOptions.cs
```diff
@@ -225,4 +225,9 @@ public sealed class MqttClientOptions
     ///     Do not change this value when no memory issues are experienced.
     /// </summary>
     public int WriterBufferSizeMax { get; set; } = 65535;
+
+    /// <summary>
+    /// When enabled, received ApplicationMessage will be deep cloned and enqueued.
+    /// </summary>
+    public bool ReceivedApplicationMessageQueueable { get; set; } = true;
```
xljiulang, line 232: I want to keep this switch configuration in Options and let users choose. Can we radically change the default to false to reduce GC pressure?

Source/MQTTnet/MqttClient.cs
```diff
@@ -947,7 +971,15 @@ async Task TryProcessReceivedPacket(MqttPacket packet, CancellationToken cancell
             switch (packet)
             {
                 case MqttPublishPacket publishPacket:
-                    EnqueueReceivedPublishPacket(publishPacket);
+                    if (Options.ReceivedApplicationMessageQueueable)
+                    {
+                        // publishPacket must be copied
+                        EnqueueReceivedPublishPacket(publishPacket.Clone());
```
xljiulang, line 977: The publishPacket here must be cloned, otherwise problems will occur after replacing it with IMqttClientAdapterFactory of mqttnet.asp.
We can consider using Pool memory for MQTTnet.Adapter.MqttChannelAdapter to avoid unnecessary cloning of publishPacket.

xljiulang:
Hi @chkr1011 
This PR is at the final stage, please take the time to review and give comments. After this PR is completed, I will focus on #2109, and finally #2103.

PR #2117, opened 2024-12-01
Improve session message injection

AntonSmolkov:
Hi there!  
I'm trying to build a scalable, persistent MQTT server around RabbitMQ, using the MQTTNet server as a frontend.

While doing this, I found that there is currently no proper way to correlate injected QoS 1 messages with the packets received by the `ClientAcknowledgedPublishPacketAsync` event. So, there is no way for me to pass the PUBACK further to RabbitMQ's message acknowledgment.

Additionally, I noticed that `DeliverApplicationMessageAsync()` method may hang indefinitely when `MqttPendingMessagesOverflowStrategy.DropNewMessage` is used, because in the case of an output buffer overflow, the `packetBusItem` being awaited is dropped from the queue without the Task's `Result` or `Exception` being set.

So, in this PR:

- I added `MqttPublishPacket` as a result of message injection methods, to be able to correlate them by packet id later.
- Fixed the potential infinite wait in the `DeliverApplicationMessageAsync()` method.  
- Added some tests.

Thanks for the great project, btw!


AntonSmolkov:
@dotnet-policy-service agree

Source/MQTTnet.Server/Status/MqttSessionStatus.cs
```diff
@@ -40,22 +41,57 @@ public Task DeleteAsync()
...
 
-    public Task DeliverApplicationMessageAsync(MqttApplicationMessage applicationMessage)
+    /// <summary>
+    /// Delivers an application message immediately to the session.
+    /// </summary>
+    /// <param name="applicationMessage">The application message to deliver.</param>
+    /// <returns>
+    /// A task that represents the asynchronous operation. The result contains the delivered MQTT publish packet.
+    /// </returns>
+    public async Task<MqttPublishPacket> DeliverApplicationMessageAsync(MqttApplicationMessage applicationMessage)
```
chkr1011, line 51: Usually, the internal MQTT packets are not exposed in public APIs of this project. Thats why this method also accepts a MqttApplicationMessage instead of an PublishPacket etc. In my opinion there is also not much data the user might be interested in because the user already passes the application message including its payload, properties etc. so that the returning object of this and the method below should only contain data which is generated inside the library like the packet ID.

chkr1011 (changes requested):
Nice feature! But I recommend changing the public API a bit.

Source/MQTTnet.Server/Status/MqttSessionStatus.cs
AntonSmolkov: I also had thoughts about returning only the `Packet Identifier`, but QoS 0 messages don’t have one, making it impossible to track which specific message was evicted from the buffer via the `QueuedApplicationMessageOverwrittenAsync` event.  

`MqttPacket` is already publicly exposed via the `QueuedApplicationMessageOverwrittenAsync` and `ClientAcknowledgedPublishPacketAsync` event arguments. I thought it would be consistent to expose it in the session injection methods as well. 

The only alternative I see for now is to add an `InternalId` property, which will be generated for every MqttPacket (including QoS 0). This internalId can then be used as the return value for the injection methods.

What do you think?

AntonSmolkov: Update:
It seems that using `MqttPacketWithIdentifier` as a return type might be a compromise alternative._(Can be compared with eventArgs by packetId or by object hashcode)_ 
I changed it to this type while waiting for your response.

chkr1011: > `MqttPacket` is already publicly exposed via the `QueuedApplicationMessageOverwrittenAsync` and `ClientAcknowledgedPublishPacketAsync` event arguments. I thought it would be consistent to expose it in the session injection methods as well.
> 
This is a mistake which should not be repeated :smile: 

> The only alternative I see for now is to add an `InternalId` _(guid?)_ property, which will be generated for every MqttPacket (including QoS 0). This internalId can then be used as the return value for the injection methods.
> 
> What do you think?

Other APIs have a dedicated return type for that.
```csharp
public class DeliverMqttApplicationMessageResult
{
    public int PacketId {get; set;}
}

```

With this approach we can also add new features as properties, methods etc. later. When returning the internal packet, the API is closed and cannot be extended without introducing breaking changes.

> I also had thoughts about returning only the `Packet Identifier`, but QoS 0 messages don’t have one, making it impossible to track which specific message was evicted from the buffer via the `QueuedApplicationMessageOverwrittenAsync` event _(it's possible now by comparing objects)_.

I don't understand the issue in identifying the message because the methods return a task with the result. If you await them, you know that the affected message is exactly the one you passed as a parameter. I don't see the need of comparing an object. At least for our use case in this ticket. Or is there something I am missing here?

When it comes to events, we may add a _Tag_ property to the application message which is not processed by the library in any way but gives the user a chance to store some correlation data etc. for custom needs.

AntonSmolkov: > I don't understand the issue in identifying the message because the methods return a task with the result.

`TryEnqueueApplicationMessage` don't return task. The obsolete one `EnqueueApplicationMessageAsync` returns task, but this is task only about enqueuing itself. 

So if we use `TryEnqueueApplicationMessage` for qos0 message, we are unable to track whether our messages was evicted from the outboud queue eventually.

AntonSmolkov: It might be not the corner case to bother much.  I think i'll go with `DeliverMqttApplicationMessageResult` by now.

Source/MQTTnet.Server/Status/MqttSessionStatus.cs
```diff
@@ -40,22 +41,64 @@ public Task DeleteAsync()
...
 
         var packetBusItem = new MqttPacketBusItem(MqttPublishPacketFactory.Create(applicationMessage));
         _session.EnqueueDataPacket(packetBusItem);
 
-        return packetBusItem.WaitAsync();
+        var mqttPacket = await packetBusItem.WaitAsync();
+
+        var result = new InjectMqttApplicationMessageResult()
+        {
+            PacketIdentifier = ((MqttPacketWithIdentifier)mqttPacket).PacketIdentifier
```
chkr1011, line 63: When you create the PublishPacket above and keep the variable, there is no need to cast it here.

Source/MQTTnet.Server/Status/MqttSessionStatus.cs
```diff
@@ -40,22 +41,64 @@ public Task DeleteAsync()
...
+    /// </returns>
+    public async Task<InjectMqttApplicationMessageResult> DeliverApplicationMessageAsync(MqttApplicationMessage applicationMessage)
     {
         ArgumentNullException.ThrowIfNull(applicationMessage);
 
         var packetBusItem = new MqttPacketBusItem(MqttPublishPacketFactory.Create(applicationMessage));
         _session.EnqueueDataPacket(packetBusItem);
 
-        return packetBusItem.WaitAsync();
+        var mqttPacket = await packetBusItem.WaitAsync();
```
chkr1011, line 59: Please add .ConfigureAwait(false).

chkr1011 (changes requested):
Just a few minor improvements.

Source/MQTTnet.Server/Status/MqttSessionStatus.cs
AntonSmolkov: Fair. Did it

chkr1011 (approved):
Good work :+1:

AntonSmolkov:
@chkr1011 Thanks!

The CI-workflow is not approved yet. It this intentional?

AntonSmolkov:
@chkr1011  Sorry to bother you again, but the CI has failed on the "sign nugets step". :))

PR #2120, opened 2024-12-03
Add Directory.Build.props.

xljiulang:
I added a Directory.Build.props file to the solution and moved the common PropertyGroups and ItemGroups in each project into this file.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,46 @@
+<Project>
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
+        <Product>MQTTnet</Product>
+        <Authors>The contributors of MQTTnet</Authors>
+        <Company>The contributors of MQTTnet</Company>
+        <Copyright>Copyright (c) .NET Foundation and Contributors</Copyright>
+
+        <DelaySign>false</DelaySign>
+        <SignAssembly>false</SignAssembly>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+        <EnableNETAnalyzers>false</EnableNETAnalyzers>
+        <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
+
+        <NeutralLanguage>en-US</NeutralLanguage>
+        <NoWarn>1591;NETSDK1138;NU1803;NU1901;NU1902</NoWarn>
+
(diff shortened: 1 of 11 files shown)


Directory.Build.props
```diff
@@ -0,0 +1,46 @@
...
+        <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
+
+        <NeutralLanguage>en-US</NeutralLanguage>
+        <NoWarn>1591;NETSDK1138;NU1803;NU1901;NU1902</NoWarn>
+
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <NuGetAuditMode>all</NuGetAuditMode>
+        <NuGetAudit>true</NuGetAudit>
+        <NuGetAuditLevel>low</NuGetAuditLevel>
+        <AnalysisLevel>latest-Recommended</AnalysisLevel>
```
xljiulang, line 22: In my IDE, the latest-Recommended value causes hundreds of syntax errors.

Directory.Build.props
```diff
@@ -0,0 +1,46 @@
...
+
+        <PackageLicenseExpression>MIT</PackageLicenseExpression>
+        <PackageProjectUrl>https://github.com/dotnet/MQTTnet</PackageProjectUrl>
+        <RepositoryUrl>https://github.com/dotnet/MQTTnet.git</RepositoryUrl>
+        <RepositoryType>git</RepositoryType>
+
+        <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
```
xljiulang, line 29: After enabling ContinuousIntegrationBuild, it affects the lower breakpoint of the upper-level project in Debug configuration. Can we add a condition here to enable it only in Release configuration?

Directory.Build.props
mregen: https://developercommunity.visualstudio.com/t/Code-Rules-that-were-Informational-are-n/10789409#TPIN-N10795393

Directory.Build.props
chkr1011: Enabling _ContinuousIntegrationBuild_ for RELEASE builds only sounds reasonable to be me because DEBUG builds are usually not built using the CI. They are only built on local machines.

Directory.Build.props
chkr1011: I fixed all errors I get. Now it builds fine.

PR #2121, opened 2024-12-05
Fixed the problem that the statistics

### LINKED ISSUE #2104: About statistics of _bytesSent and _bytesReceived
I am not sure if the following behavior is a bug. 
In [MqttChannelAdapter.cs#L166](https://github.com/dotnet/MQTTnet/blob/master/Source/MQTTnet/Adapter/MqttChannelAdapter.cs#L166) and [MqttChannelAdapter.cs#L239](https://github.com/dotnet/MQTTnet/blob/master/Source/MQTTnet/Adapter/MqttChannelAdapter.cs#L239),  sent is counted counts to _bytesReceived, and t receive counts to _bytesSentand .

xljiulang:
Currently, the values ​​of _bytesSent and _bytesReceive are opposite on both the client and server. This PR is used to fix #2104.


chkr1011:
This was by design and the property reflects the perspective of the server. To the server SENT 100 MB and the client RECEIVED 100 MB.

PR #2123, opened 2024-12-07
Added UserName property to various Server EventArgs.

xljiulang:
In server events, directly carrying the client's UserName may bring some convenience to upper-level developers, and adding the UserName will not cause any computing overhead.

### DIFF
--- a/Source/MQTTnet.Server/MqttServer.cs
+++ b/Source/MQTTnet.Server/MqttServer.cs
@@ -255,6 +255,7 @@ public class MqttServer : Disposable
 
         return _clientSessionsManager.DispatchApplicationMessage(
             injectedApplicationMessage.SenderClientId,
+            senderUserName: null,
             sessionItems,
             injectedApplicationMessage.ApplicationMessage,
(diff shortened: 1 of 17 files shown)


Source/MQTTnet.Server/MqttServer.cs
```diff
@@ -255,6 +255,7 @@ public Task InjectApplicationMessage(InjectedMqttApplicationMessage injectedAppl
 
         return _clientSessionsManager.DispatchApplicationMessage(
             injectedApplicationMessage.SenderClientId,
+            senderUserName: null,
```
chkr1011, line 258: Consider exposing the SenderUserName as well in the injected application message. Otherwise, the new properties are null in these cases.

chkr1011 (changes requested):
Good work. I have 1 improvement.

Source/MQTTnet.Server/MqttServer.cs
xljiulang: OK. It's done.

PR #2126, opened 2024-12-09
Some socket option should be set to the accepted Socket, not the listening Socket.

xljiulang:
When I adapted MqttServerTcpEndpointBaseOptions in MQTTnet.AspnetCore to the socket accepted by kestrel, I unexpectedly found that the MqttTcpServerListener of MQTTnet.Server was adapted to the socket used for listening in its implementation.
I confirmed that the NoDelay, LingerState and KeepAlive* options are all used to set the socket accepted.


xljiulang:
After analyzing the packets captured by Wireshark, I found that the settings of the socket to be listened to also automatically affect the sockets it accepts. So there is no problem here, it is just a blind spot in my personal knowledge.

Issue #2108, opened 2024-11-18
KeepAlive not triggering DisconnectEvent in due time


Erw1nT:
Setup:
1. Setup a hotspot with your smartphone
2. Connect to the hotspot network with your MQTTNet client (ManagedClient, Websocket, TLS)
3. Start the MQTTNet client and connect to a broker
4. Disable the mobile data of the smartphone so that the hotspot is available but is not connected to the internet

KeepAlive is set to 5s.

Expected Behaviour:
After KeepAlive * 1.5 (= around 7-8s), the Disconnect-Event should be fired so that it can be handled.
(i.e. show a Connection-Lost screen)

Actual Behaviour:
After a random time (sometimes between 20-30s, sometimes never) the Disconnect-Event is fired and can be handled.

MQTTnet Version: 
- 4.3.7.1207

Component: 
- ManagedClient

We are running the MQTTnet client in the browser as a blazor wasm application.

Logs:
2024-11-18 11:33:24.293 +01:00 [WRN] MqttClient / Communication error while sending/receiving keep alive packets
2024-11-18 11:33:24.295 +01:00 [VRB] MqttClient / Stopped sending keep alive packets
2024-11-18 11:33:24.296 +01:00 [INF] MqttClient / Disconnected.
2024-11-18 11:33:24.299 +01:00 [VRB] MqttClient / Stopped receiving packets
2024-11-18 11:33:24.299 +01:00 [ERR] Disconnected with reason: NormalDisconnection

Note to Logs:
The entries appear in correct order but start later than the expected 7-8s.

PR #2129, opened 2024-12-18
Corrected Cancellation Source for KeepAlive Messages

Erw1nT:
I found out that the issue found in #2108 likely comes due to the fact that the wrong duration is used for checking the keep alive interval.

In this test setup i connect to a public broker and use the approach described in #2108 to see if it is triggered within time.
https://github.com/Erw1nT/MqttKeepAlive/blob/master/Program.cs

Only when this line is added, the connection-changed/disconnected event is fired in time.
> .WithTimeout(TimeSpan.FromSeconds(KeepAliveTimeInSeconds))


Source/MQTTnet/MqttClient.cs
```diff
@@ -1018,10 +1018,10 @@ async Task TrySendKeepAliveMessages(CancellationToken cancellationToken)
 
                 if (timeWithoutPacketSent > keepAlivePeriod)
                 {
-                    using (var timeoutCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
+                    using (var keepAliveCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
                     {
-                        timeoutCancellationTokenSource.CancelAfter(Options.Timeout);
-                        await PingAsync(timeoutCancellationTokenSource.Token).ConfigureAwait(false);
+                        keepAliveCancellationTokenSource.CancelAfter(keepAlivePeriod);
```
chkr1011, line 1023: With this change the cancellation will happen earlier according to the keep alive period (15 seconds). Which is by default way smaller than the generic communication timeout (100 seconds).

But we already reached _timeWithoutPacketSent > keepAlivePeriod_ so that wait an additional of 15 seconds (so 30 in total).

Isn't it a better way to wait for _keepAlivePeriod / 2_ to match the calculation at the server instead?

chkr1011 (changes requested):
Please see my comment.

Source/MQTTnet/MqttClient.cs
chkr1011: @Erw1nT I made the change I proposed to get this out quickly. If you prefer another solution, please reopen the ticket and let's discuss another solution.

Erw1nT:
@chkr1011 Thank you for reviewing and improving the PR.

Erw1nT:
@chkr1011 We are currently using v4.3.6 in our application and rely on the ManagedMqttClient as of now.
Due to time/budget constraints it would be best for us, if this bugfix would be made available in a v4.3.8 so that we do not have to implement and test a re-connect logic that the ManagedClient offers.
Is that possible? 
Thanks in advance.

chkr1011:
I will create a port for version 4 soon.

Erw1nT:
@chkr1011 Have you had any time for the port yet. Would really appreciate it. (I can also see why it's not a high priority ;))

chkr1011:
@Erw1nT High priority does not mean it will be done tomorrow. For a project which is maintained by people in their free time, it means it is only one of the first things which is done as soon as the developers have some free time to work on it ;-)

A new PR is here: https://github.com/dotnet/MQTTnet/pull/2144
Please test latest version 4.3.8 (as soon as available) from myget feed and let me know if it works properly.

Erw1nT:
@chkr1011 It does work as expected now. Thank you for your support.

PR #2131, opened 2024-12-21
Add ability to get session from server by client id

AntonSmolkov:
Hi there!
Currently it's only possible to get all the sessions from the server object. This small PR also adds the ability to get a single session by it's id.
This is helpful in scenarious when the MQTTNet server is used as as a frontend and it's always needed to get the specific session and inject some messages to it.


chkr1011 (approved):

AntonSmolkov:
@chkr1011 Thanks! Could you please also merge https://github.com/dotnet/MQTTnet/pull/2117, it starts getting conflicts

PR #2168, opened 2025-03-09
Propose to replace AsyncLock with SemaphoreSlim

mregen:
Hi @chkr1011, sorry for my ignorance, but during debugging under high load I ended up sometimes with exceptions in AsyncLock. They may have been uncritical, but I realized the way AsyncLock is used can also be covered by SemaphoreSlim.
So I would propose to replace AsyncLock with SemaphoreSlim in the codebase and reduce the maintenace surface.
Unless there are specific perf or functional reasons to keep AsyncLock, please provide the rational.
Thanks,
Martin


chkr1011:
@mregen We started with _SemaphoreSlim_ but implemented the custom primitive to fix some issue we had with _SemaphoreSlim_. I will try to figure out the reason because I cannot remember it :smile:

mregen:
Thanks @chkr1011, would be interesting to know the issue. Apparently it can not be nested and other things.

chkr1011:
@mregen Here are at least two reasons why we created the custom async lock...

1. When disposing the _SemaphoreSlim_ (i.e. closing connection and thus deleting sessions etc.), all tasks/threads which are currently in _WaitAsync_ will never continue or throw.

2. When the _SemaphoreSlim_ is already disposed (i.e. same situation as above), subsequent calls to _Release_ (from some _finally_ blocks) will throw an ObjectDisposedException and there is no way to check if the _SemaphoreSlim_ is already disposed (apart from encapsulating it and adding this feature etc.) to avoid these calls.

I don't know if this reasons still apply to .NET 8+.

mregen:
> @mregen Here are at least two reasons why we created the custom async lock...
> 
> 1. When disposing the _SemaphoreSlim_ (i.e. closing connection and thus deleting sessions etc.), all tasks/threads which are currently in _WaitAsync_ will never continue or throw.

--> they should throw and that would be expected behaviour

> 2. When the _SemaphoreSlim_ is already disposed (i.e. same situation as above), subsequent calls to _Release_ (from some _finally_ blocks) will throw an ObjectDisposedException and there is no way to check if the _SemaphoreSlim_ is already disposed (apart from encapsulating it and adding this feature etc.) to avoid these calls.
> 

--> you may have noticed that I changed the code how the semaphore slim is entered to a `try/finally` pattern. The semaphore is waiting outside the try/finally block. 
In contrary with `using` the exception in the Release call can not be avoided.  

> I don't know if this reasons still apply to .NET 8+.

I think there is no change, but maybe how it was used previously was not quite right.

chkr1011:
@mregen Please see https://github.com/dotnet/MQTTnet/blob/b22b02a0b6db39e0674bbf27313bce7eabe881f8/Frameworks/MQTTnet.NetStandard/Internal/AsyncLock.cs

We started with the _SemaphoreSlim_ but ended up building our own primitive. I am afraid that we will introduce all issues again which led to the custom primitive.

> you may have noticed that I changed the code how the semaphore slim is entered to a try/finally pattern. The semaphore is waiting outside the try/finally block.

The issue is not the actual finally block in your changes. The issue is that the lock gets disposed from another thread instead while the current thread is waiting and was able to get into the _WaitAsync_ properly before.

mregen:
ok, sounds good...

chkr1011:
Do you have some more details regarding the exceptions you observed?
I would like to fix them.

Issue #2095, opened 2024-10-25
Error while authenticating. Extended authentication handler is not yet supported, with Azure Event Grid MQTT Broker


pre-subham:
### Describe the bug
We're facing this error: 
**Error while authenticating. Extended authentication handler is not yet supported. with the Latest MQTT net release and Azure Event Grid MQTT Broker.**
We have already tried the solution mentioned to recreate the certificate/user for the Broker, but this issue is frequent on two days basis on One particular Broker out of 3 setups. 
Everything is consistent across deployment but the issue is faced with one broker only.

### Which component is your bug related to?
- Client
- Server

### To Reproduce
1. Using this version of MQTTnet '...'.
     4.2.1.781, 4.3.7.1207
2. Run this code '....'.
```
mqttClientOptions = new MqttClientOptionsBuilder()
                    .WithClientId(MqttPublisherConstant.userId)
                    .WithTcpServer(MqttPublisherConstant.connectHost, MqttPublisherConstant.port)
                    .WithTls(new MqttClientOptionsBuilderTlsParameters
                    {
                        UseTls = true,
                        Certificates = certs
                    })
                    .WithCredentials(MqttPublisherConstant.userName)
```

### Expected behavior
Should connect without any issues.

### Screenshots
![Image](https://github.com/user-attachments/assets/31e7f9ea-cc2c-44dc-93a9-ee8ce7bf036c)

rido-min:
ExtendedAuthentication is supported with MQTT5, and usually used with OAUTH tokens, not sure what's the use case to use ExtendedAuth with certificates.

There is a sample to connect to [EventGrid with JWT tokens](https://github.com/Azure-Samples/MqttApplicationSamples/blob/main/scenarios/jwt_authentication/dotnet/jwt_authentication/Program.cs)  

note how ExtendedAuth is used in the  [WithJWT](https://github.com/Azure-Samples/MqttApplicationSamples/blob/e0411eef86958f5c91de63a0f5ecc845da88a97c/mqttclients/dotnet/MQTTnet.Client.Extensions/WithJWT.cs#L30-L35) extension method

pre-consultant:
What is the change required here for the use of certificate @rido-min. We have been running same for past 1 year, this error started coming now. We have not mentioned protocol, hence it's using default which should be v3.1.

chkr1011:
@pre-subham  It seems that this feature is broken (regardless of certificates or other variant). There is a handler for that but not used when connecting.
[...]

Issue #2029, opened 2024-06-26
Support AUTH packet exchange while connecting MqttClient


vishnureddy17:
Currently, it appears that if the client sends a `CONNECT` packet using MQTT 5 extended authentication (i.e., it sets the Authentication Method) and it receives an `AUTH` packet from the server instead of a `CONNACK` packet, it will throw an exception:

https://github.com/dotnet/MQTTnet/blob/e18a91a4b59bc312aa9acf9be575401d07b793e4/Source/MQTTnet/Client/MqttClient.cs#L466-L469

This means that the client cannot connect using extended authentication with AUTH packets being exchanged. It looks like this feature gap is not being tracked, so I am opening this issue to track it.

BillArmstrong:
This is related to issue #2095. That issue is closed and resolved with pull request #2132. This issue should be closed as it is resolved by the same pull request.

Issue #1805, opened 2023-08-02
NotSupportedException: Extended authentication handler is not yet supported


BEagle1984:
### Describe the bug
While connecting to emqx we often get the "Extended authentication handler is not yet supported" exception. My [library](https://github.com/BEagle1984/silverback) automatically retries after 500ms and it eventually successfully connects.
Of course, no setting is being changed in between the retries.

### Which component is your bug related to?
- Client

### To Reproduce
1. Using MQTTnet 4.2.1.781
2. Try to connect to EMQX from a couple of pods running on AWS EKS.
3. The first try often fails with a `NotSupportedException` (_Extended authentication handler is not yet supported_)

### Expected behavior
The connection always succeeds.

### Additional context / logging

This is the exception we get:
```batch
MQTTnet.Adapter.MqttConnectingFailedException: Error while authenticating. Extended authentication handler is not yet supported
 ---> System.NotSupportedException: Extended authentication handler is not yet supported
   at MQTTnet.Client.MqttClient.Authenticate(IMqttChannelAdapter channelAdapter, MqttClientOptions options, CancellationToken cancellationToken)
   --- End of inner exception stack trace ---
   at MQTTnet.Client.MqttClient.Authenticate(IMqttChannelAdapter channelAdapter, MqttClientOptions options, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.ConnectInternal(IMqttChannelAdapter channelAdapter, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.ConnectAsync(MqttClientOptions options, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.ConnectAsync(MqttClientOptions options, CancellationToken cancellationToken)
   at Silverback.Messaging.Broker.Mqtt.MqttClientWrapper.TryConnectClientAsync(Boolean isFirstTry, CancellationToken cancellationToken)
```

It's not thrown for all clients instantiated within the same application and not always for the same ones.

### Code example

The actual code uses [Silverback](https://github.com/BEagle1984/silverback), which wraps MQTTnet. The resulting MQTTnet related code is something like this:
```csharp
IMqttClient mqttClient = MqttFactory.CreateClient();
await mqttClient.ConnectAsync(options, cancellationToken);
```

The options passed look like this:
```batch
  {MQTTnet.Client.MqttClientOptions}
    AllowPacketFragmentation: true
    AuthenticationData: null
    AuthenticationMethod: null
[...]

PR #2171, opened 2025-04-07
Backport of: 2132 Fix support for enhanced authentication

BillArmstrong:
This PR is a backport of #2132 for the version 4 library.

This PR has a number of changes as compared to the PR on which it is based. Since this PR is a fix for the v4 library I stuck with the original "ExtendedAuthenticationExchange" naming convention that was used in the v4 library instead of the more correct "EnhancedAuthentication" naming convention used in the v5 library.

Additionally I tried to avoid any changes to existing interfaces/method signatures that would cause breaking changes. The PR on which these changes are based has many breaking changes as compared to the v4 library.

My company is unable to use the v5 library because we are constrained by having to support the .Net Framework, but we need to be able to authenticate clients using Enhanced Authentication.

This PR addresses the following issues:
* #2029
* #1805
* #2095

### DIFF
--- a/Source/MQTTnet/Client/MqttClient.cs
+++ b/Source/MQTTnet/Client/MqttClient.cs
@@ -89,6 +89,8 @@ namespace MQTTnet.Client
 
         public bool IsConnected => (MqttClientConnectionStatus)_connectionStatus == MqttClientConnectionStatus.Connected;
 
+        bool IsConnecting => (MqttClientConnectionStatus)_connectionStatus == MqttClientConnectionStatus.Connecting;
+
         public MqttClientOptions Options { get; private set; }
 
         public async Task<MqttClientConnectResult> ConnectAsync(MqttClientOptions options, CancellationToken cancellationToken = default)
@@ -304,7 +306,7 @@ namespace MQTTnet.Client
             }
 
             ThrowIfDisposed();
-            ThrowIfNotConnected();
+            ThrowIfNotConnectedOrConnecting();
 
             var authPacket = new MqttAuthPacket
             {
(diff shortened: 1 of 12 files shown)


BillArmstrong:
@dotnet-policy-service agree company="Quest Software"

Source/MQTTnet/Client/MqttClient.cs
```diff
@@ -454,27 +456,30 @@ async Task<MqttClientConnectResult> Authenticate(IMqttChannelAdapter channelAdap
                 var connectPacket = MqttPacketFactories.Connect.Create(options);
                 await Send(connectPacket, cancellationToken).ConfigureAwait(false);
 
-                var receivedPacket = await Receive(cancellationToken).ConfigureAwait(false);
-
-                switch (receivedPacket)
+                for (; ; )
```
chkr1011, line 459: What is the purpose of this loop? In v5 is also no loop required.

BillArmstrong: Thanks for the review. 

The loop is needed because when when the client sends the initial connection request, depending on what type of authentication is requested the server response will be either a MqttConnAckPacket or a MqttAuthPacket. The server will send a MqttAuthPacket if it needs to request additional authentication token data. Depending on the authentication mechanism, the server may send several MqttAuthPacket responses that the client will need to handle.

The client must respond to the MqttAuthPacket server requests until it receives a MqttConnAckPacket response.

In the Kerberos authentication sample in the MQTT v5 specification, the server sends two AUTH responses back to the client before finally sending a CONNACK response.

BillArmstrong:
```
cancellationToken.ThrowIfCancellationRequested();
```
is the first line within the loop.

BillArmstrong: @chkr1011 Do you have any additional comments on the PR?

PR #2175, opened 2025-04-11
Performance: Implemented simple topic subscription management in MqttClientSessionsManager.

AlexNik4:


AlexNik4:
@dotnet-policy-service agree

AlexNik4:
While this greatly improves performance for non-wildcard topics at the small cost of another index, there is still the issue of incorrectly excluding clients that care about this topic via wildcard only.

AlexNik4:
Ok. I think this version works for all cases. I am getting huge performance improvements when there are many clients subscribing to different non-wildcard topics. In the past 2k messages per second with 10k clients was failing. Now I can get 20k messages per second with 20k clients. This will not improve performance if all the clients are using wildcards. But it should not meaningfully degrade performance in that scenario either.

AlexNik4:
I can look into unit tests later. Would appreciate some feedback on this. Overall, this is just separating topics with and without wildcards so that we can perform additional optimizations for non wildcard topics.

Source/MQTTnet/MqttTopicFilterComparer.cs
```diff
@@ -174,5 +174,19 @@ public static unsafe MqttTopicFilterCompareResult Compare(string topic, string f
 
             return MqttTopicFilterCompareResult.NoMatch;
         }
+
+        public static bool ContainsWildcards(string topicFilter)
+        {
+            for (var i = 0; i < topicFilter.Length; i++)
+            {
+                var c = topicFilter[i];
+                if (c == MultiLevelWildcard || c == SingleLevelWildcard)
```
chkr1011, line 183: The MultiLevelWildcard (#) can only be at the end of the topic. So, we can check this once if the topic is at least 3 chars long (the char before the # must be a / in all cases).

chkr1011:
@logicaloud Please have a look at the change and let me know what you think.

logicaloud:
Looks feasible. @AlexNik4, maybe you could run benchmarks "MessageDeliveryBenchmark", "SubscribeBenchmark", and "UnsubscribeBenchmark" and compare the Before and After to cover other scenarios as well? 

So far application message handling was geared towards the "many publishers, few subscribers" scenario (see #1298 and "Other considerations"). Having lower performance for a "many subscribers" scenario is not unexpected and improvements here would be great.

Some random observations:
- The check for wildcards in `OnSubscriptionAdded` could be avoided by passing a list of `MqttSubscription` to the method instead of the list of strings. The `MqttSubscription` contains a `TopicHasWildcard` flag as well as the topic string. The `Subscribe` method would need to be modified to collect the `MqttSubscription` in `createSubscriptionResult` as `addedSubscriptions`.
- Should the `GetSimpleSubscribedTopics` property be called `SubscribedSimpleTopics`?

logicaloud:
Looks good to me. Up to @chkr1011 for review but here are some ideas regarding naming of variables that may convey the meaning a bit better. 

Instead of `_simpleTopicsToSession`: `_simpleTopicToSessions` or `_subscriberSessionsBySimpleTopic`

Instead of `out var sessionsWithSimpleTopics`: `out var simpleTopicSessions`

Issue #2196, opened 2025-08-13
Package https://www.nuget.org/packages/MQTTnet links to wrong license document


JensNordenbro:
NuGet should contain the license by content as described by https://github.com/dotnet/MQTTnet/blob/master/LICENSE during build. 

Today it links to https://licenses.nuget.org/MIT which has different text.

PR #2197, opened 2025-08-13
Use License file from code repo

JensNordenbro:
Mitigates #2196


chkr1011:
@JensNordenbro Could you please do the change for the other project files as well?

PR #2213, opened 2025-10-19
Expose local endpoint in channel adapter

chkr1011:
This PR exposes the local endpoint in the channel adapter in same way as the remote endpoint. This helps to identify the clients protocol and origin.


chkr1011:
I assume it will be released this month.

PR #2228, opened 2025-12-06
Allow user properties as pre-encoded UTF-8 binary buffer

koepalex:
Issue:
When extensively using MQTT v5 user properties, handling the key and value as string has two major downsides (allocation time and garbage collection). Keys can use string interning or defined as const but the values (when they are changing) can cause high allocation rate and latency spikes when the garbage collector has to collect them.

Proposal:
Allow value of MQTT v5 user properties to be added as bytes (pre-encoded UTF-8 binary buffer), using performance optimized dotnet types like ReadOnlyMemory<byte>. The same optimization was already implemented in version 5.x for the payload of the MQTT message itself.

Outlook:
The current implementation optimizes the MQTT message send path, it would be also possible to optimize the receive path and handling values of user properties only as ReadOnlyMemory<byte> and just allocate the string on demand. I didn't implement that yet, as this would require more API changes and might even have negative consequences (when user property values are read multiple times, caching would be required).

Pull Request contains the implementation of the proposal, including tests for binary user properties, a sample demonstrating publishing a message with binary user properties (`Samples/Client/Client_Publish_Samples.cs`) and minor fixes for consistency and correctness.

### DIFF
--- a/Samples/Client/Client_Subscribe_Samples.cs
+++ b/Samples/Client/Client_Subscribe_Samples.cs
@@ -7,6 +7,7 @@
 // ReSharper disable InconsistentNaming
 // ReSharper disable UnusedMember.Local
 
+using System.Text;
 using MQTTnet.Packets;
 using MQTTnet.Protocol;
 using MQTTnet.Samples.Helpers;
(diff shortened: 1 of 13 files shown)


koepalex:
There is a breaking change in dotnet 10, that doesn't allow VSTest runner when using dotnet test. It is recommended to switch to Microsoft.Test.Platform (MTP). I made the required changes in this PR, that the test github workflow can run successfully.

Details under: https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test

koepalex:
Avoid failing build workflow, by only signing the package when executing the pipeline from the main repo, not from a fork

koepalex:
Had to update a few more places, that also the samples and tests use only the nonobsolete methods

PR #2235, opened 2026-02-16
Fix statistics tracking for bytes sent and received in MqttChannelAdapter

RicoSuter:
The `BytesSent` and `BytesReceived` counters in `MqttChannelAdapter` are swapped — `ReceivePacketAsync` increments `_bytesSent` and `SendPacketAsync` increments `_bytesReceived`.


chkr1011:
This is intentional. It depends on the perspective of the usage of the data. When receiving data, it is the actual sent data by the client. And when the server is sending it is the received data of the client. So when I want to expose the received bytes of the client in the server I have to count the sent bytes and name them received bytes in the client statistics.

Issue #2174, opened 2025-04-11
Performance: Limited publish through-put when there are many idle clients


AlexNik4:
### Describe the bug
I can publish 25k messages per second to one client (QoS 1) if there are few other clients. But that number quickly diminishes as the number of other clients grows. The other clients are subscribed to unrelated topics. No wildcards are used at all.

### Which component is your bug related to?
- Client
- Server

### To Reproduce
Run the code below in release mode. This is hardware dependent but first try something like:
```
const int MsgPerSecond = 20000;
const int SubscriberClients = 100;
```
Observe in the console output that the publisher channel size is pretty much always at 0 and the received message count is increasing by the configured MsgPerSecond. Everything is working fine. Then try:
```
const int MsgPerSecond = 2000;
const int SubscriberClients = 10000;
```
You should notice that the publisher is no longer able to keep up. The channel size is continuously increasing in size, even though we are publishing 90% less messages and the other subscribes are not really 'relevant'.

### Expected behavior
I was hoping to support a scenario where there are many clients(10k) subscribed to unique topic like 'user/[username]' and be able to publish to a small group of those clients. Since the topics don't have any wildcards I expected this to scale at the speed of a hashmap lookup.

### Additional context / logging
If all the other clients are subscribed to a constant topic (different from the one that is actually published to), then there is no performance issue. Adding additional publishers to help did not scale either. Maybe this scenario is not really what MQTTNet was designed for, but it would be nice to be able to support this scenario along with the 'normal' cases. 

### Code example
```csharp
const int MsgPerSecond = 2000;
const int SubscriberClients = 10000;
const MqttQualityOfServiceLevel ServiceQuality = MqttQualityOfServiceLevel.AtLeastOnce;

MqttServerFactory serverFactory = new();

MqttServerOptions serverOptions = new MqttServerOptionsBuilder()
    .WithDefaultEndpoint()
    .Build();

var mqttServer = serverFactory.CreateMqttServer(serverOptions);
await mqttServer.StartAsync();

MqttClientFactory factory = new();
IMqttClient publisherClient = factory.CreateMqttClient();

var publisherOptions = new MqttClientOptionsBuilder()
            .WithTcpServer("localhost")
            .WithClientId("publisher")
            .WithKeepAlivePeriod(TimeSpan.FromSeconds(10))
            .WithCleanSession()
            .Build();

// Connect the publisher client
await publisherClient.ConnectAsync(publisherOptions);

[...]

PR #2238, opened 2026-03-03
Implement non-wildcard subscription management

jboshoven-emporia:
In a world where there are a ton of clients with many subscriptions that are not wildcard, it can get really expensive to loop through the entire list of subscribers. This change breaks the subscribers down into two lists, one for wildcards and one without. You can see in `DispatchApplicationMessage` that if there is a `_simpleTopicToSessions` it will be grabbed first and added to all wildcard sessions. If there are no wildcard sessions, this operation is exceedingly cheap. The rest of the change outside of this spot is pretty much just plumbing.

This change is based on this PR https://github.com/dotnet/MQTTnet/pull/2175 by AlexNik4 to try and solve this issue https://github.com/dotnet/MQTTnet/issues/2174.

This change also stops copying the hash set on publish. These sets have become copy on write and then the entire session management lock was removed at the expense of more garbage during subscribe/unsubscribe. This garbage was paid for every publish anyway.

### DIFF
--- a/Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
+++ b/Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
@@ -4,6 +4,7 @@
 
 using System.Collections;
 using System.Collections.Concurrent;
+using System.Collections.Frozen;
 using MQTTnet.Adapter;
 using MQTTnet.Diagnostics.Logger;
 using MQTTnet.Exceptions;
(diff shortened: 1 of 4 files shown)


Source/MQTTnet.Server/Internal/MqttClientSessionsManager.cs
```diff
@@ -161,15 +166,15 @@ public async Task<DispatchApplicationMessageResult> DispatchApplicationMessage(
                     await _retainedMessagesManager.UpdateMessage(senderId, applicationMessage).ConfigureAwait(false);
                 }
 
-                List<MqttSession> subscriberSessions;
-                _sessionsManagementLock.EnterReadLock();
-                try
+                MqttSessionLists subscriberSessions;
+                if (_simpleTopicToSessions.TryGetValue(applicationMessage.Topic, out var matchedSimpleTopicSessions))
```
jboshoven-emporia, line 170: This is the read that returns an iterator with a copy of the values.

jboshoven-emporia:
@dotnet-policy-service agree company="Emporia Energy"

PR #2239, opened 2026-03-03
Support for client specific messages

jboshoven-emporia:
This change adds DestinationClientId onto MqttApplicationMessages. If this value is non-null, only that specific subscriber will receive the message. This makes it possible to have a single topic shared by multiple clients.


insylogo:
Why not just retrieve just the client object you care about and call Deliver/Enqueue message to it?

var session = Server.GetSessionAsync(clientId) then session.TryEnqueueApplicationMessage or DeliverApplicationMessageAsync

jboshoven-emporia:
I love that idea, let me give it a shot.
I'll close the PR if that works.

jboshoven-emporia:
Thank you @insylogo that worked great.

PR #2241, opened 2026-03-18
Fix for ReadWriteLock

jboshoven-emporia:
The original ReaderWriterLock is not capable of spanning an async call. Unlocking from a different thread throws and then never unlockes it. 

I moved to SemaphoreSlim because it should be pretty lightly contended as it is a per-session and the reader/writer semantics will likely be overkill. This might be a performance problem and an async aware reader writer lock might be needed.

### DIFF
--- a/Source/MQTTnet.Server/Internal/MqttClientSubscriptionsManager.cs
+++ b/Source/MQTTnet.Server/Internal/MqttClientSubscriptionsManager.cs
@@ -25,7 +25,7 @@ public sealed class MqttClientSubscriptionsManager : IDisposable
     readonly Dictionary<string, MqttSubscription> _subscriptions = new Dictionary<string, MqttSubscription>();
 
     // Use subscription lock to maintain consistency across subscriptions and topic hash dictionaries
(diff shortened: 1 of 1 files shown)


chkr1011:
@jboshoven-emporia Do you have some performance comparisons (Benchmark.net)? I remember that the change to use a ReaderWriterLock was made when optimizing performance. I just want to make sure that there will be no performance decrease.

Source/MQTTnet.Server/Internal/MqttClientSubscriptionsManager.cs
jboshoven-emporia, line 235: This line is the line that fails if context returns on a different thread.

jboshoven-emporia:
This change isn't for performance, it's for correctness. There's an `await` inside the lock and if the thread that picks up the task is different than the one that locked it will throw.
If there's a reader/writer lock that can span `await` then that should probably be the one used.

jboshoven-emporia:
I updated the pr description to be more descriptive of what I'm trying to fix here. If there's an easy rw lock that works I'll happily swap it out.

PR #2242, opened 2026-04-09
Check if subscription object is disposed before access

fifieldt:
In production, the Meshtastic MQTT Server experienced many millions of ObjectDisposedException. In debugging, this appears to occur due to a race condition in MQtt subscription management in mqttnet.

This patch adds a flag to check whether a subscription object has been disposed before attempting to interact with it, rather than raising the exception.

Co-Authored-By: @caveman99


fifieldt:
(sorry, wrong fork - we're still debugging this)

PR #2245, opened 2026-04-22
Add EMQX-compatible delayed publish ($delayed/<seconds>/<topic>)

Romain-OD:
This PR adds a small opt-in extension package, README.md, that lets the broker hold a message and deliver it later using the EMQX convention:

```
$delayed/10/sensors/temp       → delivered in 10 seconds
$delayed/1761234567/cron/fire  → delivered at this Unix timestamp
```

Why
Delayed publish comes up a lot and isn't part of the MQTT spec. Shipping it as an EMQX-compatible topic convention makes it easy for people migrating from EMQX and stays out of the way for everyone else.

How it works
Entirely built on top of existing extension points — no changes to the core broker pipeline:
•	Hooks MqttServer.InterceptingPublishAsync, parses the $delayed/<interval>/... prefix, swallows the original publish (publisher still gets its PUBACK immediately).
•	A background worker holds pending messages in a PriorityQueue keyed on due-time and re-injects them via InjectApplicationMessage(InjectedMqttApplicationMessage, CancellationToken) when due.
•	Pluggable IDelayedMessageStore — ships with an in-memory default, ready for a SQLite/Redis implementation later.


Romain-OD:
@dotnet-policy-service agree

PR #2247, opened 2026-05-06
Client: Tolerate late acknowledgement packets (#2078, #2079)

### LINKED ISSUE #2078: MQTTClient Thread-safety: Publishing
### Describe your question
Continuing the question from this thread: https://github.com/dotnet/MQTTnet/issues/2077
Could you please dive a little deeper into why it is unsafe for multi-threading?
 
Also what would be your recommended workaround for my scenario? Should I implement locks on the MQTT clients or could you point me to a thread-safe MQTT client? Basically I want to implement a MQTT client cache so I don't need to repeatedly connect and disconnect an MQTT client. But this cache will be shared by multiple threads, meaning it's likely those threads could use the same MQTT client to publish messages simultaneously.

@chkr1011 

### Which project is your question related to?

- Client

### LINKED ISSUE #2079: Received packet 'PubAck: [PacketIdentifier=6538] [ReasonCode=Success]' at an unexpected time.
### Describe the bug
When I publish a relatively large number of messages on the same topic (the number varies) a small percentage of publish calls fail with MqttProtocolViolationException as described in the title.  For example, if I send 100,000 messages less than 100 fail, but they all fail with this exception.

I tried updating from 4.3.3.952 to 4.3.7.1207 but the problem persists.

- Client

### To Reproduce
1. Using this version of MQTTnet '...'. 4.3.7.1207
2. Run this code '....'. I have created a small test program, but it relies on some "wrapper" code that encapsulats the MQTTnet client. I may be able to reproduce the problem without that wrapper code if it would help.
3. With these arguments '....'.
MQTT client construction:

         _mqttFactory = new MqttFactory();
         _mqttClient = _mqttFactory.CreateMqttClient();

         var clientOptions = new MqttClientOptionsBuilder()
            .WithClientId("test program")
            .WithTcpServer("localhost")
            .WithProtocolVersion(MQTTnet.Formatter.MqttProtocolVersion.V500)
            .WithCleanSession()
            .WithWillTopic("test program/status")
            .WithWillQualityOfServiceLevel(MQTTnet.Protocol.MqttQualityOfServiceLevel.AtLeastOnce)
            .WithWillPayload(UTF8Encoding.UTF8.GetBytes("offline"))
            .WithWillRetain(true)
            .Build();

I am running the Mosquitto broker (version 2.0.14) locally.

The specific method that fails is IMqttClient.PublishAsync(MqttApplicationMessage)

Exception details:

2024-09-12 18:16:59.287 -07:00 [ERR] Unexpected exception occurred when publishing message to topic: notification/foo/bar error: Received packet 'PubAck: [PacketIdentifier=6538] [ReasonCode=Success]' at an unexpected time., will try again.
MQTTnet.Exceptions.MqttProtocolViolationException: Received packet 'PubAck: [PacketIdentifier=6538] [ReasonCode=Success]' at an unexpected time.
   at MQTTnet.Client.MqttClient.TryProcessReceivedPacket(MqttPacket packet, CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.PacketDispatcher.MqttPacketAwaitable`1.WaitOneAsync(CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.Request[TResponsePacket](MqttPacket requestPacket, CancellationToken cancellationToken)
   at MQTTnet.Client.MqttClient.PublishAtLeastOnce(MqttPublishPacket publishPacket, CancellationToken cancellationToken)

5. See error.

### Expected behavior
I expect all published messages to be published (so they can be received by subscribers).

araxis:
Fixes #2078
Fixes #2079

## Problem

When a `PublishAsync` / `SubscribeAsync` / `UnsubscribeAsync` call is cancelled or times out client-side, its awaitable is removed from `MqttPacketDispatcher`. If the broker''s matching `PUBACK` / `PUBCOMP` / `SUBACK` / `UNSUBACK` arrives slightly later it falls through the `default` case in `MqttClient.TryProcessReceivedPacket`, raises `MqttProtocolViolationException("Received packet ''...'' at an unexpected time.")`, and tears down the connection.

This surfaced as:
- **#2079** — sporadic `MqttProtocolViolationException` on a tiny fraction of QoS 1 publishes when sending high volumes (~100k messages).
- **#2078** — `MqttClient` appearing unsafe under concurrent publishes from multiple threads.

## Why #2078 is the same root cause (not a missing lock)

The two issues look different but share one bug. The client was already correctly synchronised at every relevant layer; concurrency just made the late-ack race far more likely to be hit.

What was already protected:

| Step | Protected | How |
| --- | --- | --- |
| `MqttPacketIdentifierProvider.GetNextPacketIdentifier` | yes | `lock (_syncRoot)` — each caller gets a unique ID |
| `MqttPacketDispatcher.AddAwaitable` | yes | `lock (_waiters)` |
| `MqttChannelAdapter.SendPacketAsync` | yes | `await _syncRoot.EnterAsync(...)` (`AsyncLock`) — wire bytes never interleave |
| Each caller awaits its own `MqttPacketAwaitable<T>` | yes | Per-call instance |
| Incoming acks routed by packet identifier | yes | `MqttPacketDispatcher.TryDispatch` matches by ID |

So two threads calling `PublishAsync` simultaneously was fine at the locking level. What broke under load was this sequence:

1. Thread A calls `PublishAsync(message, ctsA.Token)` with a tight timeout.
2. The awaitable for packet id `N` is registered, the PUBLISH goes on the wire.
3. `ctsA` fires before the broker''s PUBACK arrives - the awaitable is removed.
4. PUBACK for `N` arrives.
5. **Old behaviour:** dispatcher finds no waiter → `MqttProtocolViolationException` → connection torn down.
6. **Thread B''s** unrelated in-flight publish, sharing the same connection, now also fails.

That is why users perceived "the client breaks when used from multiple threads": one caller''s cancellation killed everyone else''s connection. The same race produced #2079 in a single-threaded high-volume scenario, where a small fraction of publishes get cancelled by their own timeout.

Adding a `SemaphoreSlim` around `PublishAsync` would have been the wrong fix - it would serialise publishes (hurting throughput) without addressing the actual bug, since the late-ack race exists regardless of how many threads are involved.

## Fix

`MqttClient.TryProcessReceivedPacket` now matches `PUBACK` / `PUBCOMP` / `SUBACK` / `UNSUBACK` explicitly. If no awaiter is registered the packet is logged as a warning and dropped, rather than treated as a protocol violation. The connection is preserved. `PUBREC` / `PUBREL` already had spec-compliant handling for unknown identifiers and are unchanged.

`IMqttClient` now documents its concurrency contract: `PublishAsync`, `SubscribeAsync`, `UnsubscribeAsync` and `PingAsync` are safe to invoke concurrently from multiple threads. `ConnectAsync` / `DisconnectAsync` must still be coordinated on a single thread.

## Tests

Two new tests in `MqttClient_Tests`:

- `Cancelled_Publish_Does_Not_Disconnect_Client` (`AtLeastOnce`, `ExactlyOnce`) - uses `InterceptingPublishAsync` to delay the broker''s ack, cancels the publish, and asserts the client stays connected and remains usable. Verified to fail on the unfixed code with `MqttProtocolViolationException`.
- `Concurrent_Publish_From_Multiple_Threads` (`AtMostOnce`, `AtLeastOnce`, `ExactlyOnce`) - 8 threads x 250 messages, asserts no disconnect and that every message is delivered to a subscriber.

Full suite: **487 / 487 passed** locally on `net8.0`.

## Release notes

Entry added to `Source/ReleaseNotes.md`.

### DIFF
--- a/Source/MQTTnet.Tests/Clients/MqttClient/MqttClient_Tests.cs
+++ b/Source/MQTTnet.Tests/Clients/MqttClient/MqttClient_Tests.cs
@@ -870,4 +870,109 @@ public sealed class MqttClient_Tests : BaseTestClass
         Assert.IsTrue(client1.TryPingAsync().GetAwaiter().GetResult());
         Assert.IsFalse(disconnectedFired);
     }
+
+    // Regression test for issue #2079.
+    // A late PUBACK (one whose matching publish request was already cancelled or timed
+    // out client-side) must NOT raise a protocol violation and must NOT disconnect the
+    // client. Before the fix, the client received the late PUBACK, failed to find a
+    // waiter and threw a MqttProtocolViolationException, tearing down the connection.
+    [TestMethod]
+    [DataRow(MqttQualityOfServiceLevel.AtLeastOnce)]
+    [DataRow(MqttQualityOfServiceLevel.ExactlyOnce)]
+    public async Task Cancelled_Publish_Does_Not_Disconnect_Client(MqttQualityOfServiceLevel qos)
+    {
+        using var testEnvironment = new TestEnvironment(TestContext);
+        var server = await testEnvironment.StartServer();
+
+        // Delay the broker's acknowledgement long enough for the publish caller to give
+        // up. When the caller's cancellation token fires the awaitable is removed; the
+        // (slightly later) PUBACK / PUBREC then arrives without a matching waiter.
+        server.InterceptingPublishAsync += async e =>
+        {
+            await Task.Delay(TimeSpan.FromSeconds(2), CancellationToken.None);
+        };
+
+        var client = await testEnvironment.ConnectClient();
+
+        var disconnected = false;
+        client.DisconnectedAsync += _ =>
+        {
+            disconnected = true;
+            return CompletedTask.Instance;
+        };
+
+        var message = new MqttApplicationMessageBuilder().WithTopic("late-ack").WithQualityOfServiceLevel(qos).Build();
+
+        using (var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(250)))
+        {
+            await Assert.ThrowsExactlyAsync<MqttCommunicationTimedOutException>(() => client.PublishAsync(message, cts.Token));
+        }
+
+        // Wait long enough for the delayed broker acknowledgement to come back.
+        await Task.Delay(TimeSpan.FromSeconds(3));
+
+        Assert.IsFalse(disconnected, "Client must not disconnect when a late acknowledgement arrives.");
+        Assert.IsTrue(client.IsConnected, "Client must still be connected after a late acknowledgement.");
+
+        // The connection must remain usable for further publishes.
+        var followUp = new MqttApplicationMessageBuilder().WithTopic("late-ack").WithQualityOfServiceLevel(MqttQualityOfServiceLevel.AtMostOnce).Build();
+        await client.PublishAsync(followUp);
+    }
+
+    // Regression test for issue #2078.
+    // The MqttClient must support concurrent publishes from multiple threads without
+    // dropping messages, mismatching acknowledgements or disconnecting the client.
+    [TestMethod]
+    [DataRow(MqttQualityOfServiceLevel.AtMostOnce)]
+    [DataRow(MqttQualityOfServiceLevel.AtLeastOnce)]
+    [DataRow(MqttQualityOfServiceLevel.ExactlyOnce)]
+    public async Task Concurrent_Publish_From_Multiple_Threads(MqttQualityOfServiceLevel qos)
+    {
+        const int Threads = 8;
+        const int MessagesPerThread = 250;
+
+        using var testEnvironment = new TestEnvironment(TestContext);
+        await testEnvironment.StartServer();
+
+        var publisher = await testEnvironment.ConnectClient();
+
+        var subscriber = await testEnvironment.ConnectClient();
+        var receivedCount = 0;
+        subscriber.ApplicationMessageReceivedAsync += _ =>
+        {
+            Interlocked.Increment(ref receivedCount);
+            return CompletedTask.Instance;
+        };
+        await subscriber.SubscribeAsync("concurrent/#", qos);
+
+        var disconnected = false;
+        publisher.DisconnectedAsync += _ =>
+        {
+            disconnected = true;
+            return CompletedTask.Instance;
+        };
+
+        var tasks = Enumerable.Range(0, Threads).Select(threadIndex => Task.Run(async () =>
+        {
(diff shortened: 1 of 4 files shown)


araxis:
@dotnet-policy-service agree

chkr1011 (approved):

PR #2250, opened 2026-05-11
Fix MQTT v5 topic alias forwarding

### LINKED ISSUE #2199: Incorrect handling of Topic Alias?
Am I right that here:

https://github.com/dotnet/MQTTnet/blob/247988e6334842e95e242a3326c43cc3293c8828/Source/MQTTnet.Server/Internal/Formatter/MqttPublishPacketFactory.cs#L62

the server simply copies Topic Alias used by the message publisher to messages sent to subscribers? In other words, subscribers get the same Topic Alias property that was present in the original published message. Correct?

If so, I suspect such a behavior is against the MQTT v5 specs. If I understand it correctly, the sender (in this case a client that publishes a message) can set a Topic Alias in a PUBLISH and then in subsequent PUBLISH messages the Topic Alias can be used instead of the Topic Name.

> The sender decides whether to use a Topic Alias and chooses the value. It sets a Topic Alias mapping by including a non-zero length Topic Name and a Topic Alias in the PUBLISH packet. The receiver processes the PUBLISH as normal but also sets the specified Topic Alias mapping to this Topic Name.
> 
> If a Topic Alias mapping has been set at the receiver, a sender can send a PUBLISH packet that contains that Topic Alias and a zero length Topic Name. The receiver then treats the incoming PUBLISH as if it had contained the Topic Name of the Topic Alias.

However, the spec also says:

> Topic Alias mappings exist only within a Network Connection and last only for the lifetime of that Network Connection. A receiver MUST NOT carry forward any Topic Alias mappings from one Network Connection to another [MQTT-3.3.2-7].

This means to me that the Topic Alias mapping between the message publisher and the server (where the client is the sender and the server is the receiver) is NOT the same mapping as between the server and a subscriber (where the server is the sender and the subscriber is the receiver). The Topic Alias mapping is specific to a given pair of sender and receiver.

For example, `client1` publishes a message on `topicA` setting a Topic Alias `1`, while `client2` publishes a message on the same `topicA` setting a Topic Alias `10`. Messages on `topicA` forwarded from `client1` to `client2` will contain Topic Alias `1` even though for `client2` such a Topic Alias is unknown.

Furthermore:

> The Topic Alias mappings used by the Client and Server are independent from each other. Thus, when a Client sends a PUBLISH containing a Topic Alias value of 1 to a Server and the Server sends a PUBLISH with a Topic Alias value of 1 to that Client they will in general be referring to different Topics.

Am I missing anything?

suhashollakc:
Fixes #2199

## Summary

This fixes MQTT v5 topic alias handling in the server publish path.

Topic Alias mappings are scoped to a single network connection. When a client publishes a message with a Topic Alias, the server should resolve that alias for the inbound connection, but it should not forward the same Topic Alias to subscribers on separate client connections.

## Changes

- Clears `TopicAlias` before dispatching a received application message to subscriber sessions.
- Updates MQTT v5 property forwarding coverage so subscriber messages no longer expect the publisher's alias.
- Extends the server topic alias test to verify subscribers receive resolved topics without forwarded aliases.

## Validation

- `dotnet test --project Source\MQTTnet.Tests\MQTTnet.Tests.csproj --framework net10.0 --filter "Topic_Alias_Tests|Publish_And_Receive_New_Properties" --no-restore`
- `dotnet build MQTTnet.sln --no-restore`

Note: a full local test run was not clean in this environment due to unrelated TLS credential/runtime/network-sensitive tests.

### DIFF
--- a/Source/MQTTnet.Server/Internal/MqttConnectedClient.cs
+++ b/Source/MQTTnet.Server/Internal/MqttConnectedClient.cs
@@ -218,6 +218,8 @@ public sealed class MqttConnectedClient : IDisposable
         HandleTopicAlias(publishPacket);
 
         var applicationMessage = MqttApplicationMessageFactory.Create(publishPacket);
+        // Topic aliases are scoped to this network connection and must not be forwarded to other clients.
+        applicationMessage.TopicAlias = 0;
 
         var dispatchApplicationMessageResult =
             await _sessionsManager.DispatchApplicationMessage(Id, UserName, Session.Items, applicationMessage, cancellationToken).ConfigureAwait(false);
--- a/Source/MQTTnet.Tests/MQTTv5/Client_Tests.cs
+++ b/Source/MQTTnet.Tests/MQTTv5/Client_Tests.cs
@@ -114,7 +114,7 @@ public sealed class Client_Tests : BaseTestClass
 
         Assert.IsNotNull(receivedMessage);
         Assert.AreEqual(applicationMessage.Topic, receivedMessage.Topic);
-        Assert.AreEqual(applicationMessage.TopicAlias, receivedMessage.TopicAlias);
+        Assert.AreEqual(0, receivedMessage.TopicAlias);
         Assert.AreEqual(applicationMessage.ContentType, receivedMessage.ContentType);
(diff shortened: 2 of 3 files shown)


suhashollakc:
@dotnet-policy-service agree

PR #2266, opened 2026-09-03
Honor disconnect Will reasons and session lifecycle

### LINKED ISSUE #2256: Server: DISCONNECT with DisconnectWithWillMessage reason code is ignored — Will message is not published
### Verification

1. Not applicable — this is a protocol-correctness bug (a Will message that should be delivered is silently dropped), not a performance/memory/CPU issue.
2. Not applicable, same reason as above.
3. Reproduced with two independent, unmodified `MQTTnet` clients talking to an in-process `MQTTnet.Server` instance — no application-specific wrapper code involved (repro below uses only `MQTTnet` / `MQTTnet.Server` APIs directly).
4. Confirmed present on the latest `master` (`Source/MQTTnet.Server/Internal/MqttConnectedClient.cs`, lines 115-117 as of this writing) as well as on the released NuGet package version 5.1.0.1559.

### Describe the bug

When a client sends `DISCONNECT` with reason code `0x04` (`DisconnectWithWillMessage`) — telling the server to publish its Will Message even though this is a clean, client-initiated disconnect (MQTT 5.0 §3.14.2.1) — the server never publishes the Will. It behaves identically to a plain `NormalDisconnection`.

The cause is in `MqttConnectedClient.RunAsync`:

```csharp
var isCleanDisconnect = DisconnectPacket != null;

if (!IsTakenOver && !isCleanDisconnect && Session.LatestConnectPacket.WillFlag && !Session.WillMessageSent)
```

`isCleanDisconnect` only checks whether *any* DISCONNECT packet was received — it never inspects `DisconnectPacket.ReasonCode`. So any client-sent DISCONNECT suppresses the Will, regardless of which reason code it carries. Per the spec, only reason codes *other than* `0x04` should suppress it; `DisconnectWithWillMessage` should still trigger delivery.

- Server

### To Reproduce

1. Using this version of MQTTnet `5.1.0.1559` (also reproduces on latest `master`).
2. Run this code:
   - Start an `MqttServer` on a local port.
   - Connect client `receiver`, subscribe to `#`.
   - Connect client `sender` with a Will registered (`WithWillTopic`/`WithWillPayload`).
   - Call `sender.DisconnectAsync(MqttClientDisconnectOptionsReason.DisconnectWithWillMessage)`.
3. With these arguments: any Will topic/payload; QoS 0 or 1, retain or not — reproduces regardless.
4. See error: `receiver` never gets the Will message. (For contrast, calling `sender.Dispose()` instead — an ungraceful drop with no DISCONNECT packet at all — *does* deliver the Will, confirming the Will registration itself is fine; only the explicit "disconnect with will" request is ignored.)

### Expected behavior

`receiver` should receive the Will message on `sender`'s configured Will topic, because `sender` explicitly disconnected with reason code `DisconnectWithWillMessage`.

### Screenshots

Not applicable.

```batch
No exceptions or errors are logged. The server logs a normal clean disconnect and simply
never logs "Published will message" for this client, whereas it does log that line for the
`Dispose()`-without-DISCONNECT case using the same Will configuration.
```

### Code example

Minimal repro using only `MQTTnet` / `MQTTnet.Server` (no application code):

```csharp
using System.Net;
using System.Net.Sockets;
using MQTTnet;
using MQTTnet.Protocol;
using MQTTnet.Server;

int GetFreePort()
{
    using var listener = new TcpListener(IPAddress.Loopback, 0);
    listener.Start();
    return ((IPEndPoint)listener.LocalEndpoint).Port;
}

var port = GetFreePort();
var serverFactory = new MqttServerFactory();
var server = serverFactory.CreateMqttServer(
    serverFactory.CreateServerOptionsBuilder().WithDefaultEndpoint().WithDefaultEndpointPort(port).Build());
await server.StartAsync();

var clientFactory = new MqttClientFactory();

var receiver = clientFactory.CreateMqttClient();
var willReceived = false;
receiver.ApplicationMessageReceivedAsync += e =>
{
    if (e.ApplicationMessage.Topic == "status") willReceived = true;
    return Task.CompletedTask;
};
await receiver.ConnectAsync(new MqttClientOptionsBuilder().WithTcpServer("127.0.0.1", port).WithClientId("receiver").Build());
await receiver.SubscribeAsync(new MqttClientSubscribeOptionsBuilder()
    .WithTopicFilter(f => f.WithTopic("status").WithQualityOfServiceLevel(MqttQualityOfServiceLevel.AtLeastOnce)).Build());

var sender = clientFactory.CreateMqttClient();
await sender.ConnectAsync(new MqttClientOptionsBuilder()
    .WithTcpServer("127.0.0.1", port)
    .WithClientId("sender")
    .WithWillTopic("status")
    .WithWillPayload("DISCONNECTED")
    .WithWillQualityOfServiceLevel(MqttQualityOfServiceLevel.AtLeastOnce)
    .Build());

// Explicitly asks the broker to deliver the Will even though this is a clean disconnect.
await sender.DisconnectAsync(MqttClientDisconnectOptionsReason.DisconnectWithWillMessage);

await Task.Delay(1000);

Console.WriteLine(willReceived ? "PASS: Will delivered" : "FAIL: Will NOT delivered");
// Actual: FAIL. Expected: PASS.
```

**Ideally a Unit Test (which shows the error) is provided so that the behavior can be reproduced easily.**

A `[TestMethod]` in the style of the existing `Source/MQTTnet.Tests/Server/Will_Tests.cs` (which already covers the "no Will on plain clean disconnect" case but has no coverage for the `DisconnectWithWillMessage` case this bug affects):

```csharp
[TestMethod]
public async Task Will_Message_Sent_On_DisconnectWithWillMessage()
{
    using var testEnvironment = CreateTestEnvironment();

    var receiver = await testEnvironment.ConnectClient();
    var receivedMessages = testEnvironment.CreateApplicationMessageHandler(receiver);
    await receiver.SubscribeAsync(new MqttTopicFilterBuilder().WithTopic("#").Build());

    var clientOptions = new MqttClientOptionsBuilder().WithWillTopic("My/last/will");
    var sender = await testEnvironment.ConnectClient(clientOptions);

    // Client explicitly asks the broker to deliver the Will despite this being a clean disconnect.
    await sender.DisconnectAsync(MqttClientDisconnectOptionsReason.DisconnectWithWillMessage);

    await LongTestDelay();

    // Currently FAILS: 0 messages received. Expected: 1 (the Will message).
    Assert.HasCount(1, receivedMessages.ReceivedEventArgs);
}
```

Suggested fix in `Source/MQTTnet.Server/Internal/MqttConnectedClient.cs`:

```diff
- var isCleanDisconnect = DisconnectPacket != null;
+ var disconnectPacket = DisconnectPacket;
+ var suppressWill = disconnectPacket != null && disconnectPacket.ReasonCode != MqttDisconnectReasonCode.DisconnectWithWillMessage;

- if (!IsTakenOver && !isCleanDisconnect && Session.LatestConnectPacket.WillFlag && !Session.WillMessageSent)
+ if (!IsTakenOver && !suppressWill && Session.LatestConnectPacket.WillFlag && !Session.WillMessageSent)
```

This keeps the existing `Will_Message_Do_Not_Send_On_Clean_Disconnect` test passing (a plain `NormalDisconnection` still suppresses the Will) while fixing the `DisconnectWithWillMessage` case per MQTT 5.0 §3.14.2.1.

YAJeff:
Fixes #2256.

The server currently suppresses the Will for every client DISCONNECT and every takeover, and otherwise publishes immediately without observing Will Delay. Honor DISCONNECT 0x04 and the MQTT Will lifecycle with an internal scheduler that captures each connection's Will, identity and deadline.

Only normal client DISCONNECT (0x00) suppresses the Will. Other eligible disconnects publish after the earlier of Will Delay and session expiry. Resuming the same session before that deadline cancels the pending Will; an overdue Will, clean start, or zero-delay takeover publishes the old connection's Will. Connection ownership prevents late cleanup from changing the new connection's Will. Captured message expiry begins at publication.

The scheduler dispatches outside session/state locks, owns its background task, and preserves undispatched messages if a callback stops the server. With persistent sessions, deadlines survive stop/restart of the same server instance and overdue work runs before accepting reconnects. Nonpersistent shutdown and final disposal end sessions and flush eligible Wills. This adds no durable storage or public API, and does not claim persistence across process restart.

Dependency: this PR consumes the session's existing ExpiryInterval property. Correct DISCONNECT expiry replacement, including explicit zero, and resumed CONNECT expiry require #2264. Those changes are deliberately not bundled here. Both PRs were also tested together in an isolated local checkout.

Specification:
- MQTT 5.0 sections 3.1.2.5, 3.1.3.2.2, 3.1.3.2.4 and 3.14.4: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html
- MQTT 3.1.1 [3.1.2-8] and [3.1.4-2] require the old connection's Will on takeover without its DISCONNECT: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html

Compatibility: normal-disconnect suppression remains; MQTT 3.1.1 takeover now publishes the old Will as required. The existing takeover test expected suppression and is corrected accordingly. Timed dispatch has 100 ms scheduling granularity. Existing persistent-session storage remains in memory.

Synthetic regressions for DISCONNECT 0x04, another nonzero reason and ignored Will Delay all fail on unmodified master 18731d9f; the normal-disconnect control passes. Tests cover takeover/resume ownership, rejected and expired reconnects, expiry versus delay, deletion/timer races, retained/message-expiry fields, disabled persistence, final disposal, active/pending stop and restart, callback-triggered shutdown from timer/startup, preserved next Wills, and nonrecursive nonpersistent shutdown.

Validation:
- All 29 focused Will/takeover cases pass on net8.0.
- Combined with #2264: all 41 selected Will/expiry cases pass on net10.0. Separate wire probes confirm Will timing for DISCONNECT expiry zero, one, absent, and a resumed CONNECT expiry one.
- Release solution build passed with three pre-existing benchmark deprecation warnings. Final full standalone net10.0 suite: 511 passed, zero failures.
- Independent design and complete staged-diff reviews completed before publication; git diff --check passes.


YAJeff:
The failed CI job reports 510/511 tests passing. The failure is `Concurrent_Publish_From_Multiple_Threads (ExactlyOnce)`: `TestEnvironment.Dispose` detects an unobserved `MqttCommunicationException: Broken pipe` from the server send loop. All Will-specific cases passed.

I reran all three concurrent-publish variants on .NET 10 locally against both unmodified master `18731d9f` and this PR's unchanged head `ede8eb86`; both selections passed. This suggests a cleanup/timing issue, but the local Windows runs do not establish the cause of the Linux CI failure.

Could a maintainer rerun the failed test job? My API rerun request was denied with HTTP 403. Job: https://github.com/dotnet/MQTTnet/actions/runs/33817487297/job/100852725884

